PR #91, opened 2020-05-20
Applications REST API

winder:
Implements the REST API described in this document: https://docs.google.com/document/d/1Mn1OPQd4QlAhupO41Rre--WDzIGshvN6Y-8Uk7OA-DI/edit

**/v2/applications**
*  application-id: fetch a particular application
* round: rewind application state to a particular round. Disabled by default similar to **/v2/accounts?round**

**/v2/applications/{application-id}**
* round: rewind application state to a particular round.

**/v2/transactions**
* application-id: Only include transactions relating to the given application-id
* created-application-index: New field to transaction object.

**/v2/accounts**
* application-id: only include accounts which use the given application

**Transaction Model Updates**
Include the new Application transaction fields.

**NOT INCLUDED**
We discussed adding a parameter to **/accounts** which configures whether to include the global state for an application. I didn't add it for several reasons:
1) I'm not sure what scenario you actually need to know the global state for all of your applications. Lets let the user run a followup query for the particular application they're interested in.
2) The data model isn't defined. Where would it go? How do we display "global" and "local" data without it being confusing?
3) Lets get it working first before complicating things by merging multiple DB layer queries.


api/generated/v2/types.go
```diff
@@ -376,6 +476,46 @@ type Transaction struct {
...
+	Accounts *[]string `json:"accounts,omitempty"`
+
+	// \[apaa\] transaction specific arguments accessed from the application's approval-program and clear-state-program.
+	ApplicationArgs *[]string `json:"application-args,omitempty"`
+
+	// \[apid\] ID of the application being configured or empty if creating.
+	ApplicationId *uint64 `json:"application-id,omitempty"`
+
+	// \[apap\] Logic executed for every application transaction, except when on-completion is set to "clear". It can read and write global state for the application, as well as account-specific local state. Approval programs may reject the transaction.
+	ApprovalProgram *[]byte `json:"approval-program,omitempty"`
```
algomaxj, line 492: I noticed that some of these generated types have `[]byte`, and others have `*[]byte`. Does this matter?

winder: Yeah, the * means it is optional. I forgot to mark required/optional on these.

algorotem:
1) Fair enough 
2) I'd expect `/v2/applications/{id}` to the return the global and `/v2/accounts?app-id = {id}` to return the local.

api/generated/common/types.go
```diff
@@ -79,6 +79,63 @@ type AccountParticipation struct {
 	VoteParticipationKey []byte `json:"vote-participation-key"`
 }
 
+// Application defines model for Application.
+type Application struct {
+
+	// \[appidx\] application index.
+	AppIndex uint64 `json:"app-index"`
```
algorotem, line 86: should we keep it consistent with assets? `app-index` vs `index`. they're both should be probably just `id` given the context.

api/generated/v2/routes.go
```diff
@@ -857,6 +936,7 @@ func (w *ServerInterfaceWrapper) SearchForTransactions(ctx echo.Context) error {
 		"address-role":          true,
 		"exclude-close-to":      true,
 		"rekey-to":              true,
+		"application-id":        true,
```
algorotem, line 939: same here - let's find one short convention.

api/generated/v2/types.go
```diff
@@ -79,6 +79,63 @@ type AccountParticipation struct {
 	VoteParticipationKey []byte `json:"vote-participation-key"`
 }
 
+// Application defines model for Application.
+type Application struct {
+
+	// \[appidx\] application index.
+	AppIndex uint64 `json:"app-index"`
```
algorotem, line 86: just to make sure we don't forget it here

api/generated/common/types.go
winder: Yes, definitely.

api/generated/v2/routes.go
winder: Looks like "index" breaking conventions. It's only used by the `Asset` model. Everything else uses "asset-id"

winder: Correction, there is also `created-asset-index`.

PR #110, opened 2020-06-03
Updating README systemd example to properly override ExecStart.

PSjoe:


winder:
Thanks for pointing this out Joe

PR #113, opened 2020-06-04
Build indexer pipeline, redux

btoll:
Completed:
- hooked up python e2e tests for jenkins
- generate releases page

This also removes the `mule` abstraction and instead calls the `make` targets directly.  In addition, this adds a new Jenkinsfile in the project root that contains a list of targets that Jenkins will call:

```
@Library('make')_

make(['package', 'stage-packages', 'test', 'test-package', 'sign', 'deploy', 'generate-releases-page'])
```

Tickets relating to first iteration of this pipeline:
https://github.com/algorand/indexer/pull/103
https://github.com/algorand/indexer/pull/108

The glue code to get this to run in Jenkins is simply:
https://github.com/btoll/test-jenkins/blob/master/vars/make.groovy


mule.yaml
```diff
@@ -79,29 +39,19 @@ tasks:
 
   - task: s3.BucketCopy
     name: setup-deb
```
winder, line 41: Missed this before, should the name be `upload-deb`?

btoll: This was intended to download the package from s3 if someone had previously built and pushed the packages.  The idea was to be able to have devs run tasks "a la carte" as another option to just running one task after another.  I don't know if this applies to the indexer project.

The name isn't the greatest.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ENV GOPATH=$HOME/go
+ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+
+RUN apt-get update && apt-get install -y apt-transport-https awscli ca-certificates build-essential curl git software-properties-common && \
+    curl https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar xzf - && \
+    mv go /usr/local && \
+    export PATH=/usr/local/go/bin:$PATH && \
+    mkdir -p $HOME/go/src/github.com/algorand/indexer
+
+# Install postgres and python3 for packaging and e23 tests.
```
brianolson, line 16: looks like ubuntu18 has an apt package `python3-psycopg2` that could replace apt libpq-dev and pip psycopg2. Would that be a meaningful simplification here?

Makefile
```diff
@@ -1,19 +1,18 @@
 SRCPATH		:= $(shell pwd)
-OS_TYPE		:= $(shell ./scripts/ostype.sh)
-   ARCH		:= $(shell ./scripts/archtype.sh)
-PKG_DIR		= $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)
+VERSION		:= $(shell $(SRCPATH)/scripts/compute_build_number.sh)
+PKG_DIR		= $(SRCPATH)/packages/$(VERSION)
 
 # This is the default target, build the indexer:
-cmd/algorand-indexer/algorand-indexer:	idb/setup_postgres_sql.go importer/protocols_json.go .PHONY
+cmd/algorand-indexer/algorand-indexer: idb/setup_postgres_sql.go importer/protocols_json.go .PHONY
```
brianolson, line 6: Why did this change? It looks like a tab was replaced with a space

build/e2e.sh
```diff
@@ -0,0 +1,16 @@
...
+PORT=5432
+VERSION=$(./scripts/compute_build_number.sh)
+
+dpkg -i "./packages/$VERSION/algorand-indexer_${VERSION}_amd64.deb"
+
+/etc/init.d/postgresql start
+sudo -u postgres bash -c "psql -c \"CREATE ROLE $USER WITH SUPERUSER CREATEDB LOGIN ENCRYPTED PASSWORD '$USER';\""
+sudo -u postgres bash -c "psql -c \"CREATE DATABASE $DB_NAME;\""
+
+python3 misc/e2etest.py --connection-string "host=localhost port=$PORT dbname=$DB_NAME sslmode=disable user=$USER password=$USER"
```
brianolson, line 15: Why not let it create a dbname as usual? Is this something that Docker needs? If this is a test that should be run inside Docker for testing then we shouldn't reference build/e2e.sh from the Makefile test target which will probably be used by developers not inside docker.

misc/e2etest.py
```diff
@@ -150,9 +150,9 @@ def main():
     else:
         psqlstring = args.connection_string
 
-    xrun(['cmd/algorand-indexer/algorand-indexer', 'import', '-P', psqlstring, os.path.join(e2edata, 'blocktars', '*'), '--genesis', os.path.join(e2edata, 'algod', 'genesis.json')], timeout=20)
+    xrun(['algorand-indexer', 'import', '-P', psqlstring, os.path.join(e2edata, 'blocktars', '*'), '--genesis', os.path.join(e2edata, 'algod', 'genesis.json')], timeout=20)
```
brianolson, line 153: This now relies on `go install` or otherwise getting the binary into the PATH?

Makefile
btoll: I had changed that to match the other targets that have spaces there.  Is the tab meaningful there?

Dockerfile
btoll: I will try that package.

misc/e2etest.py
btoll: Yes.  We can talk about this if you'd like to determine the best way to test this.

build/e2e.sh
btoll: I passed it the connection string b/c it needed the password for the user.  I was thinking about how a dev would test this, as well, for the same reasons you bring up.

It definitely needs some work to achieve a painless process for both our testing and a local dev.

Makefile
brianolson: Dunno where you're seeing spaces. In the Makefile on develop only `package:` is followed by space, so that's the odd one out. I think tab is better.

Makefile
```diff
@@ -22,14 +21,19 @@ package: clean setup
...
+sign:
+	build/sign.sh
+
+stage-packages:
+	aws s3 sync ./packages/$(VERSION) s3://algorand-staging/indexer/$(VERSION)
+
+test: mocks
 	go test ./...
+	docker build -t e2e_tests -f docker/Dockerfile.e2e .
+	docker run -t e2e_tests
```
brianolson, line 33: should this be `docker run --rm` so it cleans up after itself by default?

build/e2e.sh
brianolson: On my dev system with postgres installed $USER is a postgres super user so misc/e2etest.py creating the database just works. I'd expect line 12 above to take care of that.

Makefile
btoll: I'll change back.

Dockerfile
btoll: This worked great, thanks.

misc/e2etest.py
brianolson: these changes should be replaced by #120

btoll:
I'm closing this to remove it from the queue as another PR that fixes these issues has been merged.  I'm going to continue to work on the idea presented here.

PR #115, opened 2020-06-05
Fix docker files and readme commands

Alex99y:


brianolson:
the readme options fixes are certainly right. pinging @winder to double check the Docker stuff but it looks okay to me.

Issue #126, opened 2020-06-15
test live indexing and fetch by txid


brianolson:
start a private network, start an indexer following it, submit a txn (get its txid), wait 10 seconds, see if the txn is available by txid in the indexer

PR #128, opened 2020-06-15
Test indexer following live algod; returning a txn by txid

brianolson:
implements #126

### DIFF
--- a/misc/liveindextest.py
+++ b/misc/liveindextest.py
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+#
+# usage:
+#  python3 misc/liveindextest.py
+#
+# Requires go-algorand to be checked out on GOPATH.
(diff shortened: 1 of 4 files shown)


misc/liveindextest.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/env python3
+#
+# usage:
+#  python3 misc/liveindextest.py
+#
+# Requires go-algorand to be checked out on GOPATH.
+# Requires local postgresql and `createdb` `dropdb` standard utils.
```
winder, line 7: Is it safe to assume that this is only true when a connection string isn't provided?

brianolson: Yes. I guess I only documented the default case. An externally managed database with passed in connection string is neither created nor destroyed within this script.

PR #130, opened 2020-06-16
Fix hashes files

btoll:
This removes "./" from the beginning of some lines in the hashes files.


btoll:
For go-algorand: https://github.com/algorand/go-algorand/pull/1173

PR #151, opened 2020-07-16
Merge ApplicationLocalStates and ApplicationLocalState.

winder:
Noticed the account object had some unnecessary nesting for storing the local states.


winder:
> SDK writers know and are making a parallel change?

Yup, this will definitely effect the SDKs

PR #153, opened 2020-07-21
More Applications API

brianolson:
Implement /v2/applications and /v2/applications/{N}
Improve test code

### DIFF
--- a/idb/postgres.go
+++ b/idb/postgres.go
@@ -1079,7 +1079,7 @@ func (db *PostgresIndexerDb) GetBlock(round uint64) (block types.Block, err erro
 	return
 }
 
-func buildTransactionQuery(tf TransactionFilter) (query string, whereArgs []interface{}) {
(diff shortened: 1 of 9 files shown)


idb/postgres.go
```diff
@@ -1155,9 +1155,22 @@ func buildTransactionQuery(tf TransactionFilter) (query string, whereArgs []inte
...
 		partNumber++
 	}
-	if tf.AssetId != 0 {
+	if tf.AssetId != 0 || tf.ApplicationId != 0 {
+		var creatableId uint64
+		if tf.AssetId != 0 {
+			creatableId = tf.AssetId
+			if tf.ApplicationId != 0 {
+				if tf.AssetId == tf.ApplicationId {
+					// this is nonsense, but I'll allow it
```
winder, line 1164: This doesn't sound like you :laughing: 

I think it can be an error.

PR #160, opened 2020-07-27
Fill in app transaction fields in REST response

winder:
Update TxnRow conversion function to include application fields.


misc/generate_test_files.sh
```diff
@@ -0,0 +1,75 @@
...
+# data [query] - print all encoded transactions, useful to quickly find which
+#                TXIDs are useful for whatever test is being written.
+#                Optionally provide a query.
+
+export PGPASSWORD='harness'
+
+TEMP_FILE='/tmp/copy_tmp'
+
+
+# Arg1 TXID        txid to download using 'escape' notation, the leading slash will be added back in.
```
brianolson, line 18: such a TXID would be fragile though? What source data set is the txid from? That's likely to change, right?

winder: It's from the SDK test environment, so they should not change. This script happens to be using "dataset2" (the snapshot you emailed me recently), if/when I need to pull transactions from another dataset I'll turn that into a parameter.

Since the data being extracted should never change, including them in this script is mainly to serve as a reference for how the previous files were generated.

api/converter_utils.go
```diff
@@ -295,9 +314,42 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 			NewFreezeStatus: stxn.Txn.AssetFrozen,
 		}
 		assetFreeze = &f
-	}
+	case sdk_types.ApplicationCallTx:
+		args := make([]string, 0)
```
brianolson, line 318: should these be nil instead of empty if the source is empty?

winder: Should be fine either way.

PR #161, opened 2020-07-27
Add in application state delta to transaction responses

winder:
Update spec to have the same StateDelta fields as algod. I needed a new LocalStateDelta type as well.

Also populate the new field and update tests.


api/converter_utils.go
```diff
@@ -357,6 +358,66 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 		Sig:      sigToTransactionSig(stxn.Sig),
 	}
 
+	// The state delta bits need to be sorted for testing. Maybe it would be
+	// for end users too, people always seem to notice results changing.
+	convertDelta := func (d types.StateDelta) *generated.StateDelta {
```
brianolson, line 363: pop this out to a top level function instead of a closure?

PR #167, opened 2020-08-03
Fix Asset Config

brianolson:
Expand validation to check asset configs inside their creating accounts.
Fix accounting of changes to asset config.
Migration to repair asset configs from their original transaction data.


brianolson:
#169 is the better option now

Issue #157, opened 2020-07-23
expand validation around asset metadata


brianolson:
misc/validate_accounting.py checks Algo balances, asset balances, app local state, app global config and state, ... and is missing checking account['created-assets'] asset config.

brianolson:
probably do at the same time as #152

Issue #152, opened 2020-07-21
repeated asset-config merge incorrectly


brianolson:
It appears that multiple 'acfg' AssetConfig operations on an asset are getting merged incorrectly and blank fields in later configs are clobbering present fields in earlier configs. This was noted in that some assets are missing names.

PR #169, opened 2020-08-04
App proto fix acfg

brianolson:
This is the sum of #167 and #168 
Expand validation to check asset configs inside their creating accounts.
Fix accounting of changes to asset config.
Migration to repair asset configs from their original transaction data.
Update protocols for applications and re-key.
Addresses #152 #157


idb/postgres_migrations.go
```diff
@@ -97,6 +103,136 @@ func m2apps(db *PostgresIndexerDb, state *MigrationState) error {
...
+			return i, nil
+		}
+		txrows := db.txTransactions(tx, TransactionFilter{TypeEnum: 3, AssetId: uint64(aid)})
+		prevRound := uint64(0)
+		first := true
+		var params types.AssetParams
+		var creator types.Address
+		for txrow := range txrows {
+			if txrow.Round < prevRound {
+				log.Printf("acfg rows out of order %d < %d", txrow.Round, prevRound)
```
winder, line 189: Should this be an error? Wouldn't a return here cause the remaining assets to be skipped?

winder: It looks like there are a couple ways that this could fail. I think that we need to make sure Indexer is halted with an error rather than letting it be brought up in an inconsistent state.

idb/postgres_migrations.go
```diff
@@ -140,16 +276,22 @@ func init() {
...
+type migrationContext struct {
 	db      *PostgresIndexerDb
 	state   *MigrationState
 	lastlog time.Time
 }
+type txidFiuxpMigrationContext migrationContext
+
+/*struct {
+	migrationContext
+}*/
```
winder, line 294: looks like a left over comment

idb/postgres_migrations.go
```diff
@@ -215,7 +357,7 @@ func (mtxid *txidFiuxpMigrationContext) asyncTxidFixup() {
 		}
 	}
 	// all done, mark migration state
-	state.NextMigration = 1
+	state.NextMigration++
```
winder, line 360: :+1:

misc/buildtestdata.sh
```diff
@@ -33,6 +33,8 @@ mkdir -p "${E2EDATA}"
 
 (cd "${E2EDATA}" && tar -j -c -f net_done.tar.bz2 --exclude node.log --exclude agreement.cdv net)
 
+if false; then
```
winder, line 36: Was this supposed to be committed? Looks like a debugging thing.

idb/postgres_migrations.go
winder: Approving despite this because (a) I trust that postgres can order it's results and (b) algod shouldn't accept invalid transactions. But maybe one of the queries could get dropped and cause a failure as well, so we probably need a way to return errors instead of silently failing to complete a migration.

misc/buildtestdata.sh
brianolson: This is part of moving to a new e2e validation script. The data build gets smaller and simpler. Moving from e2etest.py which compares algod sqlite to indexer to e2elive.py which compares algod API to indexer which should be more correct and more sustainable in the future.

idb/postgres_migrations.go
brianolson: return here causes the migration to stop uncompleted. the migration will run again the next time indexer starts (possibly after a binary upgrade with a bug fix)

Issue #181, opened 2020-08-13
Postgres Tx that can be ReadOnly should be


brianolson:
in idb/postgres.go set TxOption.ReadOnly on all transactions that are read-only.
Current 'read only' mode works on Postgres database users that only have read permissions on tables, but when the whole database is in read-only mode (a hot spare or follower replicated database) it fails on transactions that _might_ write.

PR #182, opened 2020-08-14
db read-only

brianolson:
Make a read-only transaction flagged as such.
When --no-algorand flag is set then migrations and other db init should not be done.
#181

### DIFF
--- a/cmd/algorand-indexer/main.go
+++ b/cmd/algorand-indexer/main.go
@@ -66,18 +66,17 @@ var (
 	postgresAddr   string
 	dummyIndexerDb bool
 	doVersion      bool
-	dbReadOnly     bool
 	cpuProfile     string
 	pidFilePath    string
 	db             idb.IndexerDb
 	profFile       io.WriteCloser
 )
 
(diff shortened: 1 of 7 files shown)


cmd/algorand-indexer/main.go
```diff
@@ -94,6 +95,7 @@ func init() {
 
 	rootCmd.PersistentFlags().StringVarP(&postgresAddr, "postgres", "P", "", "connection string for postgres database")
 	rootCmd.PersistentFlags().BoolVarP(&dummyIndexerDb, "dummydb", "n", false, "use dummy indexer db")
+	rootCmd.PersistentFlags().BoolVarP(&dbReadOnly, "db-readonly", "", false, "connect to db in read-only mode")
```
winder, line 98: I'm not sure this needs to be an option, could it be set implicitly when (a) algod configurations aren't set / found or (b) no-algod is set?

brianolson: Hmm, maybe. It might be valid to run migrations even when there isn't algod? Making --no-algod be read-only does also make sense.

PR #190, opened 2020-08-31
some migrations should prevent Indexer from starting up until they're done

### LINKED ISSUE #188: Some migrations should prevent startup
Some migrations can run in the background but some migrations should prevent normal startup (no data fetching from algod, no data serving). Right now a backround migration followed by a stop-the-world migration can leave the system in an unworkable system while the backround migration runs. Do the dependency checking on all pending migrations and don't let the system start up if any pending migration is a stop-the-world migration.

brianolson:
add annotation to postgres migrations about whether they should "prevent startup" (e.g. due to schema change that would make operational txns fail)
simplify async migration model by making each migration blocking but maybe run the loop over migrations in a thread.
cleanup a `go vet` issue in importer.
fix #188

### DIFF
--- a/idb/postgres_migrations.go
+++ b/idb/postgres_migrations.go
@@ -6,7 +6,6 @@ package idb
 import (
 	"context"
 	"database/sql"
-	"errors"
 	"fmt"
 	"log"
 	"math"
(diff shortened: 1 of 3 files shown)


winder (approved):
LGTM

Issue #191, opened 2020-09-02
Version string not linked into release 2.1.1


jecassis:
Here is the output:
```
$ curl -sSLO https://github.com/algorand/indexer/releases/download/2.1.1/algorand-indexer_linux_amd64_2.1.1.tar.bz2
$ tar -xjf algorand-indexer_linux_amd64_2.1.1.tar.bz2
$ ./algorand-indexer_linux_amd64_2.1.1/algorand-indexer -v
(unknown version) compiled at  from git hash
```

Similarly the Debian package install also exhibits the missing version string:
```
$ algorand-indexer -v
(unknown version) compiled at  from git hash
```

brianolson:
`misc/release.py` needs to be updated with the version tricks that `Makefile` is doing

jecassis:
Closing, fixed in 2.2.1.

PR #192, opened 2020-09-02
add version strings to release.py

brianolson:
improve `algorand-indexer -v`, e.g.
2.1.1-dev.unknown compiled at 2020-09-02T20:46:00+0000 from git hash 283e9ecc2f6c0a29d6b9d99b7f9bf3a99f7f52ad (modified)
for #191


misc/release.py
```diff
@@ -97,14 +100,53 @@ def link(sourcepath, destpath):
...
+
+def compile(goos=None, goarch=None, ldflags=None):
     env = dict(os.environ)
     env['CGO_ENABLED'] = '0'
     if goos is not None:
         env['GOOS'] = goos
     if goarch is not None:
         env['GOARCH'] = goarch
-    subprocess.run(['go', 'build'], cwd='cmd/algorand-indexer', env=env).check_returncode()
+    cmd = ['go', 'build']
```
winder, line 146: is it possible to use `make cmd/algorand-indexer/algorand-indexer` instead of duplicating the flags setup?

winder (commented):
Looks pretty good, but do you think now would be a good time to also remove the .version file?

I'm curious what @btoll thinks

Makefile
```diff
@@ -4,6 +4,7 @@ OS_TYPE		:= $(shell $(SRCPATH)/mule/scripts/ostype.sh)
 ARCH		:= $(shell $(SRCPATH)/mule/scripts/archtype.sh)
 PKG_DIR		= $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(VERSION)
 
+# TODO: ensure any additions here are mirrored in misc/release.py
 GOLDFLAGS += -X github.com/algorand/indexer/version.Hash=$(shell git log -n 1 --pretty="%H")
```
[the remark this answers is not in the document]
brianolson: scroll down 4 lines to `git log -n 1 --pretty="%D"`

misc/release.py
brianolson: Maybe? I usually think of `make` as 'do the build if needed' and I want this to be more imperative 'do the build'; a preexisting build (and there will be 4 builds in a row for the os/arch combos) shouldn't prevent any of the building.

brianolson: OR, flip the script. Makefile calls this build script with options to only build for current os/arch

Makefile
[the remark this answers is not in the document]
winder: d'oh

winder:
I think it's fine to merge this in once the test passes

Issue #179, opened 2020-08-11
Consider standard json log output


ksemaev:
Current logs are like:
```
will start from round >1360401
round r=1360402 (0 txn) imported in 18.454454ms
```

There's even to timestamp. Can you please use some standard logging libs, cause it's very hard to work with such log output

ksemaev:
Also please add loglevel cli args or config options

winder:
This has been started, the logrus object is being passed around to all the important objects. We still need to change log/print calls to use the object.

PR #198, opened 2020-09-15
Use viper to manage configuration files/environment variables.

winder:
Introduce the [viper](https://github.com/spf13/viper) library to manage configuration files. In addition to superseding the current yamlish parsing it also adds in the ability to use environment variables to set configuration.

The config file must now be named `indexer`, `indexer.yml`, or `indexer.yaml`
The config file must be located in one of the following locations, if there are multiple config files the first found in the following order is used: `./`, `$HOME/`, `$HOME/.indexer/`, `$HOME/.config/indexer/`, or `/etc/indexer/`

If the same parameter is set in multiple places the following priority is used to choose a setting:
1. command line flag
2. environment variable
3. configuration file

Environment variables are upper cased, and have a `INDEXER_` prefix.

I'll update the documentation soon.


cmd/algorand-indexer/daemon.go
```diff
@@ -31,31 +28,17 @@ var (
...
 	//Args:
 	Run: func(cmd *cobra.Command, args []string) {
-		if configFilePath != "" {
-			cf, err := os.Open(configFilePath)
-			maybeFail(err, "%s: %v", configFilePath, err)
-			err = configFromStream(cf)
-			maybeFail(err, "%s: %v", configFilePath, err)
-			cf.Close()
-		}
+		util.BindFlags(cmd, "INDEXER")
```
algorotem, line 41: Let's make it a constant

cmd/algorand-indexer/daemon.go
```diff
@@ -120,123 +103,20 @@ var daemonCmd = &cobra.Command{
...
-	configStringVarP(daemonCmd.Flags(), &daemonServerAddr, "server", "S", ":8980", "host:port to serve API on (default :8980)")
-	configBoolVarP(daemonCmd.Flags(), &noAlgod, "no-algod", "", false, "disable connecting to algod for block following")
-	configStringVarP(daemonCmd.Flags(), &tokenString, "token", "t", "", "an optional auth token, when set REST calls must use this token in a bearer format, or in a 'X-Indexer-API-Token' header")
-	configBoolVarP(daemonCmd.Flags(), &developerMode, "dev-mode", "", false, "allow performance intensive operations like searching for accounts at a particular round")
-
-	daemonCmd.Flags().StringVarP(&configFilePath, "config", "c", "", "path to 'key: value' config file, keys are same as command line options")
+	logger = log.New()
+	logger.SetFormatter(&log.JSONFormatter{})
+	logger.SetOutput(os.Stdout)
+	logger.SetLevel(log.InfoLevel)
```
algorotem, line 111: Should probably be configurable (or maybe dev mode should enable a lower level one)

cmd/algorand-indexer/import.go
```diff
@@ -243,6 +244,8 @@ var importCmd = &cobra.Command{
 	Short: "import block file or tar file of blocks",
 	Long:  "import block file or tar file of blocks. arguments are interpret as file globs (e.g. *.tar.bz2)",
 	Run: func(cmd *cobra.Command, args []string) {
+		util.BindFlags(cmd, "INDEXER")
```
algorotem, line 247: As well here- make it a constant

cmd/algorand-indexer/main.go
```diff
@@ -97,6 +98,28 @@ func init() {
...
 	rootCmd.PersistentFlags().BoolVarP(&doVersion, "version", "v", false, "print version and exit")
+
+	// Setup viper configuration file
+	viper.SetConfigName("indexer")
+	viper.SetConfigType("yaml")
+	viper.AddConfigPath(".")
+	viper.AddConfigPath("$HOME")
+	viper.AddConfigPath("$HOME/.indexer/")
+	viper.AddConfigPath("$HOME/.config/indexer/")
+	viper.AddConfigPath("/etc/indexer/")
```
algorotem, line 111: nit - I'd expect the name to have the word "config" in it. 

Also, let's make all of it constants - The list of config paths should just be a list

util/config.go
```diff
@@ -0,0 +1,29 @@
+package util
```
algorotem, line 1: consider putting it in a `config` package

cmd/algorand-indexer/main.go
brianolson: maybe be like the binary name, "algorand-indexer" ?

README.md
```diff
@@ -107,6 +108,29 @@ When `--token your-token` is provided, an authentication header is required. For
...
+
+```
+./algorand-indexer daemon [options]
+```
+
+| Command Line Flag (long) | (short) | Config File                | Environment Variable               |
+| ------------------------ | ------- | -------------------------- | ---------------------------------- |
+| postgres                 | P       | postgres-conenction-string | INDEXER_POSTGRES_CONNECTION_STRING |
+| pidfile                  |         | pidfile                    | INDEXER_PIDFILE                    |
+| algod                    | d       | algod-data-dir             | INDEXER_ALGOD_DATA_DIR             |
```
brianolson, line 119: this could be just ALGORAND_DATA same as for `algod`,
and I see that code is still in there, so maybe this is just a doc change?

cmd/algorand-indexer/daemon.go
brianolson: -v/--verbose is a lovely convention

cmd/algorand-indexer/daemon.go
brianolson: util.BindFlags() is never called any other way, the string can just live inside that function

README.md
```diff
@@ -107,6 +108,29 @@ When `--token your-token` is provided, an authentication header is required. For
...
+
+### daemon mode
+
+```
+./algorand-indexer daemon [options]
+```
+
+| Command Line Flag (long) | (short) | Config File                | Environment Variable               |
+| ------------------------ | ------- | -------------------------- | ---------------------------------- |
+| postgres                 | P       | postgres-conenction-string | INDEXER_POSTGRES_CONNECTION_STRING |
```
brianolson, line 117: I'm looking at this and viper.RegisterAlias() and util.BindFlags() and wondering if f.Name might still just be "postgres" and this might build "INDEXER_POSTGRES"

cmd/algorand-indexer/daemon.go
winder: This isn't really new, there were 2 init functions that I merged together.

I'm going to improve the logger situation in #179

util/config.go
brianolson: Seems premature for one function. I think it's fine where it is.

README.md
winder: Yeah I wasn't sure how to deal with that. I didn't want to remove it so currently it's acting as the final fallback if it wasn't provided anywhere else.

brianolson: In my usage I prefer to set ALGORAND_DATA once and have it work for both `goal` and `algorand-indexer`.
Probably in a proper production system it should always be explicitly set, but this way is handy for development.

cmd/algorand-indexer/daemon.go
rotemh: I don't understand what it has to do with being called by another function. Strings, especially if they are used more than once, need to be constants. It makes it easier to read and to change.

README.md
winder: You're right, this worked for the config file but isn't working properly for the environment variable. I'll do some debugging.

cmd/algorand-indexer/daemon.go
brianolson: I'm saying we could change all the calls to `util.BindFlags(cmd)` and "INDEXER" only exists inside the function and passes the don't-repeat-yourself principle, it can't get out of sync in different places.

README.md
winder: The bug I saw was something else. For some reason viper doesn't swap - and _ automatically so I needed `viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))` as part of the config.

You're still correct that in Bindflags f.Name is the cobra name (since the alias isn't part of cobra). This code works because the alias isn't resolved until the value is looked up, in this case `viper.IsSet`. So `viper.IsSet("postgres")` and `viper.IsSet("postgres-connection-string")` both fetch the same value.

cmd/algorand-indexer/daemon.go
winder: Thanks, great suggestions. It should totally be a constant and I got a little ahead of myself when making the BindFlags function.

util/config.go
winder: I put the constant in here, so now there are 2 things. `config.EnvPrefix` looks a lot nicer than `util.EnvPrefix` so I went ahead and moved it.

cmd/algorand-indexer/main.go
winder: This is definitely a subjective one. I actually checked whether `indexer.config` was accepted before making the PR, which might have been nice. But it isn't.

I think having a yaml extension makes it pretty obvious that this is some sort of configuration. Since we don't have a specific data directory for indexer, I think that this implicit config file approach needs to be the service name. If we didn't, if there was a file named `config` in the home directory indexer would try to load it.

I don't feel strongly about any of these, but they are all reasonable options:
* indexer
* indexer-config
* algorand-indexer
* algorand-indexer-config (this one is maybe a bit much :laughing:)

README.md
winder: Yeah that seems reasonable, I'm not planning to remove ALGORAND_DATA

brianolson (approved):

PR #203, opened 2020-09-22
Add migration status to /health endpoint

winder:
Updates the /health endpoint with additional metadata relating to migrations.
Adds Health function to IndexerDb to provide status about internal migrations.
Adds a generic migration runner which is more testable (and tests for it).
Add some basic tests to the postgres backend to make sure migrations are attempted.


idb/postgres_migrations.go
```diff
@@ -59,38 +73,54 @@ func (db *PostgresIndexerDb) runAvailableMigrations(migrationStateJson string) (
...
-	} else {
-		log.Print("database migrations running in background...")
-		go db.runMigrationState(&state, false)
-		return nil
-	}
-}
+	abortChan := make(chan error)
+	blockChan := make(chan struct{})
 
-func (db *PostgresIndexerDb) runMigrationState(state *MigrationState, blocking bool) (err error) {
```
winder, line 76: The diff view probably wont be very useful with this function, probably best to look at just the new task queue I've implemented.

idb/migration/migration.go
```diff
@@ -0,0 +1,63 @@
...
+	Handler Handler
+
+	// The system should wait for this migration to finish before trying to import new data or serve data.
+	PreventStartup bool
+
+	// Description of the migration
+	Description string
+}
+
+func RunMigrations(errChan chan <-error, statusChan chan <-string, doneChan chan <-struct{}, unblockChan chan <-struct{}, migrations ...Task) {
```
winder, line 22: Per discussion with @algobolson I'm going to do some more experimenting with this function. The granular channels are great for testing but make it cumbersome to use.

idb/postgres_migrations.go
```diff
@@ -59,38 +64,35 @@ func (db *PostgresIndexerDb) runAvailableMigrations(migrationStateJson string) (
...
-		}
-		state.NextMigration = nextMigration
-		if blocking && !anyBlockers(nextMigration) {
-			log.Print("database migrations will complete in background...")
-			go db.runMigrationState(state, false)
-			return nil
-		}
 	}
+
+	// Add a task to mark migrations as done instead of using a channel.
```
winder, line 79: I thought this was cool, but maybe it's too clever? Changing start to `func Start(onCompletion func() error)` would also be a nice way to trigger a final routine:
```
db.migration.Start(db.markMigrationsAsDone)
```

idb/postgres/postgres_migrations.go
```diff
@@ -31,24 +32,28 @@ type MigrationState struct {
...
+type postgresMigrationFunc func(*PostgresIndexerDb, *MigrationState) error
 
 type migrationStruct struct {
-	migrate migrationFunc
+	migrate postgresMigrationFunc
 
-	// The system should wait for this migration to finish before trying to import new data or serve data.
-	preventStartup bool
+	// Description of the migration
+	description string
```
brianolson, line 41: still need to bring back the preventStartup functionality as prevent-api-access or prevent-db-access

brianolson: We don't need to block access if _any_ migration is running, just some of them. Some of them were extensively designed to run in background for while the system is available.

idb/postgres/postgres.go
```diff
@@ -31,17 +33,32 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
+
+	if optsCpy == nil {
+		optsCpy = &idb.IndexerDbOptions{}
+	}
+
+	if strings.Contains(connection, "readonly") {
+		optsCpy.ReadOnly = true
```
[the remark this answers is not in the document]
winder: I was mainly trying to initialize it to the default options in case nil was passed in. I'll revert back to the original code.

idb/postgres/postgres_migrations.go
winder: It should be straight forward to add that feature back in, so I'll do that. But I'm not sure if it should be enabled. If we're changing things in the DB, wouldn't we see inconsistent results?

I guess I'm not sure how to judge how "bad" it would be to knowingly return the wrong TxID or AssetConfig object.

You got me thinking about whats actually going on in the background, and I think I may have introduced a bug where the importer will now start importing earlier than before. It definitely needs that "migration-is-blocking" flag.

brianolson: it might be possible to have two attributes of a migration (prevent API, prevent import), but right now it's probably simplest to count them as the same; OR, prevent-api can be a per-migration attribute and any pending or active migration could prevent import. I think preventing import is less of a problem than preventing API access. With the txid migration query-by-txid wouldn't work (but it wouldn't work before the migration either) but other queries would work fine, so disabling the API during that migration would have prevented otherwise normal operation of the indexer. The txid migration would also be safe to run while import is operating. But, unless we carefully add both attributes to each migration, right now I'd vote for having a prevent-API flag and preventing import if there are any running migrations. And the prevent-API flag blocks if any active _or pending_ migration has it set, but after you're past the last one with it set having other migrations doesn't prevent the API from running.

winder: How about this... the migration process has 2 attributes about the progress:
* Migration Is Running
* DB Unavailable (new name for preventStartup)

We add a new CLI argument / config value to toggle whether or not queries are allowed during a migration.

brianolson: If a migration knows it is safe to run queries while the migration is running (e.g. the txid fixup), why add an option to override that and not do queries anyway?

winder: The DB is technically in an inconsistent state if a migration is running, we're returning data which we know may not be accurate. I don't think we can say for certain that it's safe to return this data since it depends on the application and user. I'm imagining a Tarantino movie with a high stakes transaction that they're checking on and it doesn't get returned because the txid was wrong, that crypto accountant really hopes their indexer admin set the more conservative "block-queries-during-migration: true" setting.

brianolson: Ok, a super-cautious mode makes sense to me now, thanks. I was thinking in terms of restarting `algorand-indexer` with a new version and it starts migrations, it could serve data right away and be no more wrong than it was a minute before the restart, but it could also be super cautious and say, "I know some things were wrong, I refuse to serve anything until everything is as correct as possible". It's probably even right to make super-cautious the default and lesser best-effort-but-more-available be the option.

brianolson (approved):

idb/migration/migration.go
```diff
@@ -0,0 +1,149 @@
...
+	// DBUnavailable indicates whether or not queries should be blocked until this task is executed. If there are
+	// multiple migrations the migration status should indicate that queries are blocked until all blocking tasks
+	// have completed.
+	DBUnavailable bool
+
+	// Description should be a human readable description of what the migration does.
+	Description string
+}
+
+type State struct {
```
algorotem, line 36: Please comment exported structs and fields

idb/migration/migration.go
```diff
@@ -0,0 +1,149 @@
...
+		if task.MigrationId > m.blockUntil {
+			blocking = false
+		}
+
+		m.update(nil, StatusActivePrefix+task.Description, true, blocking)
+		err := task.Handler()
+
+		if err != nil {
+			err := fmt.Errorf("%s%d (%s): %v", StatusErrorPrefix, task.MigrationId, task.Description, err)
+			// TODO: If a migration failed, should we block queries?
```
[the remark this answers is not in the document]
brianolson: I think there are two options if a migration fails: exit(1) or serve un-migrated data with warnings in /health
If the failed migration and all after it are non-blocking migrations then it might be possible to serve un-migrated data, but I kinda think the fact that a migration failed at all shows there are potentially deeper bugs and we should just log everything and exit(1)

idb/migration/migration.go
```diff
@@ -0,0 +1,114 @@
...
+type Migration struct {
+	mutex sync.Mutex
+
+	state State
+	tasks []Task
+}
+
+// Broken out to allow for testing.
+func (m *Migration) setTasks(migrationTasks []Task) error {
+	set := make(map[int]bool)
```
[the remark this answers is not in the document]
winder: It's for error checking to detect duplicate migration IDs, I'll rename the variable to ids

idb/migration/migration.go
```diff
@@ -0,0 +1,114 @@
...
+type State struct {
+	Err     error
+	Status  string
+	Running bool
+}
+
+type Migration struct {
+	mutex sync.Mutex
+
+	state State
```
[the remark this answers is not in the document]
winder: I'll embed the State object, it's also used as a return type so I need to keep those values grouped.

idb/migration/migration.go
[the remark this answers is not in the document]
winder: The current behavior will leave indexer disabled, but still serving /health with the error message. Sounds like we're all in agreement here.

idb/migration/migration.go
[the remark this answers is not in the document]
winder: Actually, embedding wont work. It would cause the state values to be accessed without being locked.

I think that this is going to be the cleanest implementation, the alternative would be for "GetStatus" to return 4 values but I prefer to have them wrapped up in an object in case we update the state (i.e. in the future "progress" would be nice)

idb/migration/migration.go
```diff
@@ -0,0 +1,114 @@
...
+	defer m.mutex.Unlock()
+
+	return State{
+		Err:     m.state.Err,
+		Status:  m.state.Status,
+		Running: m.state.Running,
+	}
+}
+
+func (m *Migration) update(err error, status string, running bool) {
```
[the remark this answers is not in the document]
winder: discussion in other comment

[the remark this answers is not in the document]
algorotem: The method still applies only to `State`. Why is it on Migration? (I couldn't understand it from the other comment)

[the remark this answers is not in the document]
winder: The main use for this function is the mutex which is only part of `Migration`. Anything that wants to read/write a State object without the concurrency would just do it without a helper function.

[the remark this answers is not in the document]
algorotem: If the only sensitive part is the state, consider moving the mutex to the `State` struct and also have a `GetStatus` method on `State` that returns a copy of itself. That way you can certainly assert that concurrent access will be protected.

algorotem (approved):

PR #206, opened 2020-09-24
Move the import logic into the import package.

winder:
There should be no functional changes here. I wanted to be able to call the import function to initialize a DB and that wasn't really possible.

It hopefully makes the code slightly easier to read:
* clarifies an unexpected dependency - daemon calls "updateAccounting" defined in the import command.
* the "db" variable is defined in main is shadowed in import function variables / parameters.
* protoJsonPath is declared but not used.


brianolson (approved):
LGTM 👍

PR #208, opened 2020-09-28
Make server migration aware.

winder:
This is a companion PR to the health check upgrades.

It is split out so that the two PRs can be reviewed in isolation.


api/middlewares/migration_middleware.go
```diff
@@ -0,0 +1,42 @@
...
+
+// handler returns a 500 if the IndexerDb is migrating.
+func (mm *MigrationMiddleware) handler(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(ctx echo.Context) error {
+		h, err := mm.idb.Health()
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Indexer health error: %s", err.Error())
+		}
+
+		if !h.IsMigrating {
```
brianolson, line 36: I think this would read clearer as `if migrating{return err}` but it's not super important

idb/postgres/postgres.go
```diff
@@ -2305,26 +2306,33 @@ func (db *PostgresIndexerDb) yieldApplicationsThread(ctx context.Context, rows *
...
 
-	var data = make(map[string]interface{})
-	if state.Err != nil {
-		data["migration-error"] = state.Err.Error()
-	}
-	if state.Status != "" {
-		data["migration-status"] = state.Status
-	}
+	// If we are not in read-only mode, there will be a migration object.
+	if db == nil {
```
brianolson, line 2313: `db.migration == nil` ?

brianolson:
Where's the part where this lets /health through but blocks everything else?

api/server.go
```diff
@@ -29,18 +29,20 @@ func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, log *log.Log
...
 	}
 
 	api := ServerImplementation{
 		EnableAddressSearchRoundRewind: developerMode,
 		db:                             db,
 	}
 
-	generated.RegisterHandlers(e, &api, maybeAuth...)
+	generated.RegisterHandlers(e, &api, middleware...)
 	common.RegisterHandlers(e, &api)
```
winder, line 46: @brianolson "common" is just the root endpoints, which for indexer is just /health. So I'm just not attaching the migration middleware here to allow the health endpoint to run regardless of the migration state.

idb/postgres/postgres_migrations.go
```diff
@@ -96,8 +96,7 @@ func (db *PostgresIndexerDb) runAvailableMigrations(migrationStateJson string) (
 		return err
 	}
 
-	// TODO: Start it in the background instead of blocking
-	db.migration.Start()
+	go db.migration.Start()
```
algorotem, line 99: The caller shouldn't be aware of the internals of `Migration`. Consider moving the content of `Start()` to an internal function to `Migrations` and run this it as a goroutine from `Start()`. As an alternative, if you want the caller to decide whether to run it as blocking or not, consider changing the name of `Start` to `RunMigrations` (see the comment in the other PR).

idb/postgres/postgres.go
```diff
@@ -334,6 +334,7 @@ func (db *PostgresIndexerDb) SetMetastate(key, jsonStrValue string) (err error)
 }
 
 func (db *PostgresIndexerDb) GetMaxRound() (round uint64, err error) {
+	round = 0
```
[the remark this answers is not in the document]
winder: The mock sql.Driver stuff is a little rough around the edges. This function happens to get called as part of the Health tests and I wanted to make sure the round that gets returned is 0 when there's an error to simplify the testing.

idb/postgres/postgres_test.go
```diff
@@ -29,12 +30,17 @@ func TestAllMigrations(t *testing.T) {
 			})
 
 			// This automatically runs migraions
+			//var pdb idb.IndexerDb
```
brianolson, line 33: dead code?

idb/postgres/postgres_migrations.go
winder: Thanks, I'll rename the function. I either need a "done" channel or to have this function block.

algorotem (approved):

Issue #166, opened 2020-08-01
Rewards are always 0


fabrice102:
The `rewards` field of each account is always 0 even when rewards were received:

Concretely
```bash
$ curl -s "localhost:8980/v2/accounts/SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE" | jq
```
returns
```json
{
  "account": {
    "address": "SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE",
    "amount": 882040798,
    "amount-without-pending-rewards": 882040798,
    "assets": [
      {
        "amount": 1844000079218796,
        "asset-id": 2512768,
        "creator": "",
        "is-frozen": false
      }
    ],
    "created-assets": [
      {
        "index": 2512768,
        "params": {
          "clawback": "SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE",
          "creator": "SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE",
          "decimals": 6,
          "default-frozen": false,
          "freeze": "SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE",
          "manager": "SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE",
          "metadata-hash": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
          "name": "YouNow Pending Props",
          "reserve": "SN3FSQVLLNWSI3I7QCMQEQN6L42OW5RP4FNM6BNOZHBRKHODSQYU4MUPKE",
          "total": 1844674407370955,
          "unit-name": "xPROPS",
          "url": "https://propsproject.com/"
        }
      }
    ],
    "pending-rewards": 0,
    "reward-base": 134204,
    "rewards": 0,
    "round": 8208842,
    "sig-type": "sig",
    "status": "Offline"
  },
  "current-round": 8208842
}
```

brianolson:
Indexer has made no attempt to track 'the sum of all rewards for an account'. This can be a feature request for the future.

fabrice102:
[...]

PR #221, opened 2020-10-07
Initial rewards implementation. #166

winder:
Adds cumulative rewards:
- [x] rewind (account at round)
- [x] migration
- [x] set reward in query results
- [x] tests

Adds a new column in postgres to store cumulative rewards.

New test script to load a pg_dump, run migrations, then verify with test queries that the migration worked.


accounting/accounting.go
```diff
@@ -94,11 +94,19 @@ func bytesAreZero(b []byte) bool {
...
+func (accounting *AccountingState) updateAlgo(addr types.Address, amount int64) {
+	accounting.updateAlgoAndRewards(addr, amount, 0)
+}
+
+func (accounting *AccountingState) updateAlgoAndRewards(addr types.Address, amount, rewards int64) {
 	if accounting.AlgoUpdates == nil {
-		accounting.AlgoUpdates = make(map[[32]byte]int64)
+		accounting.AlgoUpdates = make(map[[32]byte]*idb.AlgoUpdate)
+	}
+	if _, hasKey := accounting.AlgoUpdates[addr]; !hasKey {
```
brianolson, line 105: If we're going to do the lookup, we should use the value if it exists instead of doing 3 lookups and maybe one insert. Since the map value is a pointer type this could be one lookup and maybe one insert.

accounting/accounting.go
```diff
@@ -94,11 +94,19 @@ func bytesAreZero(b []byte) bool {
 	return true
 }
 
-func (accounting *AccountingState) updateAlgo(addr types.Address, d int64) {
+func (accounting *AccountingState) updateAlgo(addr types.Address, amount int64) {
+	accounting.updateAlgoAndRewards(addr, amount, 0)
+}
+
+func (accounting *AccountingState) updateAlgoAndRewards(addr types.Address, amount, rewards int64) {
```
brianolson, line 101: this is a nice pattern, it fits nicely into all the updates below

accounting/accounting.go
winder: Something like this?
```
	var acctUpdate *idb.AlgoUpdate
	var hasKey bool
	if acctUpdate, hasKey = accounting.AlgoUpdates[addr]; !hasKey {
		acctUpdate = &idb.AlgoUpdate{}
		accounting.AlgoUpdates[addr] = acctUpdate
	}
	acctUpdate.Balance = accounting.AlgoUpdates[addr].Balance + amount
	acctUpdate.Rewards = accounting.AlgoUpdates[addr].Rewards + rewards
```

brianolson: yeah but you can just assign to the map with an AlgoUpdate{amount,rewards} and then return.
and `acctUpdate.Balance += amount`

idb/postgres/postgres_migrations.go
```diff
@@ -22,13 +23,33 @@ import (
...
+	// NextRound used for m0 to checkpoint progress.
 	NextRound int64 `json:"round,omitempty"`
 
+	// NextAssetId used for m3 to checkpoint progress.
 	NextAssetId int64 `json:"assetid,omitempty"`
+
+	// NextAccount used for m5 to checkpoint progress.
+	NextAccount []byte `json:"nextaccount,omitempty"`
+
+	// Note: a generic "data" field here could be a good way to deal with this growing over time.
```
brianolson, line 50: I think MigrationState as json with fields only present if used satisfies this?
Also I think it wouldn't be bad for migrations to re-use NextFoo if they wind up iterating over that field similar to how a prior migration did.

accounting/rewind.go
```diff
@@ -71,15 +72,18 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
...
 			if addr == stxn.Txn.CloseRemainderTo {
 				// unwind receiving a close-to
 				acct.AmountWithoutPendingRewards -= uint64(stxn.ClosingAmount)
 				acct.AmountWithoutPendingRewards -= uint64(stxn.CloseRewards)
+				acct.Rewards -= uint64(stxn.CloseRewards)
 			} else if !stxn.Txn.CloseRemainderTo.IsZero() {
 				// unwind sending a close-to
 				acct.AmountWithoutPendingRewards += uint64(stxn.ClosingAmount)
-				acct.AmountWithoutPendingRewards += uint64(stxn.CloseRewards)
+				// TODO: this doesn't seem right. Rewards don't come from the account balance so what is this doing?
```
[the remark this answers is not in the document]
brianolson: No, wait, CloseRewards ... go to the close receiver? Now I'm confused and will need to re-read the impl in algod; but now I'm guessing this is pending rewards on the account that weren't paid out before it was closed. The virtual sequence would be: apply pending rewards, pay amount receiver, everything else to close-to addr.

idb/postgres/postgres_migrations.go
winder: Currently there's no cleanup mechanism, so they keep adding up. That also means we can't reuse a field because there isn't a good way of telling whether the individual migration is just starting, or resuming from where the earlier migration left off.

idb/postgres/postgres_migrations.go
```diff
@@ -257,50 +278,165 @@ func m3acfgFixAsyncInner(db *PostgresIndexerDb, state *MigrationState, assetIds
...
+		if len(accounts) == batchSize {
+			db.log.Println("Processing batch of accounts.")
+			m5accountCumulativeRewardsUpdateAccounts(db, state, accounts)
+			accounts = accounts[:0]
+		}
+	}
+
+	// Get the remainder
+	if len(accounts) > 0 {
+		db.log.Println("Processing final batch of accounts.")
```
brianolson, line 328: other migrations have a way to clear out their temp state when committing the last batch. How can we set `state.NextAccount = nil` here?

accounting/accounting.go
```diff
@@ -243,8 +258,9 @@ func (accounting *AccountingState) AddTransaction(txnr *idb.TxnRow) (err error)
 	accounting.updateAlgo(accounting.feeAddr, int64(stxn.Txn.Fee))
 
 	if stxn.SenderRewards != 0 {
-		accounting.updateAlgo(stxn.Txn.Sender, int64(stxn.SenderRewards))
-		accounting.updateAlgo(accounting.rewardAddr, -int64(stxn.SenderRewards))
+		v := int64(stxn.SenderRewards)
```
algorotem, line 261: Consider giving the variable full names. It makes it difficult to follow.

accounting/accounting.go
```diff
@@ -262,12 +278,14 @@ func (accounting *AccountingState) AddTransaction(txnr *idb.TxnRow) (err error)
...
+			v := int64(stxn.ReceiverRewards)
+			accounting.updateAlgoAndRewards(stxn.Txn.Receiver, v, v)
+			accounting.updateAlgo(accounting.rewardAddr, -v)
 		}
 		if stxn.CloseRewards != 0 {
-			accounting.updateAlgo(stxn.Txn.CloseRemainderTo, int64(stxn.CloseRewards))
-			accounting.updateAlgo(accounting.rewardAddr, -int64(stxn.CloseRewards))
+			v := int64(stxn.CloseRewards)
+			accounting.updateAlgoAndRewards(stxn.Txn.CloseRemainderTo, v, v)
+			accounting.updateAlgo(accounting.rewardAddr, -v)
```
algorotem, line 288: Since it's repeated multiple times, consider making it a function.

accounting/rewind.go
algorotem, line 85: comment just in order to not forget this todo

cmd/algorand-indexer/main.go
```diff
@@ -97,6 +99,11 @@ func globalIndexerDb(opts *idb.IndexerDbOptions) idb.IndexerDb {
 }
 
 func init() {
+	logger = log.New()
+	logger.SetFormatter(&log.JSONFormatter{})
+	logger.SetOutput(os.Stdout)
+	logger.SetLevel(log.InfoLevel)
```
algorotem, line 105: Consider making it a flag (Either in this PR, or ticket it for another one)

idb/postgres/postgres_migrations.go
```diff
@@ -257,50 +278,165 @@ func m3acfgFixAsyncInner(db *PostgresIndexerDb, state *MigrationState, assetIds
...
+		if acct.Error != nil {
+			err := fmt.Errorf("%s: problem querying accounts: %v", cumulativeRewardsUpdateErr, acct.Error)
+			db.log.Errorln(err.Error())
+			return err
+		}
+
+		accounts = append(accounts, acct.Account.Address)
+
+		if len(accounts) == batchSize {
+			db.log.Println("Processing batch of accounts.")
```
algorotem, line 320: nit - consider logging the batch number (e.g. 5/2303) in order to get meaningful data and to avoid a long list of "Processing batch of accounts."

algorotem (commented):
Looks good overall. A few small comments. 
Also, It seems like all the modifications to the logging should probably be in another PR.

cmd/algorand-indexer/main.go
winder: Definitely. This is a baby step towards #179

accounting/accounting.go
winder: I normally agree with that, but here I intentionally kept the short variable to make it easier to see that it's being repeated in several places. Your other suggestion to make a function which avoids repeating the value is probably a better solution.

accounting/rewind.go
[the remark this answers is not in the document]
winder: I haven't verified this. If there are "CloseRewards" then the CloseToAddress must have been set and the Sender account needs to be empty at the end of the transaction. Therefore "SenderRewards" had better be equal to zero, I'm having a hard time coming up with a scenario where CloseRewards and SenderRewards are both non zero.

I'm pretty convinced that this was a bug and removing the line is correct. Rewards always originate from the RewardsPool. Therefore when rewinding transactions rewards should never be added to an account other than the RewardsPool. Since we shouldn't be rewinding the RewardPool account we should never be adding CloseRewards/ReceiverRewards/SenderRewards to an account in this function.

accounting/accounting.go
winder: Done

idb/postgres/postgres_migrations.go
winder: Fixed, thanks

idb/postgres/postgres_migrations.go
winder: Looks like I might have been wrong about that comment. Added the cleanup code to my migration so this will continue to be an option in the future.

accounting/rewind.go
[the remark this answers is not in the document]
brianolson: Looking at algod payment.go and eval.go, CloseRewards is what is accrued on the CloseRemainderTo account which gets added in at this time just because the account is being touched. It shouldn't go to the Sender, that was a bug. ClosingAmount is everything left over after normal payment Amount and Fee.
Now the subtle bug might be that I think `ClosingAmount = previousBalance + SenderRewards - Fee - Amount` and we might be double-rewinding SenderRewards by doing it separately _and_ as ClosingAmount.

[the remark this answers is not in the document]
brianolson: Looking again again I think the Sender previous balance is okay with a closeremainderto. algebra on the previous comment:
`sender.previousBalance = +ClosingAmount + Fee + Amount - SenderRewards`
and I see all those pieces here.

brianolson (approved):

PR #225, opened 2020-10-16
Fix golint

winder:


brianolson (approved):
Thanks, Will. Hopefully this big change will help things stay tidy in the future.

PR #227, opened 2020-10-21
Switch to algod v2 client

winder:


fetcher/fetcher.go
```diff
@@ -120,7 +123,7 @@ func (bot *fetcherImpl) followLoop() {
 		if err != nil {
 			return
 		}
-		err = bot.handleBlockBytes(blockbytes)
+		err = bot.handleBlockBytes(msgpack.Encode(block))
```
winder, line 126: The SDK discards the certificate, so this only works because we aren't using it.

brianolson: If the Go v2 client doesn't have a BlockRaw function we should fix that instead of re-encoding here.

brianolson: The Go v2 client is even fetching raw msgpack bytes, just hasn't exposed an API to return those bytes and not unpack them. Should be a pretty simple addition.

brianolson (approved):

PR #229, opened 2020-10-27
Config file updates.

winder:
I noticed that there were strange errors when launching indexer from the CWD, it turns out the config file library was trying to read the binary as a config file.

This change renames the base config filename from `algorand-indexer` to `indexer`, and tweaks console output.

### DIFF
--- a/cmd/algorand-indexer/main.go
+++ b/cmd/algorand-indexer/main.go
@@ -173,10 +173,7 @@ func configureLogger() error {
 
 func main() {
 	if err := rootCmd.Execute(); err != nil {
-<<<<<<< HEAD
 		logger.WithError(err).Error("an error occurred running indexer")
-=======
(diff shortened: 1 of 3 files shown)


cmd/algorand-indexer/main.go
```diff
@@ -125,21 +125,28 @@ func init() {
 
 	if err := viper.ReadInConfig(); err != nil {
 		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
-			// Config file not found; the error message indicates locations where we look
-			fmt.Println(err.Error())
+			// Config file not found. Returns an error string like:
```
brianolson, line 128: replace this entire comment block with effectively, "It's okay not to have a config file, maybe everything was set in command line options"

brianolson (approved):
one comment cleanup idea; otherwise good

PR #230, opened 2020-10-27
Add `/openapi` endpoint

winder:
Add `/openapi` endpoint for fetching the spec from indexer.

The code generation is split up between the root and v2 endpoints, so this endpoint uses a path parameter to select which spec to retrieve:
```
/openapi/common
/openapi/v2
```


algorotem:
is there an option to unify them? I believe the user doesn't really need to know about the split.

winder:
@algorotem I'll need to think about that a bit. I have a couple ideas that I'll need to try out the next time I work on this.

winder:
This isn't a priority right now, closing.

PR #233, opened 2020-11-04
Error when rewinding special accounts.

winder:
Add a check to the AccountAtRound function to avoid rewinding special accounts.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -643,13 +643,11 @@ func (si *ServerImplementation) fetchAccounts(ctx context.Context, options idb.A
 				_, isSpecialAccountRewindError := err.(*accounting.SpecialAccountRewindError)
 				if len(options.EqualToAddress) != 0 || !isSpecialAccountRewindError {
 					return nil, fmt.Errorf("%s: %v", errRewindingAccount, err)
-
 				}
+				// If we didn't return, continue to the next account
+				continue
 			}
-			// Because the error is a conditional return, double check that there was no error.
-			if err == nil {
-				account = acct
-			}
+			account = acct
 		} else {
 			account = row.Account
 		}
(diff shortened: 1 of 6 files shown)


api/handlers.go
```diff
@@ -583,7 +583,7 @@ func (si *ServerImplementation) fetchBlock(round uint64) (generated.Block, error
 	}
 
 	rewards := generated.BlockRewards{
-		FeeSink:                 "",
+		FeeSink:                 blk.FeeSink.String(),
```
winder, line 586: oops.

winder:
@brianolson 
This introduces a bug with the dev mode accounts `bulk-rewind` endpoint. These special accounts may be included in the `/v2/accounts` search endpoint, which makes it very easy to accidentally rewind a special account.

I'm thinking it would be good to filter special accounts from `/v2/accounts` somewhere.

brianolson:
Not listing the Fee and Rewards accounts in a /v2/accounts listing could make sense. Might be best to have them filtered out by default with a `?include-special=1` option. Or maybe only filter them out of the list when `?round=N` is set?

idb/postgres/postgres.go
```diff
@@ -2396,6 +2402,39 @@ func (db *IndexerDb) Health() (health idb.Health, err error) {
 	}, err
 }
 
+// GetSpecialAccounts is part of idb.IndexerDB
+func (db *IndexerDb) GetSpecialAccounts() (accounts idb.SpecialAccounts, err error) {
+	var cache string
```
brianolson, line 2407: I'd be tempted to not cache this in metastate but cache it in RAM in the IndexerDb object.

winder: I'm actually doing both, it's cached in ram in rewind.go, so GetSpecialAccounts() should only be called once.

I wanted to implement that outside of IndexerDb so that alternative implementations wouldn't need to worry about that optimization.

brianolson (approved):

PR #234, opened 2020-11-05
Naming cleanup and append to log file.

winder:
Noticed some weird double negatives when using "db-unavailable" instead of "db-available".

Also noticed that the log file wasn't being opened in append mode.


api/indexer.oas3.yml
```diff
@@ -1041,7 +1041,7 @@
             "properties": {},
             "type": "object"
           },
-          "db-unavailable": {
+          "db-available": {
```
brianolson, line 1044: was `db-unavailable` released? Will anyone be sad if it goes away?

brianolson: db-UNavailable has the advantage that it can be default-false and not present in the json in the default state. It can be `{'db-unavailable':true}` or just not there at all.

winder: It hasn't been released yet, so this is just in time.

A lot of the public fields (including this one) is marked to serialize regardless of whether or not it's a zero value, but that's an interesting point that I usually don't think about.

PR #235, opened 2020-11-05
IndexerDb should block importer until db is available.

winder:
Block the importer from starting until IndexerDb health reports the DB is available.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -79,6 +79,7 @@ var daemonCmd = &cobra.Command{
 			maybeFail(err, "import proto, %v", err)
 		}
 		if bot != nil {
+			logger.Info("Initializing block import handler.")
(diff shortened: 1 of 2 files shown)


cmd/algorand-indexer/daemon.go
```diff
@@ -107,6 +112,35 @@ var daemonCmd = &cobra.Command{
 	},
 }
 
+// waitForDBAvailable wait for the IndexerDb to report that it is available.
+func waitForDBAvailable(db idb.IndexerDb) {
```
brianolson, line 116: I think this is a race between the migration starting and this checking to see if a migration is running

winder: This should be OK. If there is going to be a migration it is initialized in the init function BEFORE starting the migration in a separate go routine. The `db.migration` object is initialized with `blocking = true` to ensure there's no race condition where it incorrectly unblocks..

PR #236, opened 2020-11-05
Better progress for cumulative rewards migration

winder:


idb/postgres/postgres_migrations.go
```diff
@@ -287,6 +287,27 @@ func m4cumulativeRewardsDBUpdate(db *IndexerDb, state *MigrationState) error {
...
+	}
+	return i
+}
+
+func addrToPercent(addr string) float64 {
+	if len(addr) < 3 {
+		return 0.0
+	}
+
+	pt1 := float64(b32ToIndex(addr[0])) / 32.0 * 100.0
```
brianolson, line 304: I usually see things added up in int.
```
val := b32ToIndex(addr[0])
val = (val * 32) + b32ToIndex(addr[1])
val = (val * 32) + b32ToIndex(addr[2])
return float64(val) / (32*32*32)
```

winder: Nice, I knew there must be a better way

brianolson (approved):

PR #240, opened 2020-11-09
Parameterize STAGING in build task

btoll:
Since our staging bucket is private, let's not hardcode it.

### DIFF
--- a/mule.yaml
+++ b/mule.yaml
@@ -20,7 +20,7 @@ tasks:
   - task: s3.BucketCopy
     name: stage
     src: $HOME/projects/indexer/tmp/node_pkgs/linux/amd64/$VERSION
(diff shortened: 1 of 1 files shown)


onetechnical (approved):

winder (commented):
Does mule ensure that environment variables are set? Seems like a good change.

btoll:
> Does mule ensure that environment variables are set? Seems like a good change.

It depends on the task, but generally, yes.

Most `mule` tasks call targets that invoke a shell script, so those will check for needed env vars.  Others, like the one in question here, call an internal library function to upload to S3 using boto3, so that will expect the env var to be set in the calling environment.

Here is how I did the upload to staging:
```
STAGING=algorand-staging VERSION=2.2.0-beta mule -f mule.yaml stage
```

I'll add all of this to the documents.

PR #241, opened 2020-11-09
CumulativeRewards account close handling

winder:
Added handling for resetting cumulative rewards when an account is closed.


accounting/accounting.go
```diff
@@ -97,6 +97,28 @@ func bytesAreZero(b []byte) bool {
 	return true
 }
 
+func (accounting *State) closeAccount(addr types.Address) {
+	if accounting.AlgoUpdates == nil {
```
brianolson, line 101: This should be impossible because something earlier in the payment txn must have set it

accounting/accounting.go
```diff
@@ -97,6 +97,28 @@ func bytesAreZero(b []byte) bool {
...
+		accounting.AlgoUpdates = make(map[[32]byte]*idb.AlgoUpdate)
+	}
+
+	update, hasKey := accounting.AlgoUpdates[addr]
+
+	if !hasKey {
```
brianolson, line 107: this should also never happen because the account should have been touched earlier in the "pay" block of AddTransaction()

accounting/accounting.go
```diff
@@ -97,6 +97,28 @@ func bytesAreZero(b []byte) bool {
...
+			Balance: 0,
+			Rewards: 0,
+		}
+		accounting.AlgoUpdates[addr] = update
+		return
+	}
+
+	// If the key was there, override the rewards by setting to zero.
+	// The balance will be updated with the delta as usual.
+	update.Rewards = 0
```
brianolson, line 118: I don't think we should muck with this, I'm _pretty sure_ stxn.SenderRewards will be zero, but it could also be possible that `(previous balance - Fee + SenderRewards - Amount - ClosingAmount) == 0`

idb/postgres/postgres.go
```diff
@@ -733,10 +733,25 @@ func (db *IndexerDb) CommitRoundAccounting(updates idb.RoundUpdates, round, rewa
 			return fmt.Errorf("prepare update algo, %v", err)
 		}
 		defer setalgo.Close()
+
+		// If the account is closing the cumulative rewards field needs to be specially overwritten.
+		resetalgo, err := tx.Prepare(`INSERT INTO account (addr, microalgos, rewardsbase, rewardsTotal) VALUES ($1, $2, $3, $4) ON CONFLICT (addr) DO UPDATE SET microalgos = account.microalgos + EXCLUDED.microalgos, rewardsbase = EXCLUDED.rewardsbase, rewardsTotal = EXCLUDED.rewardsTotal`)
```
brianolson, line 738: $3 and $4 could just be 0
We can leave $2 to ensure our accounting is right.

accounting/accounting.go
winder: This is required to handle the different edge cases.

Suppose 3 transactions evaluated in the following order
B closeTo A, closeToRewards = 100
A closes, [cumulative] rewards must be reset to 0
C closeTo A, closeToRewards = 100

At the end of all this A needs to have rewards = 100

idb/postgres/postgres.go
winder: I'm going to fuss with this a bit in another PR which adds in createdAt/closedAt fields.

Specifically we don't need the `ON CONFLICT` bit. If an account is closing itself it must have been created already so no need for the `INSERT` part of the query.

Aside from that, I think we still need to set all of these fields in case the account was reopened in the same round that it was closed.

brianolson: I was wrong, rewardsbase should always be set to the current round, not zero.
The order of the txns should take care of the close+reopen case?
1. something to A, A gets receiver rewards, rewardsbase set to current round
2. A closes, rewardsTotal = 0, (so I still think $4 can just be 0 as applied to resetalgo)
3. something else to A, A gets no receiver rewards because rewards base is the current round, but that works the same as any other txn.

brianolson: Ohhhhkay, the complication is that a round worth of txn is being aggregated in accounting.go into an AlgoUpdate. AlgoUpdate fields are normally deltas, but sometimes we need to zero-out totalRewards instead of add a delta to it. Algo balances can still just work by deltas (and validations should prove that the Algo balance goes to zero as appropriate). Okay, I get it, this is all correct (there will never be non-zero rewards within a round where an account is closed and reopened, but the math isn't wrong to allow for that).

idb/dummy.go
```diff
@@ -476,6 +476,7 @@ type TxnAssetUpdate struct {
 type AlgoUpdate struct {
 	Balance int64
 	Rewards int64
+	Closed  bool
```
brianolson, line 479: `// Closed changes the nature of the Rewards field. Balance and Rewards are normally deltas added to the microalgos and totalRewards columns, but if an account has been Closed then Rewards becomes a new value that replaces the old value (always zero by current reward logic)`

idb/postgres/postgres.go
winder: I think that there 1 scenario where there would be non-zero rewards, if things work the way I think they do.

If account A is re-opened with another account closing into A, then the rewards which would have gone to the closing account would instead go to A. I didn't actually test this scenario...

brianolson: I think if another account is closed onto A, rewards to that other account are applied there first and then the balance is transferred to A. The 'Close Rewards' are what the close-to destination gets just because it's being touched.

idb/dummy.go
winder: done

accounting/accounting.go
winder: I agree, but prefer to keep all the edge cases handled. Maybe in the future `Clear` can re-initialize to empty maps.

Issue #246, opened 2020-11-12
In 2.2.0 - GET v2/accounts now takes 2+ minutes to complete in MainNet


purestaketdb:
In previous version, the GET v2/accounts endpoint for MainNet takes under 400ms to complete using Postman. After upgrade to 2.2.0 in our testing environment the endpoint ceased to return results, exceeding our API gateway timeout of 30seconds. Further research, moving up the stack to the indexer reader node had curl requests timing out around the 2 minute mark and finally direct sql queries on the database showed a result around the 2-3 minute mark. 

This is dramatically slower and sub-usable in production. Below is the EXPLAIN ANALYZE capture 

```SQL                                                           

QUERY PLAN                                                                           
---------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=7090210.28..7090210.29 rows=2 width=402) (actual time=119169.202..119169.202 rows=2 loops=1)
   CTE qaccounts
     ->  Index Scan using account_pkey on account a  (cost=0.56..705231.33 rows=6756985 width=259) (actual time=0.048..26080.648 rows=6758410 loops=1)
   CTE qaa
     ->  GroupAggregate  (cost=1931065.23..2019160.34 rows=200 width=128) (actual time=45904.839..56407.660 rows=6439212 loops=1)
           Group Key: xa.addr
           ->  Sort  (cost=1931065.23..1948683.55 rows=7047329 width=46) (actual time=45904.817..49089.458 rows=6460461 loops=1)
                 Sort Key: xa.addr
                 Sort Method: external merge  Disk: 410568kB
                 ->  Hash Join  (cost=273544.46..695899.29 rows=7047329 width=46) (actual time=2721.642..39067.379 rows=6460461 loops=1)
                       Hash Cond: (xa.addr = aa.addr)
                       ->  CTE Scan on qaccounts xa  (cost=0.00..135139.70 rows=6756985 width=32) (actual time=0.001..28466.629 rows=6758410 loops=1)
                       ->  Hash  (cost=135986.98..135986.98 rows=6461398 width=47) (actual time=2716.636..2716.636 rows=6460461 loops=1)
                             Buckets: 65536  Batches: 256  Memory Usage: 2471kB
                             ->  Seq Scan on account_asset aa  (cost=0.00..135986.98 rows=6461398 width=47) (actual time=0.034..1111.861 rows=6460461 loops=1)
   CTE qap
[...]

PR #247, opened 2020-11-12
Acct query fix

brianolson:
fix for #246 
move LIMIT inside an earlier clause to fix postgres query handling
for future bad queries, log any SQL that takes longer than 1 second to run
change vs 2.2.0 branch

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -1520,7 +1520,6 @@ var statusStrings = []string{"Offline", "Online", "NotParticipating"}
 const offlineStatusIdx = 0
 
 func (db *PostgresIndexerDb) yieldAccountsThread(req *getAccountsRequest) {
-	//ctx context.Context, opts idb.AccountQueryOptions, rows *sql.Rows, tx *sql.Tx, blockheader types.Block, out chan<- idb.AccountRow) {
 	defer req.tx.Rollback()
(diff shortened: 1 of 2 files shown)


idb/postgres/postgres.go
```diff
@@ -1518,10 +1519,17 @@ var statusStrings = []string{"Offline", "Online", "NotParticipating"}
...
-	defer tx.Rollback()
+func (db *PostgresIndexerDb) yieldAccountsThread(req *getAccountsRequest) {
+	defer req.tx.Rollback()
 	count := uint64(0)
-	for rows.Next() {
+	defer func() {
+		end := time.Now()
+		dt := end.Sub(req.start)
+		if dt > (1 * time.Second) {
+			log.Printf("long query %fs: %s", dt.Seconds(), req.query)
```
winder, line 1529: I think this should be at least a `log.Warn`. Printf ends up at the `Info` level.

winder: Did you fork master? This is missing some commits for develop. I renamed `postgres.PostgresIndexerDb` to `postgres.IndexerDb` to appease the linter. Also it should now have a logger instance, so you can use `db.log.Printf` instead of this StandardLogger trick.

brianolson:
replaced with #248 which is branched against master

PR #248, opened 2020-11-13
Acct query fix

brianolson:
fix for #246
move LIMIT inside an earlier clause to fix postgres query handling
for future bad queries, log any SQL that takes longer than 1 second to run
change against `master` branch


winder (approved):
LGTM

PR #249, opened 2020-11-16
Make sure GetMaxRound returns a value if there are no rounds processed.

winder:


idb/postgres/postgres.go
```diff
@@ -368,6 +368,13 @@ func (db *IndexerDb) GetMaxRound() (round uint64, err error) {
 	round = 0
 	row := db.db.QueryRow(`SELECT max(round) FROM block_header`)
 	err = row.Scan(&round)
+
```
brianolson, line 371: `if err == sql.ErrNoRows { return 0, nil }`
Otherwise let the err propagate if it's a connection error or something more serious.

brianolson (approved):

idb/postgres/postgres.go
winder: Unfortunately we don't get an easy error type to key off of: `failed to get max round, sql: Scan error on column index 0, name \"max\": converting NULL to uint64 is unsupported`

Take a look at the latest change, I found this way to deal with nullable fields. I could write the Uint64 version of this interface, but it seems kind of pointless considering the data we're working with here.

brianolson: okay, yup, nuts. aggregation turns no rows into a null row. O.o

brianolson (approved):

PR #256, opened 2020-12-04
Tests for the new created_at/closed_at values

winder:
## Summary

These values are unique to Indexer, so I curated a few state transitions in the go-algorand e2e script tests and am using them here.

## Test Plan

Extended the migration test script with an integration test entry point.


test/common.sh
```diff
@@ -0,0 +1,357 @@
...
+function cumulative_rewards_tests() {
+    call_and_verify 'Ensure migration updated specific account rewards.' '/v2/accounts/FZPGVIFCMHCE2HC2LEDD7IZQLKZVHRV5PENSD26Y2AOS3OWCYMKTY33UXI' 200 '"rewards":80000539878'
+    call_and_verify 'Ensure migration updated specific account rewards @ round = 810.' '/v2/accounts/FZPGVIFCMHCE2HC2LEDD7IZQLKZVHRV5PENSD26Y2AOS3OWCYMKTY33UXI?round=810' 200 '"rewards":80000539878'
+    call_and_verify 'Ensure migration updated specific account rewards @ round = 800.' '/v2/accounts/FZPGVIFCMHCE2HC2LEDD7IZQLKZVHRV5PENSD26Y2AOS3OWCYMKTY33UXI?round=800' 200 '"rewards":68000335902'
+    call_and_verify 'Ensure migration updated specific account rewards @ round = 500.' '/v2/accounts/FZPGVIFCMHCE2HC2LEDD7IZQLKZVHRV5PENSD26Y2AOS3OWCYMKTY33UXI?round=500' 200 '"rewards":28000055972'
+    call_and_verify 'Ensure migration updated specific account rewards @ round = 100.' '/v2/accounts/FZPGVIFCMHCE2HC2LEDD7IZQLKZVHRV5PENSD26Y2AOS3OWCYMKTY33UXI?round=100' 200 '"rewards":7999999996'
+}
+
+# $1 - the DB to query
+function create_delete_tests() {
```
winder, line 292: Here are the new tests, I run in two contexts to test the normal import, and the migration.
1) I took an e2edata archive snapshot with your script, and run these tests after setting up a DB with `algorand-indexer import`
2) I took the same e2edata archive and setup a DB using the current 2.2.1 release, then took a `pg_dump` of the database. I initialize a database with the dump, then run the migration.

misc/buildtestdata.sh
```diff
@@ -33,7 +43,7 @@ mkdir -p "${E2EDATA}"
 
 (cd "${E2EDATA}" && tar -j -c -f net_done.tar.bz2 --exclude node.log --exclude agreement.cdv net)
 
-if false; then
```
brianolson, line 36: Sorry, I think I left this half done, there are two tests, `misc/e2etest.py` which is the older one and `misc/e2elive.py` which is newer. The 'live' test compares `algod` API output to Indexer API output and should be better and more sustainable. The older e2etest.py uses Python to pick apart the algod account database. The extra 320 rounds below are only needed for e2etest.py.

brianolson (commented):
can we avoid putting a 300kB blob in source git? Maybe keep it in S3 with other test data?

misc/buildtestdata.sh
brianolson: See #264 
I hope to just switch over to the new test there.

PR #257, opened 2020-12-05
Fix Total Rewards Rewind

winder:
## Summary

Due to the total rewards being reset when an account is closed/reopened, extra metadata is required in the transaction to rewind.

## Remaining Work

- [ ] Create the test data using new testing harnesses
- [ ] The postgres_migration algorithm still needs updating to set the extra metadata (does this need to be split into a new migration or can we push the indexer release out another week...?).
- [ ] A similar change (import / migration) for all of the`closed_at` fields: account, asset, account_asset, app, account_app


winder:
At this time rewinding total_rewards and closed_at is not going to be supported.

PR #261, opened 2020-12-11
Clear Rewards During Migration

winder:
## Summary

During the migration there may be partial rewards reported, this change makes sure that those partial values are not returned from the REST API until the migration is finished.

## Test Plan

I'm running this code against a mainnet DB that is currently migrating, and verified that the rewards field is getting reset to 0.
Also ran Indexer in --no-algod mode and verified that it is also able to detect the incomplete migration and reset the rewards field to 0 during the migration.

I'll make sure these configurations start displaying the rewards once the migration has finished.


idb/postgres/postgres_migrations.go
```diff
@@ -122,6 +136,59 @@ func (db *IndexerDb) markMigrationsAsDone() (err error) {
...
+
+// Cached values for processAccount
+var hasRewardsSupport = false
+var lastCheckTs time.Time
+
+// processAccount is a helper to modify accounts based on migration state.
+func (db *IndexerDb) processAccount(account *generated.Account) {
+	if hasRewardsSupport {
+		// skip this whole if/else block
+	} else if db.migration != nil {
```
brianolson, line 163: To be sneakily thread-safe-ish without a mutex, grab a local copy of db.migration and check if that local copy is not nil. db.migration could go from not nil to nil between check and .GetStatus()

idb/postgres/postgres_migrations.go
```diff
@@ -892,22 +959,21 @@ func (mtxid *txidFiuxpMigrationContext) putTxidFixupBatch(batch []idb.TxnRow) er
 	row := tx.QueryRow(`SELECT v FROM metastate WHERE k = $1`, migrationMetastateKey)
 	var migrationStateJSON []byte
 	err = row.Scan(&migrationStateJSON)
-	var txstate MigrationState
-	if err == sql.ErrNoRows && state.NextMigration == 0 {
-		// no previous state, ok
-	} else if err != nil {
+	var txstate *MigrationState
+
+	txstate, err = getMigrationState(db)
```
brianolson, line 964: isn't this redundant with line 959 QueryRow above?

idb/postgres/postgres_migrations.go
winder: I made an attempt to simplify this logic. For the nil check I modified GetStatus to handle the nil case, I've never used that go feature before but I think it works nicely here.

brianolson (approved):

PR #264, opened 2020-12-14
remove support for e2etest.py and use e2elive.py

brianolson:
## Summary

Switch to better test script. Old e2etest.py uses Python to pick apart `algod` account database sqlite3 file. This is fragile if `algod` ever changes its schema (which is likely). Instead, `e2elive.py` starts a local running `algod` and runs API queries against both algod and algorand-indexer and compares the results (which should be the same if indexer accounting is correct).

## Test Plan

This is a test. I've used it locally a bunch. If it passes in cloud automated testing environments then call it good.


misc/buildtestdata.sh
```diff
@@ -33,49 +33,8 @@ mkdir -p "${E2EDATA}"
 
 (cd "${E2EDATA}" && tar -j -c -f net_done.tar.bz2 --exclude node.log --exclude agreement.cdv net)
```
btoll, line 34: I believe you can use `tar -C` to specify the directory in which to perform the `tar` instead of `cd`ing in a subshell.

winder (commented):
Looks good, I just have a couple question about the repercussions of this change.

I'm using this block and the e2edata.tar.bz2 file for creating some of my tests.

It looks like the alternative is to use net_done.tar.bz2 instead. Is this file the entire private network directory, so I can resume it and use it to initialize the indexer? It sounds like we should be able to kill the import subcommand soon.

Also, I'm using import in some of my scripts (for indexer, and for the SDK tests), so I'll need to update those before we can completely remove import.

algobolson:
Will, right, the `import` is replace by running indexer against an algod running the Archival node of the private network (with about 70 rounds stored).

brianolson:
Got this all passing automated checks.

winder (approved):
LGTM

PR #268, opened 2020-12-18
Performance Test

winder:
## Summary

This adds some basic queries against a mainnet database to help ensure we don't regress on query performance.

This test requires an external database, so it's possible that the DB will not have the necessary schema. The test will be disabled in that case.


.travis.yml
```diff
@@ -1,30 +1,30 @@
 language: go
 go:
-  - 1.13.x
```
winder, line 3: the `travis encrypt ...` command reformatted this file.

test/perf.sh
```diff
@@ -0,0 +1,58 @@
...
+source common.sh
+trap cleanup EXIT
+
+start_indexer_with_connection_string "$1" "read-only" > /dev/null
+
+wait_for_started
+
+# Disable the test instead of exiting with an error.
+set +e
+if ! IGNORED="$(wait_for_migrated '1')"; then
```
winder, line 27: At the moment this case is being hit because the mainnet DBs don't have the cumulative rewards changes yet.

brianolson:
Out of my comfort zone for bash scripting (something this complex I would have done in Python). But looks like it's probably good.

PR #269, opened 2020-12-18
Include txn type into panic message

algorandskiy:
## Summary

Current message 
```
echo: http: panic serving 127.0.0.1:64670: unknown txn type
```
does not give any clue what's going on. Adding txn type would give some more details.

## Test Plan

Simple change, no tests.


winder (approved):
LGTM

PR #224, opened 2020-10-15
Fast asset filter

brianolson:
do has-asset and has-app filtering in sql instead of client side.
total query looks like
``` sql
WITH qasf AS (SELECT addr FROM account_asset WHERE assetid = $1 AND amount > $2),
qaccounts AS (SELECT a.addr, a.microalgos, a.rewardsbase, a.keytype, a.account_data FROM account a JOIN qasf ON a.addr = qasf.addr ORDER BY a.addr ASC),
qaa AS (SELECT xa.addr, json_agg(aa.assetid) as haid, json_agg(aa.amount) as hamt, json_agg(aa.frozen) as hf FROM account_asset aa JOIN qaccounts xa ON aa.addr = xa.addr GROUP BY 1), qap AS (SELECT ya.addr, json_agg(ap.index) as paid, json_agg(ap.params) as pp FROM asset ap JOIN qaccounts ya ON ap.creator_addr = ya.addr GROUP BY 1),
qapp AS (SELECT app.creator as addr, json_agg(app.index) as papps, json_agg(app.params) as ppa FROM app JOIN qaccounts ON qaccounts.addr = app.creator GROUP BY 1),
qls AS (SELECT la.addr, json_agg(la.app) as lsapps, json_agg(la.localstate) as lsls FROM account_app la JOIN qaccounts ON qaccounts.addr = la.addr GROUP BY 1)
SELECT za.addr, za.microalgos, za.rewardsbase, za.keytype, za.account_data, qaa.haid, qaa.hamt, qaa.hf, qap.paid, qap.pp, qapp.papps, qapp.ppa, qls.lsapps, qls.lsls FROM qaccounts za LEFT JOIN qaa ON za.addr = qaa.addr LEFT JOIN qap ON za.addr = qap.addr LEFT JOIN qapp ON za.addr = qapp.addr LEFT JOIN qls ON qls.addr = za.addr ORDER BY za.addr ASC LIMIT 100;
```
#222

PR #272, opened 2021-01-06
Fix app key deletion when the only one key in deltas

algorandskiy:
## Summary

When local state delta has a single key, deletion did not happen.

## Test Plan

Added unit test, added state test (see https://github.com/algorand/go-algorand/pull/1812)


algorandskiy:
Closes #224

winder (approved):
LGTM

brianolson (approved):

PR #280, opened 2021-01-15
Remove idb from accounting.State

winder:
## Summary

Move db.CommitRoundAccounting out of the accounting package. This will make it easier to use the AddTransaction function in different ways. For example the RoundAccounting object can be extracted and filtered such that it only applies updates for a single account.

### DIFF
--- a/accounting/accounting.go
+++ b/accounting/accounting.go
@@ -2,6 +2,7 @@ package accounting
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"math/big"
 
@@ -14,12 +15,9 @@ import (
 
 // State is used to record accounting changes as a result of processing transactions.
 type State struct {
-	db idb.IndexerDb
-
 	defaultFrozen map[uint64]bool
 
 	currentRound uint64
-	dirty        bool
 
 	idb.RoundUpdates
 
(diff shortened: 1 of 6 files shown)


importer/helper.go
```diff
@@ -252,17 +252,33 @@ func updateAccounting(db idb.IndexerDb, genesisJSONPath string, numRoundsLimit i
...
 	txns := db.YieldTxns(context.Background(), state.AccountRound)
 	currentRound := uint64(0)
 	roundsSeen := 0
 	lastRoundsSeen := roundsSeen
+	var blockPtr *types.Block = nil
 	for txn := range txns {
 		maybeFail(txn.Error, l, "updateAccounting txn fetch, %v", txn.Error)
 		if txn.Round != currentRound {
+			if blockPtr != nil {
+				// TODO: commit rounds with empty paysets to avoid a special case to update the db metastate.
```
brianolson, line 265: I think with current logic metastate doesn't update when there are empty blocks, and this works for accounting but looks weird when following along trying to guess index db state from doing SQL queries.

accounting/accounting.go
```diff
@@ -36,52 +34,18 @@ type State struct {
...
-func (accounting *State) commitRound() error {
-	if !accounting.dirty {
-		return nil
-	}
-	err := accounting.db.CommitRoundAccounting(accounting.RoundUpdates, accounting.currentRound, accounting.rewardsLevel)
-	if err != nil {
-		return err
-	}
-	accounting.RoundUpdates.Clear()
-	accounting.dirty = false
```
brianolson, line 76: looks like we lost these two lines in the shuffle

brianolson (commented):
I agree. Generally cleaner interfaces this way.

importer/helper.go
winder: The special case I'm talking about in this comment is the HandleBlock function in `daemon.go`. It checks the payset size and updates the metastate, which seems really out of place. If `CommitRoundAccounting` were able to handle rounds with no transactions I think we could remove it:
```
func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
        ...
	bih.imp.ImportDecodedBlock(block)
	importer.UpdateAccounting(bih.db, genesisJSONPath, logger)
	dt := time.Now().Sub(start)
	if len(block.Block.Payset) == 0 {
		// update the meta state here!!
	}
	logger.Infof("round r=%d (%d txn) imported in %s", block.Block.Round, len(block.Block.Payset), dt.String())
	bih.round = uint64(block.Block.Round)
}
```

accounting/accounting.go
winder: Thanks, missing the round update clear is probably why the tests failed.
I think that `accounting.dirty` is no longer needed since I'm checking the payset length in importer now.

brianolson (approved):
LGTM

Issue #279, opened 2021-01-13
Indexer Tech Debt


winder:
## Summary
List of specific tech debt items that needs minor/major refactoring to cleanup. Probably split this up into separate tickets when it's time to implement.

## Main loop refactoring

- The main block handler is designed so that for every new block, it will process the current block and any unprocessed block before it. In normal operation all previous blocks have already been processed, but in the case of a sneaky migration or a node restart, this behavior is critical. This cleanup phase should be done directly during an initialization phase (like the genesis import) instead of implicitly while handling a block.
- `Importer` has logic for managing tar files in it, this stuff should get moved to the `import.go` command.
- `accounting.addrIsZero` can be removed, address.IsZero() is part of the SDK.
- Move `UpdateAccounting` and other related functions from `importer/helper.go` to `importer/importer.go`. Rename `helper.go` to `tar_importer.go`.
- The `import` subcommand is deprecated. Anything using it needs to be refactored. The main thing here is the new test scripts, they need to be regenerated and updated to use the algod node archive.

- ~Initialize maps in `RoundUpdates::Clear` function, remove nil checks in the round update helper functions.~ #330
- ~`updateAccounting` could be passed in the block instead of querying the current DB state. Leave it up to the idb layer to decide whether or not to commit the accounting changes.~ #331 
- ~`importer.updateAccounting` requires a path to the genesis file in some cases. This should be removed and put into a `importer.InitializeAccounting` function which will be a No-Op if there is nothing to initialize.~ #331 
- ~`importer.updateAccounting` uses `db.GetMetstate`. It should use `db.GetMaxRound` instead.~ #330
- ~`daemon.HandleBlock` has logic for updating metastate when the number of transactions in a round is 0. This should be moved down into `importer.UpdateAccounting` if `idb.CommitRoundAccounting` could handle empty round updates.~ We just removed it.
- ~`accounting.State.dirty` could be removed by making it so `db.CommitRoundAccounting` can handle rounds with no transactions. This might also resolve the special case in `daemon.HandleBlock` for rounds with zero transactions.~ #280 
- ~`accounting.State.AddTransaction` is responsible for calling `db.CommitRoundAccounting`, this should be moved out of `AddTransaction` into `importer.updateAccounting`.~ #280
- ~`accounting.State` should not have access to the db object. There are only 3 places where it's used. Two of them are to get the number of transactions in a round, and the last one (above) should be moved to another package.~ #280 

## IDB Interface
[...]

PR #283, opened 2021-01-21
[prototyping] Re-Index Account Subcommand

winder:
# **This issue is deprecated, but contains some refactoring which should be pulled into a new PR.**

## Summary

Refactor `UpdateAccounting`:
- Move the init code to a separate function
- New "Filter" parameters for StartRound, RoundLimit, and a new Address filter. 
     - Use by `YieldTxns` wrapper, there is a different algorithm when there is an account filter.
     - `RoundUpdates` has a new `Filter` function, it is a no-op unless there is an address filter.

IndexerDb changes:
- Added `GetImportState` / `SetImportState` functions because ImportState is used outside of the postgres package.
- Added `DeleteAccount` function because the re-index algorithm needs a way to ensure the account is deleted before starting a re-index.
- Added an option to `Transactions` to configure whether the sort order is ascending or descending. It is not exposed in the API, I just needed it for fetching account transactions oldest to newest.

Reindex account subcommand:
- With the above changes this is little more than a wrapper to the `UpdateAccounting` function with an address filter.

## Test Plan

So far I'm using `Sandbox` to test, I'll use some of our e2e data archives next.

### DIFF
--- a/cmd/algorand-indexer/reindexer.go
+++ b/cmd/algorand-indexer/reindexer.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/algorand/go-algorand-sdk/types"
+	"github.com/spf13/cobra"
+
+	"github.com/algorand/indexer/config"
+	"github.com/algorand/indexer/idb"
+	"github.com/algorand/indexer/importer"
+)
+
+var (
+	reindexAccount string
+)
+
+var reindexCmd = &cobra.Command{
+	Use:   "reindex",
+	Short: "reindex a single account",
+	Run: func(cmd *cobra.Command, args []string) {
+		var err error
+
+		config.BindFlags(cmd)
+		err = configureLogger()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to configure logger: %v", err)
+			os.Exit(1)
+		}
+
+		opts := idb.IndexerDbOptions{}
+		db := globalIndexerDb(&opts)
+
+		addr, err := types.DecodeAddress(reindexAccount)
+		maybeFail(err, "unable to decode address: %s", reindexAccount)
+
+		accounts := db.GetAccounts(context.Background(), idb.AccountQueryOptions{
+			EqualToAddress: addr[:],
+		})
+
+		num := 0
+		var createdAt int64
+		for acct := range accounts {
+			num++
+			createdAt = int64(*acct.Account.CreatedAtRound)
+		}
+
+		if num != 0 {
+			logger.Errorf("expected one account but found %d for address: %s", num, addr.String())
+		}
+
+		filter := idb.UpdateFilter{Address: &addr, StartRound: createdAt}
+		importer.UpdateAccounting(db, filter, logger)
+	},
+}
+
+func init() {
+	reindexCmd.Flags().StringVarP(&reindexAccount, "account", "a", "", "account which should be re-indexed")
+	reindexCmd.MarkFlagRequired("account")
+	reindexCmd.Hidden = true
+}
--- a/cmd/algorand-indexer/reindexer.go
+++ b/cmd/algorand-indexer/reindexer.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"os"
 
@@ -36,22 +35,17 @@ var reindexCmd = &cobra.Command{
 		addr, err := types.DecodeAddress(reindexAccount)
 		maybeFail(err, "unable to decode address: %s", reindexAccount)
 
-		accounts := db.GetAccounts(context.Background(), idb.AccountQueryOptions{
-			EqualToAddress: addr[:],
-		})
+		maxRounds, err := db.GetMaxRound()
+		maybeFail(err, "unable to fetch max rounds")
 
-		num := 0
-		var createdAt int64
-		for acct := range accounts {
-			num++
-			createdAt = int64(*acct.Account.CreatedAtRound)
+		limit := int(maxRounds)
+		filter := idb.UpdateFilter{
+			Address: &addr,
+			StartRound: 0,
+			RoundLimit: &limit,
 		}
-
-		if num != 0 {
-			logger.Errorf("expected one account but found %d for address: %s", num, addr.String())
-		}
-
-		filter := idb.UpdateFilter{Address: &addr, StartRound: createdAt}
+		err = db.DeleteAccount(addr)
+		maybeFail(err, "unable to delete the account: %s", addr.String())
 		importer.UpdateAccounting(db, filter, logger)
 	},
 }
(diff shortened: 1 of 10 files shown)


idb/dummy.go
```diff
@@ -516,19 +543,103 @@ type RoundUpdates struct {
...
+	//ru.AccountDataUpdates = nil
+	//ru.AssetUpdates = nil
+
+	ru.AlgoUpdates = make(map[[32]byte]*AlgoUpdate)
+	ru.AccountTypes = make(map[[32]byte]string)
+	ru.AccountDataUpdates = make(map[[32]byte]map[string]interface{})
 	ru.AcfgUpdates = nil
 	ru.TxnAssetUpdates = nil
-	ru.AssetUpdates = nil
+	ru.AssetUpdates = make(map[[32]byte][]AssetUpdate)
```
winder, line 556: I'm guessing that these were left uninitialized for performance reasons, so this is a performance vs maintainability debate.

Initializing them here would let us avoid the boilerplate nil checking in several places.

idb/dummy.go
```diff
@@ -516,19 +543,103 @@ type RoundUpdates struct {
...
 }
 
+type UpdateFilter struct {
+	StartRound int64
+	RoundLimit *int
+	Address    *types.Address
+}
+
+// Filter is used to get a RoundUpdates copy containing only particular RoundUpdates.
+func (ru *RoundUpdates) Filter(filter UpdateFilter) RoundUpdates {
```
winder, line 571: I think this is the most likely spot where I have bugs. I need to ensure that only `account`, `account_app`, and `account_asset` are modified.

While writing this comment I realized that the `CommitRoundAccounting` probably needs to be changed, it needs to avoid setting the accounting round when re-indexing.

idb/postgres/postgres.go
```diff
@@ -2640,6 +2676,12 @@ func (db *IndexerDb) GetSpecialAccounts() (accounts idb.SpecialAccounts, err err
 	return
 }
 
+// DeleteAccount is used to remove an account from the database, needed for re-indexing an account. part of idb.IndexerDB.
+func (db *IndexerDb) DeleteAccount(address atypes.Address) error {
+	_, err := db.db.Exec(`DELETE FROM account WHERE account.addr = $1`, address[:])
```
winder, line 2681: I think I need to delete `account_app` and `account_asset` as well.

importer/helper.go
```diff
@@ -92,7 +81,23 @@ func (h *ImportHelper) Import(db idb.IndexerDb, args []string) {
...
+	}
+	if !initialImport {
+		state, err := db.GetImportState()
+		maybeFail(err, h.Log, "problem getting the import state")
+		filter.StartRound = state.AccountRound
+	}
+	accountingRounds, txnCount := updateAccounting(db, filter, h.Log)
+	if initialImport {
+		accountingRounds++
+	}
```
winder, line 100: Since we're deprecating `import` I'm not too concerned about how ugly this became.

idb/postgres/postgres.go
winder: Thinking about this more, I think the RoundUpdate filter may need to exclude things that modify `asset` and `app`.

`asset` because there may be a different manager address, and including those transactions would be difficult.
`app` I'm not so sure about.

cmd/algorand-indexer/reindexer.go
```diff
@@ -0,0 +1,57 @@
...
+		maxRounds, err := db.GetMaxRound()
+		maybeFail(err, "unable to fetch max rounds")
+
+		limit := int(maxRounds)
+		filter := idb.UpdateFilter{
+			Address: &addr,
+			StartRound: 0,
+			RoundLimit: &limit,
+		}
+		err = db.DeleteAccount(addr)
```
brianolson, line 47: This looks like it should only be run on a database with no daemon running.
Hm, I'd previously balked at trying to put app locking into the database to prevent multiple writers but now I think about it and maybe there's a reasonable way to do it to at least provide a warning.
Add a field to the ImportState which is a timestamp of when it is written.
Add another row to metastate that is an app lock claim, probably it would have start-time and process-id. On graceful exit (^C or kill other than -9) that would be cleared.
On startup check that row, if not present, set one. If present, check ImportState last written time, if that's older than about a minute assume the prior task died, otherwise log error ("looks like there's another writer, exiting") and exit.
This whole interlock system is probably out of scope for this change and could be a separate issue/PR

idb/dummy.go
brianolson: initialized non-empty for map is probably nicer. append(nil, ...) works but maps panic if nil.

idb/dummy.go
```diff
@@ -516,19 +546,109 @@ type RoundUpdates struct {
...
+	var response RoundUpdates
+	response.Clear()
+
+	var addrBytes [32]byte
+	copy(addrBytes[:], filter.Address[:])
+
+	if val, ok := ru.AlgoUpdates[addrBytes]; ok {
+		response.AlgoUpdates[addrBytes] = val
+	}
+
```
brianolson, line 594: missing AccountTypes

idb/postgres/postgres.go
```diff
@@ -1397,13 +1422,24 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
...
 	}
+
 	if joinParticipation {
+		order := "DESC"
+		if tf.ResultOrder != nil {
+			order = string(*tf.ResultOrder)
+		}
 		// this should match the index on txn_particpation
-		query += " ORDER BY p.addr, p.round DESC, p.intra DESC"
+		query += " ORDER BY p.addr, p.round " + order + ", p.intra " + order
```
brianolson, line 1432: It might suck if it's not DESC, the index is DESC ordered

idb/postgres/postgres.go
```diff
@@ -1397,13 +1422,24 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
...
 	} else {
+		order := ""
+		if tf.ResultOrder != nil {
+			order = " " + string(*tf.ResultOrder)
+		}
 		// this should explicitly match the primary key on txn (round,intra)
-		query += " ORDER BY t.round, t.intra"
+		query += " ORDER BY t.round, t.intra" + order
```
brianolson, line 1439: probably want order applied to .round and .intra like above.

idb/postgres/postgres.go
brianolson: Now I slightly wish I'd set up FOREIGN_KEY constraints in the schema and we could do delete with CASCADE.
Or maybe this is always a hackish tool that should never actually be used in prod? If that's the case maybe split it out to a separate cmd/reindex/

brianolson (commented):
A bunch of misc comments. It looks like there's some big TODOs left too.

cmd/algorand-indexer/reindexer.go
winder: Yes, this command is really dangerous. I made sure to set the "Hidden" flag so it doesn't show up in the help and should probably remove it altogether (or move it to a different root command) before merging any of this work in.

Some sort of locking scheme would be great.

winder:
The most useful refactoring from this issue was committed in #330 / #331. More details in #279

PR #284, opened 2021-01-21
Json for postgres

brianolson:
## Summary

Fix handling of some transactions whose JSON encoding would fail to be received by PostgreSQL. Emit stricter JSON, keys are strings, strings are utf8 or base64

## Test Plan

Tested against test data from go-algorand e2e test scripts (including new data added there to exercise this condition).


tsachiherman (commented):
Two things that I think that are still missing here:
    - The value in the jt.EvalDelta.GlobalDelta/jt.EvalDelta.LocalDeltas needs also to be encoded in base64.
    - When making a query, you would need to make a b64 decode to restore the original byte array.

brianolson:
Indexer mostly doesn't use the json encoding of the txn. A few specific fields are used in filters but they aren't affected by this change. The value side of fields was already base64, just the key strings needed fixup.

tsachiherman:
> Indexer mostly doesn't use the json encoding of the txn. A few specific fields are used in filters but they aren't affected by this change. The value side of fields was already base64, just the key strings needed fixup.

I meant that when a new block is being received and applied, you need to decode the old keys since these might need to get overridden by the new ones. ( i.e. the old one is the b64('key') whereas the new one just just the `key` )

idb/postgres/postgres_migrations.go
```diff
@@ -34,6 +34,7 @@ func init() {
 		{m3acfgFix, false, "Recompute asset configurations with corrected merge function."},
 		{m4RewardsAndDatesPart1, true, "Update DB Schema for cumulative account reward support and creation dates."},
 		{m5RewardsAndDatesPart2, false, "Compute cumulative account rewards for all accounts."},
+		{m6MarkTxnJSONSplit, true, "record round at which txn json recording changes, for future migration to fixup prior records"},
```
winder, line 37: This will cause `m5RewardsAndDatesPart2` to block, so it will need to be moved to m4.

brianolson: m5 has been released as 2.3.0, so I don't think we can move it.
m5 only becomes blocking if both are applied at the same time. If an Indexer has been upgraded to 2.3.0, then this (call it 2.3.1 ?) should be fine.

winder (approved):

PR #288, opened 2021-01-25
Will/prevent concurrent update

winder:
## Summary

This seems to fix the multiple writer issue.

## Test Plan

I have a version of sandbox which launches multiple sandbox instances. Without this fix it reliably double-counts transactions. With this fix things are correct.


winder:
This PR is obsolete. Closing.

PR #289, opened 2021-01-26
Round writer check

brianolson:
## Summary

Simplify daemon import code. Always do round-update inside CommitRoundAccounting even if no txns.
Fix --keep-temps in e2elive.py
multi-writer resistance: CommitRoundAccounting will rollback if metastate has already recorded a round >= the round being committed.
Fix sql usage to actually be ACID

## Test Plan

Passes e2e tests.


idb/postgres/postgres.go
```diff
@@ -1222,6 +1232,11 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
 			return
 		}
 	}
+	if istate.AccountRound >= int64(round) {
```
winder, line 1235: I think there's a race condition here. If both transactions run the select before the upsert then they could both still be committed.

What do you think about
1) Removing the metastate query and setting the account round directly with `WHERE round < $1`.
2) Returning an error if the updated rows equal zero (meaning another transaction beat us to the punch)

idb/postgres/postgres.go
```diff
@@ -39,6 +40,9 @@ const stateMetastateKey = "state"
 const migrationMetastateKey = "migration"
 const specialAccountsMetastateKey = "accounts"
 
+// Be a real ACID database
+var serializable = sql.TxOptions{Isolation: sql.LevelSerializable}
```
winder, line 44: I'm confused at how this helps. I thought we need to enable dirty reads with `sql.LevelReadUncommitted` in order to see if multiple transactions are working on the same data concurrently.

winder: Looks like it will fail when each transaction tries to write the same thing with different values. Nice.

idb/dummy.go
```diff
@@ -218,7 +218,7 @@ type IndexerDb interface {
 	// YieldTxns returns a channel that produces the whole transaction stream after some round forward
 	YieldTxns(ctx context.Context, prevRound int64) <-chan TxnRow
 
-	CommitRoundAccounting(updates RoundUpdates, round, rewardsBase uint64) (err error)
+	CommitRoundAccounting(updates RoundUpdates, round uint64, blockPtr *types.Block) (err error)
```
winder, line 221: What do you think about adding `round` to the `RoundUpdates` object instead of adding a new parameter?

winder: We could put the block data in there as well actually.

brianolson: Yes, and I got started making this change, then stopped. RoundUpdates is used heavily in accounting/ code; but those two fields are just metadata to CommitRoundAccounting().
Alternate theory: RoundUpdates struct definition could move to the accounting package.
Not sure which way I want to go so I guess leave it for now?

winder (approved):
LGTM

PR #292, opened 2021-01-29
Fix asset close accounting issue.

winder:
## Summary

When an asset is closed we need to know the exact asset balance at that time. With the existing logic, we apply all asset payments before closing the asset. Because of this, if the asset receives payment after it was closed, Indexer incorrectly sends that payment to the close-to address.

## Test Plan

1) New unit tests verify that subround accounting is done as designed for asset closes.
2) New unit/integration test to execute the subround accounting object on a DB, with this test I was able to test the before and after of this change and see that something was fixed.
3) Re-indexing betanet, I'll verify that the 5 bad transactions have been corrected.


idb/postgres/postgres.go
```diff
@@ -871,52 +871,97 @@ func (db *IndexerDb) CommitRoundAccounting(updates idb.RoundUpdates, round, rewa
...
+						db.log.Errorf("%d axfer %s %s", round, b64(addr[:]), obs(au))
 					}
-					for !au.Delta.IsInt64() {
-						_, err = seta.Exec(addr[:], au.AssetID, step, au.DefaultFrozen, round)
+
+					// Apply deltas
+					if au.Delta.IsInt64() {
+						// easy case
+						delta := au.Delta.Int64()
+						// don't skip delta == 0; mark opt-in
```
brianolson, line 918: I think in this code we _can_ now skip if (delta == 0 && au.Close != nil)

idb/postgres/postgres.go
```diff
@@ -871,52 +871,97 @@ func (db *IndexerDb) CommitRoundAccounting(updates idb.RoundUpdates, round, rewa
...
+		defer acs.Close()
+		// On asset opt-out mark the account_asset as closed with zero balance.
+		acd, err := tx.Prepare(`UPDATE account_asset SET amount = 0, closed_at = $1 WHERE addr = $2 AND assetid = $3`)
+		if err != nil {
+			return fmt.Errorf("prepare asset close2, %v", err)
+		}
+		defer acd.Close()
+
+
+		for _, subround := range updates.AssetUpdates {
```
brianolson, line 907: Idea for refactoring this. Keep the fused update-close apply section of code here, but move the 'subround' behavior into the existing aulist []AssetUpdate. In updateAsset() search the aulist from the end. If the item found has a close (or as is now, if no AU for the assetid), create a new AssetUpdate and append it. There will now be two AssetUpdate records in the list to be applied. So a close & reopen in one round should do the right thing.
This rearranges the complexity; not sure if it's better than the 'subround' way. It might create fewer database Exec statements by allowing all the assets not closed to continue aggregating txns into one update delta for the round.

winder: I like this idea and spent a little time trying to get it to work before the current implementation. I wanted to keep as much of the complexity in the accounting package as possible and avoid changing the type of `updates.AssetUpdates`.

The problem occurs when there are multiple assets closing into each other. When any asset closes it might close into another asset which will close later. It is probably possible to do some sort of dependency resolution and recursively commit the things which need to be committed in the right order.

My hunch is that in the vast majority of cases there will only be a single "subround", so this sort of optimization probably isn't worthwhile.

brianolson: Two senses of the ultimate problem here:
* Indexer tried to prematurely optimize expecting that PostgreSQL would be slow, so aggregate transactions in Go code and apply fewer comprehensive transactions that achieve the net delta of the round. This emulation of the is tricky and not quite right.
* Indexer is also not following the algod model of storing all the AccountData in one blob, fetching it, and writing back one new AccountData blob at the end of the round. Feeding actions to PostgreSQL to make the read-modify-write happen server side should be faster, but is just different than what the accounting logic really wants to do. Algorand transactions only sometimes have all the delta they need to apply that way. Payment txns work better that way.

There's been some talk of "maybe we should make the algod accounting logic a library that both algod and Indexer can include". It would be _two_ big projects to make that change on the algod side and then make that change on the Indexer.

idb/postgres/postgres_integration_test.go
```diff
@@ -0,0 +1,136 @@
+package postgres
+
+import (
+	"database/sql"
+	"fmt"
+	"testing"
+
+	_ "github.com/lib/pq"
+	"github.com/orlangure/gnomock"
```
brianolson, line 9: I still don't like it but I'll let it go since it works on travis. Next time I try to run tests I'll have to find a way to hack it out and make it optional and disabled if `docker` isn't available.

brianolson (approved):

idb/postgres/postgres_integration_test.go
winder: I'm sure we can figure something out, the connection string environment var should work.

Issue #287, opened 2021-01-25
migration: json and close fixup


brianolson:
## Summary
Write a migration to fix up old data for

- the account-close bug fixed in #286
- the json encoding bug fixed in #284 
- per 3ba2065646a1d1cd0557cba777bd567993e026da , if []TealKeyValue is length 0; remove the field from storage.
- like #290 check if a DefaultFrozen was missed and if so apply it to an asset holding.

The migration should probably do one read of all the txns, taking no action on most, updating `account.account_data = NULL` on account closes and doing a JSON encode and writing it to db if the new encoding is different than the previous encoding.

The `metastate` table marker row added in #284 should be removed when this migration finishes.

## Scope/Requirements
Write a new non-blocking migration to update data.

## Urgency/Relative Priority
Urgentish/soonish. Some deployed Indexers have data that is confusing apps written against them.

brianolson:
Also adjust asset records for possible mis-set frozen from default-frozen.
Also maybe adjust some state delta records.

brianolson:
per 3ba2065646a1d1cd0557cba777bd567993e026da
if []TealKeyValue is length 0; remove the field from storage.

PR #293, opened 2021-02-01
Migration omnibus

brianolson:
## Summary

Implement #287 

## Test Plan

I'll run this against a mainnet db.


idb/postgres/postgres_migrations.go
```diff
@@ -35,6 +36,8 @@ func init() {
 		{m4RewardsAndDatesPart1, true, "Update DB Schema for cumulative account reward support and creation dates."},
 		{m5RewardsAndDatesPart2, false, "Compute cumulative account rewards for all accounts."},
 		{m6MarkTxnJSONSplit, true, "record round at which txn json recording changes, for future migration to fixup prior records"},
+		{m7StaleClosedAccounts, false, "clear some stale data from closed accounts"},
+		{m8TxnJsonEncoding, false, "some txn JSON encodings need app keys base64 encoded"},
```
winder, line 40: m5 has not been released, it's a long blocking migration so it needs to happen at the end

winder (changes requested):
Looks good, but please add a migration test. See `test/postgres_migration_test.sh`

algorotem:
Test plan wise - running it against MainNet isn't sufficient. Consider having a smaller scale test that has the different potential scenarios. MainNet doesn't cover all of them.

idb/postgres/postgres_migrations.go
brianolson: m7 is trivially small. 3 lines of sql.
m8 is structurally different, it needs to scan every txn and maybe update a field. (it's based on the copy of the prior txid update)

brianolson: m6 could be marked non-blocking. It doesn't actually need to be precise. The json update is harmless if it reads a few more txns and find the json already in good form because it was written after the bugfix started running.

algorotem: How long will it take to run all of these? it seems like m8 is as long as m5.

idb/postgres/postgres_integration_test.go
```diff
@@ -28,6 +38,9 @@ func getAccounting(round uint64, cache map[uint64]bool) *accounting.State {
 
 // setupPostgres starts a gnomock postgres DB then returns the connection string and a shutdown function.
 func setupPostgres(t *testing.T) (*sql.DB, string, func()) {
+	if !enableDocker {
```
winder, line 41: These can't be disabled by default, it needs to be an action people take so that they aren't forgotten about. You could use the "short" option and have them disabled `go test -short`, or use something like "go test `go list ./... | grep -v integration_test`"

winder (changes requested):
Changes look good, we just need a solution for the tests.

winder (approved):

PR #294, opened 2021-02-02
Add 'deleted' flag to the database, update migration and integration tests

winder:
## Summary

With `created_at` meaning "first_created_at" and `closed_at` meaning "most_recently_closed_at", it is impossible to know whether or not a particular account/creatable/holding is deleted or not. Specifically when there are multiple opt-in transactions for empty objects.

This change adds a boolean flag next to all of the created_at/deleted_at fields to indicate whether or not something is deleted.

## Test Plan

There are comprehensive integration and migration tests for the created_at / deleted_at, these were extended to also check for the deleted flag.

## TODO

API updates in a later PR


idb/postgres/postgres_migrations.go
```diff
@@ -371,14 +371,19 @@ func m4RewardsAndDatesPart1(db *IndexerDb, state *MigrationState) error {
 		`ALTER TABLE account ADD COLUMN rewards_total bigint NOT NULL DEFAULT 0`,
 
 		// created/closed round
+		`ALTER TABLE account ADD COLUMN deleted boolean DEFAULT NULL`,
```
brianolson, line 374: This needs to be a new migration because there are systems that have already applied m4

idb/postgres/setup_postgres_sql.go
```diff
@@ -52,6 +52,7 @@ CREATE TABLE IF NOT EXISTS account (
   microalgos bigint NOT NULL, -- okay because less than 2^54 Algos
   rewardsbase bigint NOT NULL,
   rewards_total bigint NOT NULL,
+  deleted boolean NOT NULL DEFAULT false,
```
brianolson, line 55: for got to add setup_postgres.sql to the PR?

brianolson (changes requested):

idb/postgres/setup_postgres_sql.go
winder: looks like it, fixed.

idb/postgres/postgres_migrations.go
winder: I think those systems are out of luck

brianolson: That can't be the answer.
m4 went in 2020-12-18
Release 2.3.0 happened Jan 5. We have to support that release.
Though, https://releases.algorand.com/ is still only showing 2.2.1, so, 2.3.0 never really happened and is just fake garbage in github? Can we delete the tag?

brianolson: Also this could be `DEFAULT FALSE` and simplify some pointers

winder: If it is `DEFAULT FALSE` is there a way to know whether the migration has initialized them or not?

Regarding a new migration or not, I'm not sure why beta code is pushed to `master` but at the moment it is. The release notes will have to say something like "this release is not compatible with 2.3.0-beta".

brianolson: It's not just `2.3.0-beta`, I see a `2.3.0` tag which apparently was premature.

winder: ah, yeah that would be a problem.

@algojohnlee could you confirm whether or not 2.3.0 has been released?

onetechnical: 2.3.0 tag was created in anticipation of merge to master, but we held off on actually officially publishing. It would be a problem for anyone who built off master, but not for anyone using our repositories. I would feel okay with not supporting it as an official release, despite the icky merge to master.

idb/postgres/postgres_migrations.go
```diff
@@ -616,6 +651,12 @@ func processAccountTransactions(txnrows <-chan idb.TxnRow, addressStr string, ad
 			}
 		}
 
+		if txn.AssetID == 82 && stxn.Txn.Sender.String() == "CM333ZN3KMASBRIP7N4QIN7AANVK7EJGNUQCNONGVVKURZIU2GG7XJIZ4Q" && stxn.Txn.Type == sdk_types.ApplicationCallTx {
```
brianolson, line 654: stale debugging code?

brianolson (approved):
One bit of dead code that should be deleted.
Since we're under time pressure I'm not going to ask for changes, but I figured out what I want instead of deleted = NULL/TRUE/FALSE or nil/&true/&false ; I want an enum {Unknown=0, Deleted=1, Active=2}. NULL/TRUE/FALSE is probably more idiomatic SQL but the enum would be okay there. I think the enum is a better fit for Go.

PR #302, opened 2021-02-04
DefaultFrozen Asset Accounting Fix

winder:
## Summary

Fix the asset default frozen cache so that asset holdings are created correctly.
Fix accounting issues related to re-opened asset holdings.

## Test Plan

New unit and integration tests.


idb/postgres/postgres_migrations.go
```diff
@@ -651,6 +651,12 @@ func processAccountTransactions(txnrows <-chan idb.TxnRow, addressStr string, ad
 			}
 		}
 
+		if txn.AssetID == 82 && stxn.Txn.Sender.String() == "CM333ZN3KMASBRIP7N4QIN7AANVK7EJGNUQCNONGVVKURZIU2GG7XJIZ4Q" && stxn.Txn.Type == sdk_types.ApplicationCallTx {
```
brianolson, line 654: debug code

importer/helper.go
```diff
@@ -252,7 +252,9 @@ func updateAccounting(db idb.IndexerDb, genesisJSONPath string, numRoundsLimit i
 	}
 
 	lastlog := time.Now()
-	act := accounting.New()
+	cache, err := db.GetDefaultFrozen()
```
brianolson, line 255: if we keep a persistent default frozen map, accounting.go will update it on acfg

idb/postgres/postgres.go
```diff
@@ -284,9 +289,9 @@ func (db *IndexerDb) GetAsset(assetid uint64) (asset types.AssetParams, err erro
 	return
 }
 
-// GetDefaultFrozen get {assetid:default frozen, ...} for all assets
+// GetDefaultFrozen get {assetid:default frozen, ...} for all assets, needed by accounting.
 func (db *IndexerDb) GetDefaultFrozen() (defaultFrozen map[uint64]bool, err error) {
-	rows, err := db.db.Query(`SELECT index, params -> 'df' FROM asset a`)
+	rows, err := db.db.Query(`SELECT index, coalesce((params ->> 'df')::boolean, false) FROM asset a`)
```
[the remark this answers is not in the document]
winder: Cool trick, in that case I'd need to add `WHERE df is not NULL` to the query. I had some code to verify that the cache contained the key, but maybe that error checking is overkill.

importer/helper.go
winder: Oops. Yeah I didn't mean to initialize this for every round.

idb/postgres/postgres.go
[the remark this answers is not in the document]
winder: Looks like I got rid of that check anyway, so this should work nicely.

brianolson (approved):

PR #303, opened 2021-02-04
Move m6 to the m4 spot.

winder:
## Summary

Avoid a blocking migration after `RewardsAndDatesPart2`.

### DIFF
--- a/idb/postgres/postgres_migrations.go
+++ b/idb/postgres/postgres_migrations.go
@@ -23,8 +23,8 @@ import (
 	"github.com/algorand/indexer/types"
 )
 
-// rewardsMigrationIndex is the index of m5RewardsAndDatesPart2.
-const rewardsMigrationIndex = 5
+// rewardsMigrationIndex is the index of m6RewardsAndDatesPart2.
+const rewardsMigrationIndex = 6
 
 func init() {
 	migrations = []migrationStruct{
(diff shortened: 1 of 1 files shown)


idb/postgres/postgres_migrations.go
```diff
@@ -23,22 +23,22 @@ import (
...
-const rewardsMigrationIndex = 5
+// rewardsMigrationIndex is the index of m6RewardsAndDatesPart2.
+const rewardsMigrationIndex = 6
 
 func init() {
 	migrations = []migrationStruct{
 		{m0fixupTxid, false, "Recompute the txid with corrected algorithm."},
 		{m1fixupBlockTime, true, "Adjust block time to UTC timezone."},
 		{m2apps, true, "Update DB Schema for Algorand application support."},
 		{m3acfgFix, false, "Recompute asset configurations with corrected merge function."},
```
brianolson, line 34: add a comment after m3acfgFix
`// Above are the migrations as of 2.2.1`
And after 2.3.1 we should add similar after m6
I could go back and look up the history of when even earlier things were released but it doesn't matter right now and this bit of accounting is repeatedly confusing to me about what we can and mustn't do.

winder: good idea, I'll add some comments

brianolson (approved):

Issue #297, opened 2021-02-03
Endpoint `/v2/accounts/{account-id}` with round before app call fails


fabrice102:
## Summary

Calling `/v2/accounts/{account-id}` with a round query parameter corresponding a round after which an application call is made fails by the indexer not returning any response.

Concretely, on TestNet, the last application call of `FRYCPGH25DHCYQGXEB54NJ6LHQG6I2TWMUV2P3UWUU7RWP7BQ2BMBBDPD4` was 12088087 at the time where this issue was wrote.

We get:

```
$ curl -i "http://localhost:8980/v2/accounts/FRYCPGH25DHCYQGXEB54NJ6LHQG6I2TWMUV2P3UWUU7RWP7BQ2BMBBDPD4?round=12088086"
curl: (52) Empty reply from server

$ curl -i "http://localhost:8980/v2/accounts/FRYCPGH25DHCYQGXEB54NJ6LHQG6I2TWMUV2P3UWUU7RWP7BQ2BMBBDPD4?round=12088090"
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Content-Type: application/json; charset=UTF-8
Vary: Origin
Date: Wed, 03 Feb 2021 03:37:03 GMT
Transfer-Encoding: chunked

{"account":{"address":"FRYCPGH25DHCYQGXEB54NJ6LHQG6I2TWMUV2P3UWUU7RWP7BQ2BMBBDPD4","amount":391141183,"amount-without-pending-rewards":391122415,"apps-local-state":[{"id":13273130,"schema":{"num-byte-slice":0,"num-uint":0}},{"id":13310783,"key-value":[{"key":"T3B0ZWRBcHBJZA==","value":{"bytes":"","type":2,"uint":13310783}},{"key":"QWR2ZXJ0aXNlclVybA==","value":...
```

## Scope/Requirements

As a temporary workaround, it would be great to have a clearer error message and some warning in the documentation.

As a longer-term solution, application call transactions do not seem to be supported in the rewinding logic: https://github.com/algorand/indexer/blob/ffab5bcd7d169949276381931861af35f8ccfa48/accounting/rewind.go#L110

It would actually be great to be able to rewind local application state and also global application state. The later would require adding a new query parameter `round` to `/v2/application`.

brianolson:
Known issue in #62 
[...]

Issue #62, opened 2020-04-21
backend: rewind Application state


algobolson:
Given data from #61  implement 'rewind' state-at-round functionality.
Local state rewind for an account works with just the txns on that account.
Global state rewind for an app fetches all txns on the appid.
Fuzed global+local rewind does global state rewind but _also_ filters all those txns into doing rewind on an account's local state.

tsachizehub:
isn't this one already done ?

winder:
> isn't this one already done ?

No

yaovi-a:
Would this be a feature add to the indexer?

winder:
@yaovi-a yes, the idea is that the indexer would return the application state at a particular round. It could enable useful features like replaying app calls after the fact by fetching the before state then re-executing the app call in a simulated environment.

PR #304, opened 2021-02-05
better rewind error message

brianolson:
## Summary

Tweak for #297 
Report a better error message.

## Test Plan

I have demonstrated this works locally. Too small a feature to build an automated test for. Future work should eliminate this shortcoming. (See #62 )

### DIFF
--- a/accounting/rewind.go
+++ b/accounting/rewind.go
@@ -145,7 +145,8 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 			}
 		case atypes.AssetFreezeTx:
 		default:
-			panic(fmt.Sprintf("unknown txn type %s", stxn.Txn.Type))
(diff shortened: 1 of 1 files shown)


accounting/rewind.go
```diff
@@ -145,7 +145,8 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 			}
 		case atypes.AssetFreezeTx:
 		default:
-			panic(fmt.Sprintf("unknown txn type %s", stxn.Txn.Type))
+			err = fmt.Errorf("%s[%d,%d]: cannot rewind past txn type %s", account.Address, txnrow.Round, txnrow.Intra, stxn.Txn.Type)
```
winder, line 148: It's not totally clear that the reason why it cannot rewind is that it isn't supported. Maybe something like `txn type %s is not currently supported by rewind`?

brianolson: Updated with variation on this language

PR #308, opened 2021-02-08
always send json to PostgreSQL as a Go string not []byte

brianolson:
## Summary

`lib/pq` can choke on sending json as []byte where it works as string.

## Test Plan

This worked locally on e2elive.py with extended data to exercise a seen bug.


winder (approved):
LGTM

PR #314, opened 2021-02-09
Another validate_accounting script.

winder:
## Summary

Attempt at a algod/indexer comparision script with fewer moving parts.

I got carried away here. The shell script worked pretty well but was slow so I ported it to go.

Both  implementations take the approach of accepting accounts from standard in, so a secondary program is needed to feed accounts to the script.

### DIFF
--- a/cmd/validator/dynamic_processor.go
+++ b/cmd/validator/dynamic_processor.go
@@ -9,8 +9,8 @@ import (
 	"sort"
 )
 
-// GenericProcessor implements process address by dynamically normalizing the JSON and ensuring the results are equal.
-type GenericProcessor struct {
+// DynamicProcessor implements process address by dynamically normalizing the JSON and ensuring the results are equal.
+type DynamicProcessor struct {
 }
 
 func mustEncode(data interface{}) string {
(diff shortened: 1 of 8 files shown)


cmd/validator/dynamic_processor.go
```diff
@@ -0,0 +1,275 @@
+package main
```
winder, line 1: This file has a comparison function that follows the same algorithm as the jq normalization in the bash script. The output it provides on a mismatch isn't as good as the other algorithm, so this is mainly for reference....

cmd/validator/struct_processor.go
```diff
@@ -0,0 +1,466 @@
+package main
```
winder, line 1: This is the main comparison function. It serialized the responses into a struct and compares them field by field. The report on a difference is pretty nice and will tell you exactly which fields are different.

cmd/validator/validate_accounting.go
```diff
@@ -0,0 +1,230 @@
+package main
```
winder, line 1: Main program, args, concurrency, reporting, etc

misc/generate_accounts.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
```
winder, line 1: A better script is needed to stream accounts from postgres. Rotem gave me a python script that exports accounts directly from algod, so I exported them to a file. My current tests are using a file to feed the validator tool.

misc/validate_accounting.sh
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env bash
```
winder, line 1: For reference only. This script works fine but it's too slow to run against all mainnet accounts.

brianolson (commented):
It's not a complete system, but if you find it useful then okay. A couple tweak suggestions.

Issue #317, opened 2021-02-10
Compute "account.rewards" field the same way that Algod computes it


winder:
The initial implementation stores "rewards" as the distributed rewards, the same way algod does. But when algod reports the rewards it adds in the real-time pending rewards. There are arguments for both approaches, but algod has already picked a side so we may as well use the same approach with Indexer.

PR #318, opened 2021-02-10
Add pending rewards to rewards in account APIs.

winder:
## Summary

Fix for #317 

## Test Plan

The new validated accounting test scripts verify that the rewards match.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -658,6 +658,9 @@ func (si *ServerImplementation) fetchAccounts(ctx context.Context, options idb.A
 			account = row.Account
 		}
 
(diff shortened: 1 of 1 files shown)


api/handlers.go
```diff
@@ -658,6 +658,9 @@ func (si *ServerImplementation) fetchAccounts(ctx context.Context, options idb.A
 			account = row.Account
 		}
 
+		// Adjust rewards field to match algod
```
algorotem, line 661: I'd add a word about the "adjustment" to market it a bit more clear, but looks good.

algorotem (approved):

PR #321, opened 2021-02-11
GetMaxRound use metastate account_round instead of max block header.

winder:
## Summary

Modify GetMaxRound to return the accounting_round instead of the max block round.

## Test Plan

New unit tests.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -398,13 +398,13 @@ func (db *IndexerDb) SetMetastate(key, jsonStrValue string) (err error) {
 
 // GetMaxRound is part of idb.IndexerDB
 func (db *IndexerDb) GetMaxRound() (round uint64, err error) {
-	var nullableRound sql.NullInt64
-	round = 0
-	row := db.db.QueryRow(`SELECT max(round) FROM block_header`)
-	err = row.Scan(&nullableRound)
+	//var round uint64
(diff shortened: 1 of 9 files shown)


idb/postgres/postgres.go
```diff
@@ -398,13 +398,13 @@ func (db *IndexerDb) SetMetastate(key, jsonStrValue string) (err error) {
 
 // GetMaxRound is part of idb.IndexerDB
 func (db *IndexerDb) GetMaxRound() (round uint64, err error) {
-	var nullableRound sql.NullInt64
-	round = 0
-	row := db.db.QueryRow(`SELECT max(round) FROM block_header`)
-	err = row.Scan(&nullableRound)
+	//var round uint64
```
brianolson, line 401: daemon.go uses GetMaxRound() specifically for the blocks imported, not accounting

brianolson: I can't recommend changing the behavior of daemon data loading as a 'small change'.
I was hoping that _just_ /health would return account_round.
Though, I see that the majority of GetMaxRound() is attached to API calls, so maybe it should change, but without more testing I think daemon needs to still act based on _loaded_ data not _accounted_ data.

brianolson: maybe there should be GetMaxRoundAccounted() and GetMaxRoundLoaded()

winder: Good idea, now that you mention it there's no reason why both of those numbers couldn't be the same (and they often will). So we need a way to distinguish between them.

brianolson (approved):

PR #328, opened 2021-02-16
Change GetBlock to retrieve content in a single transaction

tsachiherman:
## Summary

Followup on the suspected GetBlock issue, this PR intend to perform both queries in a single transactional query.

## Test Plan

Unit tests added ( Thanks @winder !)

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -400,8 +400,7 @@ func (si *ServerImplementation) SearchForAssets(ctx echo.Context, params generat
 // LookupBlock returns the block for a given round number
 // (GET /v2/blocks/{round-number})
 func (si *ServerImplementation) LookupBlock(ctx echo.Context, roundNumber uint64) error {
-	filter := idb.TransactionFilter{Round: uint64Ptr(roundNumber)}
-	blk, err := si.fetchBlockAndTransactions(ctx.Request().Context(), roundNumber, filter)
+	blk, err := si.fetchBlockAndTransactions(ctx.Request().Context(), roundNumber)
 	if err != nil {
 		return indexerError(ctx, err.Error())
 	}
@@ -574,8 +573,8 @@ func (si *ServerImplementation) fetchAssetBalances(ctx context.Context, options 
 	return balances, nil
 }
 
-func (si *ServerImplementation) fetchBlockAndTransactions(ctx context.Context, round uint64, filter idb.TransactionFilter) (generated.Block, error) {
-	blk, transactions, err := si.db.BlockWithTransactions(ctx, round, filter)
+func (si *ServerImplementation) fetchBlockAndTransactions(ctx context.Context, round uint64) (generated.Block, error) {
+	blk, transactions, err := si.db.GetBlock(ctx, round, idb.GetBlockTransactions)
 	if err != nil {
 		return generated.Block{}, fmt.Errorf("%s '%d': %v", errLookingUpBlock, round, err)
 	}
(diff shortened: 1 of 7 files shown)


idb/dummy.go
```diff
@@ -233,6 +238,7 @@ type IndexerDb interface {
 	CommitRoundAccounting(updates RoundUpdates, round uint64, blockPtr *types.Block) (err error)
 
 	GetBlock(round uint64) (block types.Block, err error)
+	BlockWithTransactions(ctx context.Context, round uint64, tf TransactionFilter) (block types.Block, transactions []TxnRow, err error)
```
winder, line 241: If we add this feature to the interface I would probably merge this into `GetBlock`, adding in a flag (or "Options" object) to request that transactions be included in the result.

Also the TransactionFilter object isn't required, inside the function we would initialize the `TransactionFilter{Round:round}`

idb/postgres/postgres.go
```diff
@@ -1476,6 +1477,50 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	return
 }
 
+// BlockWithTransactions is part of idb.IndexerDB
```
winder, line 1480: The transaction code here can be replaced with a call to `txTransactions` to prevent the 3rd copy of this logic. `Transactions` should probably be refactored to call `txTransactions`. We could make a similar function to use common code for the block request.

api/handlers.go
```diff
@@ -400,19 +400,12 @@ func (si *ServerImplementation) SearchForAssets(ctx echo.Context, params generat
 // LookupBlock returns the block for a given round number
 // (GET /v2/blocks/{round-number})
 func (si *ServerImplementation) LookupBlock(ctx echo.Context, roundNumber uint64) error {
-	blk, err := si.fetchBlock(roundNumber)
```
winder, line 403: `fetchBlock` could now be deleted.

idb/dummy.go
winder: If we merge these two functions, the rest of my feedback probably no longer applies.

winder (commented):
Looks pretty good

idb/dummy.go
```diff
@@ -232,7 +231,7 @@ type IndexerDb interface {
 
 	CommitRoundAccounting(updates RoundUpdates, round uint64, blockPtr *types.Block) (err error)
 
-	GetBlock(round uint64) (block types.Block, err error)
+	GetBlock(ctx context.Context, round uint64, options ...GetBlockOptions) (block types.Block, transactions []TxnRow, err error)
```
winder, line 234: This is a neat pattern, but it's inconsistent with the other functions which use parameter objects. One example where this might be important is if the indexer ever supported co-chains, we might want one of the options to be a co-chain ID.

Since the API doesn't provide a way to pass in block parameters let's just leave off options for now and always fetch the transactions.

tsachiherman: adding the options was your idea.. but I don't think we can *always* fetch the transactions as it will have a negative performance impact on calls that doesn't need it.

winder:
For a test you could copy this one:
https://github.com/algorand/indexer/blob/develop/idb/postgres/postgres_integration_test.go#L147

Then change the assert to hit the block and transaction functions and make sure all transactions are returned.

idb/dummy.go
winder: In that case, then a parameter object instead of an int would be fine

tsachiherman: oh, I think that I get what you meant by parameter objects.. you meant like in `GetAccounts`.

winder: Yes exactly

tsachiherman: ok, updated.

winder (approved):

PR #329, opened 2021-02-16
Validator improvements

winder:
## Summary

Some validator improvements.

* Retry only updates indexer account data. Because indexer  is generally
one round behind, this allows busy accounts to be verified.
* Simplify processor interface by passing account data to process cmd.
* Configurable retry delay.
* Flag to enable curl command in error log to avoid leaking tokens.

## Test Plan

Ran against mainnet with a list of false-positive accounts to verify that the accounts are no longer flagged.

### DIFF
--- a/cmd/validator/dynamic_processor.go
+++ b/cmd/validator/dynamic_processor.go
@@ -3,8 +3,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
-	"net/http"
 	"reflect"
 	"sort"
 )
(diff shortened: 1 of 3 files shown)


cmd/validator/dynamic_processor.go
```diff
@@ -22,32 +20,34 @@ func mustEncode(data interface{}) string {
 }
 
 // ProcessAddress is the entrypoint for the DynamicProcessor.
-func (gp DynamicProcessor) ProcessAddress(addr string, config Params) (Result, error) {
-	indexerURL := fmt.Sprintf("%s:/v2/accounts/%s", config.indexerURL, addr)
-	indexerAcct, err := getResponse(indexerURL, config.indexerToken)
+func (gp DynamicProcessor) ProcessAddress(algodData, indexerData []byte) (Result, error) {
```
winder, line 23: account data now passed in, deserialize instead of making HTTP calls.

cmd/validator/struct_processor.go
```diff
@@ -17,74 +14,27 @@ import (
...
-	var response generated.Account
-	err = json.Unmarshal(data, &response)
-	return response, err
-}
-
 // ProcessAddress is the entrypoint for the StructProcessor
-func (gp StructProcessor) ProcessAddress(addr string, config Params) (Result, error) {
-	indexerURL := fmt.Sprintf("%s:/v2/accounts/%s", config.indexerURL, addr)
-	indexerAcct, err := getAccountIndexer(indexerURL, config.indexerToken)
+func (gp StructProcessor) ProcessAddress(algodData, indexerData []byte) (Result, error) {
```
winder, line 18: account data now passed in, deserialize instead of making HTTP calls.

algorotem (approved):

PR #330, opened 2021-02-16
Small refactoring / cleanup

winder:
## Summary

Some safe and relatively uncontroversial refactoring. Relates to #279 

## Test Plan

No new functionality to test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ services:
 - docker
 install:
 - go get -u golang.org/x/lint/golint
-- pip3 install boto3 markdown2
+- pip3 install -r misc/requirements.txt
+- echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 addons:
   apt:
     packages:
@@ -16,7 +17,6 @@ addons:
     - python3-setuptools
     - python3-wheel
 script:
-- echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 - make lint
 - make
 - make integration
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 go:
-- 1.13.x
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -3,28 +3,28 @@ go:
 - 1.13.x
```
brianolson, line 3: we should definitely drop 1.13 now. 1.16 is out and only 1.16 and 1.15 will be getting fixes now.

idb/postgres/postgres.go
```diff
@@ -375,8 +375,8 @@ func (db *IndexerDb) GetProto(version string) (proto types.ConsensusParams, err
 	return
 }
 
-// GetMetastate is part of idb.IndexerDB
-func (db *IndexerDb) GetMetastate(key string) (jsonStrValue string, err error) {
+// getMetastate is part of idb.IndexerDB
```
brianolson, line 378: drop this comment line

idb/postgres/postgres.go
```diff
@@ -390,12 +390,37 @@ func (db *IndexerDb) GetMetastate(key string) (jsonStrValue string, err error) {
 
 const setMetastateUpsert = `INSERT INTO metastate (k, v) VALUES ($1, $2) ON CONFLICT (k) DO UPDATE SET v = EXCLUDED.v`
 
-// SetMetastate is part of idb.IndexerDB
-func (db *IndexerDb) SetMetastate(key, jsonStrValue string) (err error) {
+// setMetastate is part of idb.IndexerDB
```
brianolson, line 393: also not quite true anymore

brianolson (commented):
lots of good stuf. a couple tweaks

.travis.yml
winder: Good idea. Dropped 1.13, looks like there's some minor work to get 1.16 working so I'll leave that off for now.

brianolson (approved):

PR #331, opened 2021-02-16
Remove InitialImport logic out of UpdateAccounting

winder:
## Summary

This PR builds on #330 with some riskier changes. Relates to #279 

The main change here is splitting the "InitialImport" stage out of "UpdateAccounting". This lets us avoid thinking about things like the path to the genesis file when processing new block data.

Parameters that narrow the set of blocks considered for the update have been moved into a parameter object.

We added a "round" parameter recently, I'm not exactly sure it's purpose. It looks like it serves as a "max round" to process, but we weren't checking it in all places where we commit round accounting.

## Test Plan

No new functionality.


cmd/algorand-indexer/daemon.go
```diff
@@ -175,7 +179,11 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	}
 	_, err := bih.imp.ImportDecodedBlock(block)
 	maybeFail(err, "ImportDecodedBlock %d: %v", block.Block.Round, err)
-	importer.UpdateAccounting(bih.db, bih.cache, uint64(block.Block.Round), genesisJSONPath, logger)
+	filter := idb.UpdateFilter{
+		StartRound: uint64(block.Block.Round)-1,
```
brianolson, line 183: This would break reset-rebuild.
Right now with UpdateAccounting doing its own fetch of accounting metastate it notices nothing is there and does the rebuild.

winder: do you think there's any value in doing a "catchup round accounting" elsewhere? I guess it would basically be this but run with:
```
    StartRound: db.GetMaxRoundAccounted() - 1,
    MaxRound: db.GetMaxRoundLoaded(),
```

brianolson: Yeah. It would be totally fair to explicitly do that before even trying to talk to algod. Expand the startup code to do more checks and things like that. Do the block-contiguiity check. Etc.

winder: I'm not going to take this on now, but I added it to the list in https://github.com/algorand/indexer/issues/279

cmd/algorand-indexer/daemon.go
```diff
@@ -174,8 +178,15 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
...
 	}
 	_, err := bih.imp.ImportDecodedBlock(block)
-	maybeFail(err, "ImportDecodedBlock %d: %v", block.Block.Round, err)
-	importer.UpdateAccounting(bih.db, bih.cache, uint64(block.Block.Round), genesisJSONPath, logger)
+	maybeFail(err, "ImportDecodedBlock %d", block.Block.Round)
+	maxRoundAccounted, err := bih.db.GetMaxRoundAccounted()
+	maybeFail(err, "failed to get max round accounted.")
+	// During normal operation StartRound and MaxRound will be the same round.
+	filter := idb.UpdateFilter{
+		StartRound: maxRoundAccounted + 1,
```
brianolson, line 186: I don't think this is consistent for the initial case. With no prior state, maxRoundAccounted will be 0, but yieldTxns returns > round, so round needs to be -1, so StartRound needs to be 0.

winder: I liked the name `StartRound` so I flipped the off-by-one to `prevRound` in UpdateAccounting:
```
txns := db.YieldTxns(context.Background(), int64(filter.StartRound)-1)
```

brianolson: I think this could miss txns in round=1 (or round=0) when rebuilding after reset.

winder: it took me a while but I finally see it... 0 vs none strikes again...

cmd/algorand-indexer/daemon.go
```diff
@@ -174,8 +180,22 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 		logger.Errorf("received block %d when expecting %d", block.Block.Round, bih.round+1)
 	}
 	_, err := bih.imp.ImportDecodedBlock(block)
-	maybeFail(err, "ImportDecodedBlock %d: %v", block.Block.Round, err)
```
tolikzinovyev, line 177: why not print `err`?

idb/postgres/postgres.go
```diff
@@ -425,15 +425,27 @@ func (db *IndexerDb) SetImportState(state idb.ImportState) (err error) {
 
 // GetMaxRoundAccounted is part of idb.IndexerDB
 func (db *IndexerDb) GetMaxRoundAccounted() (round uint64, err error) {
+	var nullableRound sql.NullInt64
+
 	//var round uint64
```
tolikzinovyev, line 430: could you remove this?

idb/postgres/postgres.go
```diff
@@ -554,9 +573,9 @@ func (db *IndexerDb) yieldTxnsThread(ctx context.Context, rows *sql.Rows, result
 }
 
 // YieldTxns is part of idb.IndexerDB
-func (db *IndexerDb) YieldTxns(ctx context.Context, prevRound int64) <-chan idb.TxnRow {
+func (db *IndexerDb) YieldTxns(ctx context.Context, firstRound uint64) <-chan idb.TxnRow {
 	results := make(chan idb.TxnRow, 1)
-	rows, err := db.db.QueryContext(ctx, yieldTxnQuery, prevRound)
+	rows, err := db.db.QueryContext(ctx, yieldTxnQuery, int64(firstRound) - 1)
```
tolikzinovyev, line 578: would it make sense to also modify `yieldTxnQuery` to use `firstRound` directly?

importer/helper.go
```diff
@@ -217,38 +236,41 @@ func loadGenesis(db idb.IndexerDb, in io.Reader) (err error) {
...
+// InitialImport imports the genesis block if needed. Returns true if the initial import occurred.
+func InitialImport(db idb.IndexerDb, genesisJSONPath string, l *log.Logger) bool {
 	state, err := db.GetImportState()
-	maybeFail(err, l, "getting import state, %v", err)
-	if state == nil {
-		state = &idb.ImportState{AccountRound: -1}
+
+	// Only import when the state is not found.
+	if err == idb.ErrorNotInitialized {
+		state.AccountRound = -1
```
tolikzinovyev, line 245: In general it would be nice to store "next round" instead of "last round" everywhere to avoid these issues.

tolikzinovyev (commented):
I didn't try to read carefully, but looks good I think.

cmd/algorand-indexer/daemon.go
brianolson: see the updated definition of maybeFail in main.go, it adds the error to the log message in a structured way

idb/postgres/postgres.go
winder: Yes, and I actually made an attempt to do it but there were subtle off-by-one issues so I decided not to tackle the problem in this PR.

importer/helper.go
winder: Agreed. Related to the comment above about `yieldTxnQuery`. I've been throwing comments about what I'd like to refactor into this PR and added an entry for this refactoring: https://github.com/algorand/indexer/issues/279

importer/helper.go
```diff
@@ -90,7 +90,26 @@ func (h *ImportHelper) Import(db idb.IndexerDb, args []string) {
 		h.Log.Infof("%d blocks in %s, %.0f/s, %d txn, %.0f/s", blocks, dt.String(), float64(time.Second)*float64(blocks)/float64(dt), txCount, float64(time.Second)*float64(txCount)/float64(dt))
 	}
 
-	accountingRounds, txnCount := updateAccounting(db, h.DefaultFrozenCache, 0, h.GenesisJSONPath, h.NumRoundsLimit, h.Log)
+	state, err := db.GetImportState()
+	var startRound uint64 = 0
+	accountingRounds := 0
+	if err == idb.ErrorNotInitialized {
+		if InitialImport(db, h.GenesisJSONPath, h.Log) {
+			accountingRounds ++
```
tolikzinovyev, line 98: I think this shouldn't count as a round. Then `InitialImport()` doesn't need a return value and should probably be renamed to `MaybeImportGenesis()`.

winder: The initialize code is pretty strange. Something like this suggestion would be great for a next step. At the moment this function is used by the importer as well as the daemon, so it isn't quite so easy to change. Since the importer is deprecated removing it would probably be next, and then we can redo the init process.

cmd/algorand-indexer/daemon.go
winder: Fixed by having GetMaxRoundAccounted return ErrorNotInitialized.

winder:
@brianolson @tolikzinovyev this is rebased and ready for another review. There are further improvements to make, but I think this is a good incremental step.

importer/helper.go
tolikzinovyev: The daemon doesn't use the result of `InitialImport()`. The change should be easy.

tolikzinovyev (approved):

brianolson (approved):

PR #333, opened 2021-02-17
Update supported consensus protocols - v25

tsachiherman:
## Summary

Update the consensus protocols support for the indexer; use the `genconsensusconfig` to regenerated the `protocols.json`, and manually update `ConsensusParams`.

## Test Plan

No tests around this.


types/README.md
```diff
@@ -0,0 +1,11 @@
+# Updating protocols.json
+
+The main go-algorand repo contains a tool named `genconsensusconfig` that generates the supported consensus protocols version file.
```
brianolson, line 3: pretty sure the current protocols.json was generated with `goal protocols`

brianolson: It might have the advantage of using go-codec to already emit sorted keys, no need for jq. Seems to produce the same diff otherwise.

winder:
In addition to adding the protocol version we also need to support the new protocol, sandbox is currently failing due to this as well, the error message is:
```
{"error":"msgpack decode error [pos 11]: no matching struct field found when decoding stream map with key cc","level":"error"
,"msg":"err handling catchup block 1","time":"2021-02-18T03:09:26Z"}
```

tsachiherman:
@winder @algobolson could you please take another look ?

brianolson (approved):

tsachiherman:
@winder I can see what I did wrong here that would failed the travis build. Any suggestions ?

winder (approved):
The latest changes look good

Issue #187, opened 2020-08-20
Unexpected DB Migration


winder:
## Summary
In a multi-reader postgres configuration I've been running a node with no algod parameters expecting it to effectively be in read-only mode. It turns out if there is a DB migration, the migration will be attempted regardless.

## Scope
This could be solved in a couple ways:
1) require a specific '--perform-migration' option, without which the migration will not be performed.
2) ~if there are no algod settings generate an error, so that the user must either set `--no-algod` (putting the daemon in read-only mode), or they must provide the algod settings (allowing a proper migration, if algod is needed for the migration).~

## Urgency
It's really easy to accidentally break the indexer DB without this feature. We should fix asap to avoid people breaking

ian-algorand:
Decided to go with the first approach, and to also detect if a migration is actually needed. If it's needed, throw an error and mention the new flag.

brianolson:
read-only mode, or no-algod mode, should definitely imply no-migrations

brianolson:
but with-algod mode, implying being a writer, should automatically do migrations (unless maybe --no-migrations is set? but I'm skeptical of the value of deliberately being stuck in the past with known bugs)

winder:
@brianolson the main thing is that we need to make it harder to accidentally have multiple nodes perform migrations. With the proposed solution it will be much harder because there's a special flag needed for the "migrator". The`/health` endpoint should display something to the effect of "this DB needs a migration" to hopefully let the admin know they need to do a migration, and inform users that the data is stuck in the past for some reason.

brianolson:
The normal case is to have one indexer. One writer indexer and several read-only indexers shouldn't be that hard to manage. I don't want to waste too much time preventing against operator error (multiple writers) especially when solving that right is a hard problem with little reward. I think doing migrations is the right default action and requiring to add "--actually-do-migrations" will make operation harder for more users and will leave databases unmigrated and mysteriously failing. If someone upgrades the software I don't want them to be left with the question, "I upgraded, why is it broken and not upgraded?"

PR #338, opened 2021-02-18
Reset command

brianolson:
## Summary

Productionize resetting the database. A 'reset' deletes all accounting and derived data while retaining original block_header and txn data (faster than re-loading block by block from the archival algod). When Indexer is again started as a daemon it would re-calculate all accounting from round 0 up to the latest block.

## Test Plan

Tested this manually working with the toy 50-round dataset.


cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,49 @@
...
+		var err error
+		config.BindFlags(cmd)
+		err = configureLogger()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to configure logger: %v", err)
+			os.Exit(1)
+		}
+		opts := idb.IndexerDbOptions{}
+		db := globalIndexerDb(&opts)
+		os.Stdout.Write([]byte("Are you sure? [y/N] "))
```
winder, line 31: Be more aggressive here.
```suggestion
		fmt.Println("Prior to resetting the database make sure no daemons are connected.")
		fmt.Println("After this command finishes start the daemon again and it will begin to")
		fmt.Println("re-index the data. This will take hours or days depending on the network")
		fmt.Println("size. During this time data will be incomplete, similar to the initial import.")
		
		fmt.Print("\nAre you sure you would like to reset? [y/N] "))
```

cmd/algorand-indexer/main.go
```diff
@@ -110,6 +110,7 @@ func init() {
 
 	rootCmd.AddCommand(importCmd)
 	rootCmd.AddCommand(daemonCmd)
+	rootCmd.AddCommand(resetCmd)
```
winder, line 113: I'm not sure we should advertise this feature.
```suggestion
	rootCmd.AddCommand(resetCmd)
	resetCmd.Hidden = true
```

winder (commented):
Looks mostly good

idb/postgres/reset.sql
```diff
@@ -0,0 +1,25 @@
+TRUNCATE account;
+TRUNCATE account_app;
+TRUNCATE account_asset;
+TRUNCATE app;
+TRUNCATE asset;
+TRUNCATE metastate;
+UPDATE txn SET extra = NULL WHERE extra IS NOT NULL;
+
+-- schema changes from various migrations, made re-run safe by 'IF NOT EXISTS'
```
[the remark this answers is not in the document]
brianolson: I set up some things to mostly automate this. For example the ALTER TABLE lines from m5 are now captured from code.

cmd/algorand-indexer/daemon.go
```diff
@@ -102,6 +102,17 @@ var daemonCmd = &cobra.Command{
 			}()
 		} else {
 			logger.Info("No block importer configured.")
+			waitForDBAvailable(db)
```
winder, line 105: Should this block of code go into the if? I think this code could cause read-only nodes to attempt to update the accounting.

winder: Besides this, it looks ready to merge in.

brianolson: Hm, we have a kinda weird 3-state situation
-no-algod : daemon runs in read-only mode
(nothing specified): daemon isn't read-only, but usually has nothing to do because there's no input
-algod/-algod-net/-algod-token: connects to algod and imports data

I thought this code would only be in the second case, but now I think it would run in the -no-algod case and that would be wrong. I think I'll move it into the reset subcommand and make an option there for `reset -and-rebuild`

brianolson: Ooo, I 'fixed a bug' that exposes that the integration test was relying on the 'nothing specified' behavior above. I changed the code to eliminate that gap so that case behaved like read-only; and the integration tests failed because they rely on running migrations in that state. I think I'll add back an explicit '--allow-migrations' option?

cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+		err = scanner.Err()
+		maybeFail(err, "getting prompt char")
+		fmt.Println("not resetting")
+	},
+}
+
+func init() {
+	resetCmd.Flags().BoolVarP(&andRebuild, "and-rebuild", "", false, "re-run accounting after reset (this could take a while)")
```
tolikzinovyev, line 88: It's a personal taste but "rebuild" would be more concise.

cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,90 @@
...
+		if scanner.Scan() {
+			answer := scanner.Text()
+			if len(answer) > 0 && (answer[0] == 'y' || answer[0] == 'Y') {
+				fmt.Println("Resetting...")
+				time.Sleep(2) // leave some ^C time
+				err = db.Reset()
+				maybeFail(err, "database reset failed")
+				if andRebuild {
+					fmt.Println("Done resetting. Re-building accounting...")
+					accountround, err := db.GetMaxRoundAccounted()
```
tolikzinovyev, line 64: If we just reset accounting, do we still need to query this?

idb/postgres/postgres.go
```diff
@@ -81,7 +81,7 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
 	// e.g. a user named "readonly" is in the connection string
 	readonly := (opts != nil) && opts.ReadOnly
 	if !readonly {
-		err = pdb.init()
+		err = pdb.init(opts)
```
tolikzinovyev, line 84:
```suggestion
    pdb.GetSpecialAccounts()
    if !noMigration {
      err = pdb.maybeRunMigrations()
    }
```

idb/postgres/postgres.go
```diff
@@ -107,15 +107,16 @@ type IndexerDb struct {
 	accountingLock sync.Mutex
 }
 
-func (db *IndexerDb) init() (err error) {
+func (db *IndexerDb) init(opts *idb.IndexerDbOptions) (err error) {
```
tolikzinovyev, line 110:
```suggestion
func (db *IndexerDb) maybeRunMigrations() (err error) {
```

idb/postgres/postgres.go
```diff
@@ -130,6 +131,28 @@ func (db *IndexerDb) init() (err error) {
...
 
+// Reset is part of idb.IndexerDB
+func (db *IndexerDb) Reset() (err error) {
+	// new database, run setup
+	_, err = db.db.Exec(reset_sql)
+	if err != nil {
+		return fmt.Errorf("db reset failed, %v", err)
+	}
+	db.log.Debugf("reset.sql done")
+	for i, cmd := range registerAlterTableSQLLines {
```
tolikzinovyev, line 142: Do we really need to alter the tables? If we are removing all accounting data anyway, can we just drop the tables and create new ones with the latest schema versions?

tolikzinovyev (changes requested):
Sorry for the delay. Been busy with other things.

cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to configure logger: %v", err)
+			os.Exit(1)
+		}
+		opts := idb.IndexerDbOptions{NoMigrate: true}
+		db := globalIndexerDb(&opts)
+		timeout := time.Now().Add(5 * time.Second)
+		health, err := db.Health()
+		maybeFail(err, "could not get db health, %v", err)
+		for health.IsMigrating || !health.DBAvailable {
```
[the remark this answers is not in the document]
brianolson: I wrote this before I added the NoMigrate:true option to the database init. It might not be necessary but it shouldn't be wrong to leave it in.

cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,90 @@
...
+		fmt.Println("re-index the data. This can take hours or days depending on the network")
+		fmt.Println("size. During this time data will be incomplete, similar to the initial import.")
+
+		fmt.Print("\nAre you sure you would like to reset? [y/N] ")
+		scanner := bufio.NewScanner(os.Stdin)
+		if scanner.Scan() {
+			answer := scanner.Text()
+			if len(answer) > 0 && (answer[0] == 'y' || answer[0] == 'Y') {
+				fmt.Println("Resetting...")
+				time.Sleep(2) // leave some ^C time
```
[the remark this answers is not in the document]
brianolson: Because I've been adminning things and hit return and half a second later hit ^C to try and stop it.

cmd/algorand-indexer/reset.go
brianolson: Maybe. I thought `algorand-indexer reset -and-rebuild` read nicely, but maybe that's too cute?

cmd/algorand-indexer/daemon.go
winder: I missed this change before, +1 to --allow-migrations.

It means we can also close this ticket: https://github.com/algorand/indexer/issues/187

idb/postgres/postgres.go
winder: Good idea. I guess this would mean changing reset.sql to drop the tables instead of truncating. Do we need anything else?

brianolson: This is probably a good idea. I should branch my branch to write that code and see if it works out well.

brianolson:
replaced by #455

PR #344, opened 2021-02-19
Fix block header for protocol upgrade.

winder:
## Summary

Some of the new compact cert fields were changed since we added them. I'm not sure why the tests don't catch this, they may use the stable channel. I found them due to an incompatibility with one of the sandbox configurations.

## Test Plan

I started a private network with the nightly channel and connected indexer built from this branch, I see that it is now able to process blocks.


brianolson (approved):
Yeah, the problem with testing from go-algorand `future` protocol is that it can change before it's done. The Compact Cert feature changed its fields after I had already started to import them.

PR #345, opened 2021-02-21
Display Fetcher Errors in Health Check

winder:
## Summary

API Change - health endpoint has an optional "errors" field that can be an array of errors.
Fetcher now stores the most recent error it has encountered.

(not happy about this part - open to suggestions)
Fetcher implements the `error` interface and is passed into the REST serve function.
The health endpoint checks for DB health and also checks if the fetcher "error" is not an empty string.

## Test Plan

To force a fetcher error I changed one of the round field name from "rnd" to "rand", then ran the e2e tests. Now that they print out the health response you can see the error:
```
~$ python3 misc/e2elive.py --connection-string 'host=localhost port=5432 dbname=postgres sslmode=disable user=postgres password=postgres' --indexer-bin cmd/algorand-indexer/algorand-indexer --indexer-port 9890
INFO:util:s3://algorand-testdata/indexer/e2e2/fe340d7e/net_done.tar.bz2 -> /tmp/tmpsl4jidmd/net_done.tar.bz2
ERROR:__main__:could not get indexer health, or did not reach round=58
b'{"data":{"migration-required":false,"read-only-node":true},"db-available":true,"errors":["unable to decode block: msgpack decode error [pos 171]: no matching struct field found when decoding stream map with key rnd"],"is-migrating":false,"message":"0","round":0}\n'
...
```


api/server.go
```diff
@@ -20,7 +20,7 @@ import (
 var indexerDb idb.IndexerDb
 
 // Serve starts an http server for the indexer API. This call blocks.
-func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, log *log.Logger, tokens []string, developerMode bool) {
+func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError error, log *log.Logger, tokens []string, developerMode bool) {
```
brianolson, line 23: was confused for a minute because I expected a function pointer or interface that _returned_ `error` so that it could return `nil` if there was no error yet. Then I got to Fetcher _implementing_ error and I get it now. My bias says `err == nil` is better than `foo.Error() == ""`, but probably this is fine.

fetcher/fetcher.go
```diff
@@ -86,12 +99,14 @@ func (bot *fetcherImpl) catchupLoop() {
 
 		blockbytes, err = aclient.BlockRaw(bot.nextRound).Do(context.Background())
 		if err != nil {
+			bot.err = err
```
brianolson, line 102: Looks like bot.err is never cleared. Which of these errors recoverable? If normal progress resumes we should clear bot.err.

api/server.go
winder: Yeah, this is a total hack... probably some sort of error registry would be better

fetcher/fetcher.go
winder: Good idea, each time we finish processing a block I clear this field now.

winder:
@brianolson updated the PR based on your feedback.

brianolson (approved):

Issue #310, opened 2021-02-08
Refuse to return balance for Rewards/FeeSink


brianolson:
## Summary
Indexer cannot rewind the special accounts for Rewards and the FeeSink, and Indexer's accounting is slightly different than `algod`, so to not be confusing, refuse to return any data for these accounts.
See also #300

## Scope/Requirements
API code looks up special accounts from genesis block, takes exception to them and doesn't return data when requested.

## Urgency/Relative Priority
Easy/soon.

PR #348, opened 2021-02-22
Make sure special accounts are not returned in results 

algorotem:
## Summary

Solves #310 

## Test Plan
integration test for both endpoints

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -81,6 +81,16 @@ func (si *ServerImplementation) LookupAccountByID(ctx echo.Context, accountID st
 		return badRequest(ctx, errors[0])
 	}
 
+	// Special accounts non handling
+	isSpecialAccount, err := si.isSpecialAccount(accountID)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


api/handlers.go
```diff
@@ -706,3 +726,18 @@ func max(x, y uint64) uint64 {
 	}
 	return y
 }
+
+// isSpecialAccount returns true if addr belongs to a special account, false otherwise.
+// The function returns an error in case it fails to retrieve the special accounts list.
+func (si *ServerImplementation) isSpecialAccount(addr string) (bool, error) {
+	// Special accounts non handling
+	sa, err := si.db.GetSpecialAccounts()
```
brianolson, line 734: probably worth caching this in a field of ServerImplementation so we don't hit the db every time.

algorotem: I thought about it but the db implementation already caches it, should we cash it twice?

brianolson: Oh oops, yeah postgres.go caching is enough

brianolson (approved):

winder (approved):
LGTM

PR #350, opened 2021-02-22
Ensure indexer doesn't return deleted entries unless requested explicitly 

algorotem:
## Summary

## Test Plan


brianolson:
So far so good.
Should a single account lookup `/v2/accounts/AAAA...` implicitly fetch that account even if it has been deleted?

algorotem:
> So far so good.
> Should a single account lookup `/v2/accounts/AAAA...` implicitly fetch that account even if it has been deleted?

Will and I discussed that and we think that the answer is no. The rationale was to keep it consistent.

test/common.sh
```diff
@@ -405,11 +424,28 @@ function create_delete_tests() {
       "t|13|37|82"
     rest_test "[rest] app create & delete (app-id=82)" \
       "/v2/applications/82?pretty" \
+      404 \
+      true \
+      ''
+    rest_test "[rest] app create & delete (app-id=82)" \
+      "/v2/applications/82?pretty&include-deleted=true" \
```
winder, line 431: I wasn't able to figure out how to get OpenAPI/oapi-codegen to use `include-deleted` as a flag.

idb/postgres/postgres_migrations.go
```diff
@@ -655,6 +656,7 @@ func processAccountTransactions(txnrows <-chan idb.TxnRow, addressStr string, ad
 
 		if accounting.AssetCreateTxn(stxn) {
 			result.asset[txn.AssetID] = updateCreate(result.asset[txn.AssetID], txn.Round)
+			result.assetHolding[txn.AssetID] = updateCreate(result.assetHolding[txn.AssetID], txn.Round)
```
winder, line 659: Accounting bug 1, we weren't marking deleted = false for the creator's asset holding.

idb/postgres/postgres_migrations.go
```diff
@@ -665,7 +667,7 @@ func processAccountTransactions(txnrows <-chan idb.TxnRow, addressStr string, ad
 			result.assetHolding[txn.AssetID] = updateCreate(result.assetHolding[txn.AssetID], txn.Round)
 		}
 
-		if accounting.AssetOptOutTxn(stxn) {
+		if accounting.AssetOptOutTxn(stxn) && stxn.Txn.Sender == address {
```
winder, line 670: Migration accounting bug 2, we were marking all accounts referenced in asset closes as deleted = true instead of just the sender.

winder, line 670: Migration accounting bug 2, we were marking all accounts referenced in asset closes as deleted = true instead of just the sender.

idb/postgres/postgres.go
```diff
@@ -2440,16 +2444,29 @@ func (db *IndexerDb) buildAccountQuery(opts idb.AccountQueryOptions) (query stri
 	// TODO: asset holdings and asset params are optional, but practically always used. Either make them actually always on, or make app-global and app-local clauses also optional (they are currently always on).
 	withClauses = append(withClauses, "qaccounts AS ("+query+")")
 	query = "WITH " + strings.Join(withClauses, ", ")
-	if opts.IncludeAssetHoldings {
-		query += `, qaa AS (SELECT xa.addr, json_agg(aa.assetid) as haid, json_agg(aa.amount) as hamt, json_agg(aa.frozen) as hf, json_agg(aa.created_at) as holding_created_at, json_agg(aa.closed_at) as holding_closed_at, json_agg(aa.deleted) as holding_deleted FROM account_asset aa JOIN qaccounts xa ON aa.addr = xa.addr GROUP BY 1)`
-	}
-	if opts.IncludeAssetParams {
-		query += `, qap AS (SELECT ya.addr, json_agg(ap.index) as paid, json_agg(ap.params) as pp, json_agg(ap.created_at) as asset_created_at, json_agg(ap.closed_at) as asset_closed_at, json_agg(ap.deleted) as asset_deleted FROM asset ap JOIN qaccounts ya ON ap.creator_addr = ya.addr GROUP BY 1)`
+	if opts.IncludeDeleted {
```
winder, line 2447: Toggles whether `WHERE coalesce(___.deleted, false) = false` is in the query or not. This is already a little out of hand, we need to be careful about what we add here.

Should default the account endpoint to excluding deleted things with no option to include them? The creatable-specific endpoints could have the option and the implementation would be cleaner.

idb/dummy.go
```diff
@@ -335,6 +335,9 @@ type AccountQueryOptions struct {
 	IncludeAssetHoldings bool
 	IncludeAssetParams   bool
 
+	// IncludeDeleted indicated whether to include deleted Assets, Applications, etc within the account.
```
algorotem, line 338: leaving a comment here as a reminer in case we decide to have the original names.

idb/postgres/postgres.go
brianolson: This is getting a bit nuts. Would it be easier to client-side-filter some of these things for deletion? Is it weird that IncludeDeleted becomes recursive on an account query and is applied to whether sub-fields of an account are deleted or not?

brianolson (commented):
99% ready? maybe 100% ready?

brianolson (approved):
I can't think of any way to make this better right now, there are some comments about how to maybe make it better in the future.

PR #354, opened 2021-02-23
Make sure the special accounts are initialized.

winder:
## Summary

The special accounts weren't being included in the create/delete migration. Their values were null.

## Test Plan

Integration test.


brianolson:
weren't m5 and m6 already released? they've been run by deploys that would run them and won't be run again?

winder:
@brianolson yes and no. 
Yes, there have been some limited releases, they all used a sneaky migration.
No, 2.3.1 never made it out of beta.

I added the proper migration for you which should cover all cases.

idb/postgres/postgres_migrations.go
```diff
@@ -1142,3 +1166,41 @@ func m4MarkTxnJSONSplit(db *IndexerDb, state *MigrationState) error {
...
+		return fmt.Errorf("failed to begin m9 migration: %v", err)
+	}
+	defer tx.Rollback() // ignored if .Commit() first
+
+	initstmt, err := tx.Prepare(`UPDATE account SET deleted=false, created_at=0 WHERE addr=$1`)
+	if err != nil {
+		return fmt.Errorf("failed to prepare m9 query: %v", err)
+	}
+
+	for _, account := range []string{accounts.FeeSink.String(), accounts.RewardsPool.String()} {
```
brianolson, line 1187: why bounce off .String() and then back through DecodeAddress ?

idb/postgres/postgres_migrations.go
```diff
@@ -711,7 +734,7 @@ func processAccountTransactions(txnrows <-chan idb.TxnRow, addressStr string, ad
 // Note: These queries only work if closed_at was reset before the migration is started. That is true
 //       for the initial migration, but if we need to reuse it in the future we'll need to fix the queries
 //       or redo the query.
-func m5RewardsAndDatesPart2UpdateAccounts(db *IndexerDb, state *MigrationState, accounts []string, finalBatch bool) error {
+func m5RewardsAndDatesPart2UpdateAccounts(db *IndexerDb, state *MigrationState, accounts []string, specialAccounts []string, finalBatch bool) error {
```
brianolson, line 737: we don't need to change m5, m9 will do it

brianolson (commented):
I think we don't need to hack m5/m6 to duplicate what's being done in the new m9.

PR #360, opened 2021-02-25
Add a round accounting lock to synchronize background migrations.

winder:
## Summary

A quick fix that should allow migrations to synchronize themselves with CommitRoundAccounting.

## Test Plan

No functional changes, existing tests.


brianolson (approved):
I think this should be safe. I still kinda want to see more retry logic around a bunch of operations, but this should make the system safer by not even trying to do a thing that could get it tripped up.

PR #362, opened 2021-02-25
Fix a bug: rekeying to our own address should reset the spending key

### LINKED ISSUE #337: Omit Auth-Addr when it is the same as the address
## Summary
When an account is rekeyed, and then rekeyed back to the original account, Indexer incorrectly returns an auth-addr field equal to the address field. The correct behavior would be to omit the auth-addr field.

## Scope/Requirements
1. In the API layer check if the AuthAddr equals the Address. If it is set AuthAddr to nil.
2. I'm not sure it's worth the extra effort in accounting.go/postgres.go, but we could set the "spend" key to nil in accounting.go and use that in postgres.go to remove the key altogether.

## Urgency/Relative Priority
Low priority, this only affects 2 accounts on mainnet, and it's technically still correct (though maybe unusual that we would return accounts rekeyed to themselves when using the rekey feature).

tolikzinovyev:
## Summary
Fixes #337. This modifies the database creation logic. If we rekey to our own address, remove the spending key from the account_data json field in the account table.

## Test Plan
Added two integration tests. One tests the basic rekey functionality. The other checks that rekeying to our own address removes the spending key.

### DIFF
--- a/accounting/accounting.go
+++ b/accounting/accounting.go
@@ -125,10 +125,19 @@ func (accounting *State) updateAccountType(addr types.Address, ktype string) {
 func (accounting *State) updateAccountData(addr types.Address, key string, field interface{}) {
 	au, ok := accounting.AccountDataUpdates[addr]
 	if !ok {
-		au = make(map[string]interface{})
+		au = make(map[string]idb.AccountDataUpdate)
 		accounting.AccountDataUpdates[addr] = au
 	}
-	au[key] = field
+	au[key] = idb.AccountDataUpdate{true, field}
+}
+
+func (accounting *State) removeAccountData(addr types.Address, key string) {
+	au, ok := accounting.AccountDataUpdates[addr]
+	if !ok {
+		au = make(map[string]idb.AccountDataUpdate)
+		accounting.AccountDataUpdates[addr] = au
+	}
+	au[key] = idb.AccountDataUpdate{false, struct{}{}}
 }
 
 func (accounting *State) updateAsset(addr types.Address, assetID uint64, add, sub uint64) {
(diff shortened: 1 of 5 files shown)


idb/postgres/postgres_integration_test.go
```diff
@@ -381,7 +382,8 @@ func TestMultipleWriters(t *testing.T) {
 	///////////
 	// Given // Send amt to AccountA
 	///////////
-	_, payAccountA := test.MakePayTxnRowOrPanic(test.Round, 1000, amt, 0, 0, 0, 0, test.AccountD, test.AccountA, types.ZeroAddress)
+	_, payAccountA := test.MakePayTxnRowOrPanic(test.Round, 1000, amt, 0, 0, 0, 0, test.AccountD,
+																							test.AccountA, types.ZeroAddress, types.ZeroAddress)
```
winder, line 386: too many tabs here

idb/dummy.go
```diff
@@ -503,6 +492,12 @@ type AlgoUpdate struct {
 	Closed bool
 }
 
+// Encodes an update or remove operation on an account_data json key.
```
winder, line 495: The linter is complaining because it wants this comment to start with `AccountDataUpdate`

winder (changes requested):
Looks good, just some minor requests

winder (approved):

idb/dummy.go
```diff
@@ -503,6 +492,12 @@ type AlgoUpdate struct {
 	Closed bool
 }
 
+// AccountDataUpdate encodes an update or remove operation on an account_data json key.
+type AccountDataUpdate struct {
+	Update bool        // true if update, false if delete
```
algorotem, line 497: Consider changing this field to `Delete`. That way, the code won't have mixed semantics (Using the term update for deletion).

accounting/accounting.go
```diff
@@ -125,10 +125,19 @@ func (accounting *State) updateAccountType(addr types.Address, ktype string) {
...
+	au[key] = idb.AccountDataUpdate{true, field}
+}
+
+func (accounting *State) removeAccountData(addr types.Address, key string) {
+	au, ok := accounting.AccountDataUpdates[addr]
+	if !ok {
+		au = make(map[string]idb.AccountDataUpdate)
+		accounting.AccountDataUpdates[addr] = au
+	}
+	au[key] = idb.AccountDataUpdate{false, struct{}{}}
```
algorotem, line 140: Optional - With the struct change, you can also easily merge these two functions to match the struct.

algorotem (changes requested):

idb/dummy.go
tolikzinovyev: Good idea.

algorotem (approved):

PR #364, opened 2021-02-26
Minor validator tool improvements

winder:
## Summary

Minor improvements to validator tool:
- Use `http://` scheme if none was set.
- Set status return code to 1 if there were any errors

## Test Plan

Manual tested.

### DIFF
--- a/cmd/validator/validate_accounting.go
+++ b/cmd/validator/validate_accounting.go
@@ -312,18 +312,18 @@ func resultsPrinter(config Params, printCurl bool, results <-chan Result) int {
 
 		numResults++
 		numRetries += r.Retries
-		if r.Error != nil || !r.Equal {
+
+		if r.Error != nil {
+			// Print error message if there is one.
+			numErrors++
+			errorLog.Printf("Processor error: %v\n", r.Error)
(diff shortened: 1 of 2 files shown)


cmd/validator/validate_accounting.go
```diff
@@ -307,7 +320,9 @@ func resultsPrinter(config Params, printCurl bool, results <-chan Result) {
 
 			// Print error message if there is one.
 			if r.Error != nil {
+				numErrors++
 				errorLog.Printf("Processor error: %v\n", r.Error)
+				continue
 			}
 			// Print error details if there are any.
```
tolikzinovyev, line 327: It would be nice to move everything after this under an else statement. Not a fan of continue. :)

tolikzinovyev, line 327: So if r.Error == nil, !r.Equal but r.Details == nil, we don't need to increment numErrors?

winder: I'll make the error increment unconditional here. There should always be details but no need to depend on it.

tolikzinovyev (approved):

PR #367, opened 2021-03-01
M6 migration page accounts/transactions

winder:
## Summary

Page through transactions and accounts rather than having large queries. This should help with databases that cache results to disk agressively.

## Test Plan

Existing integration tests. I also set the batch size very small in some manual tests to force extra paging.


idb/postgres/postgres_migrations.go
```diff
@@ -575,18 +637,48 @@ func initM5AccountData() *m5AccountData {
...
+	out := make(chan idb.TxnRow, 10)
+
+	go func() {
+		defer close(out)
+		token := ""
+		for true {
+			txnrows := db.Transactions(ctx, idb.TransactionFilter{
+				Address:   address[:],
+				MaxRound:  finalRound,
+				Limit:     1,
```
brianolson, line 652: I'd do a larger batch, ~1000, fetch them all into RAM quickly so as to let the sql transaction close and not be held open, then yield results from that buffer to the output chan

winder: oops, this low value was for testing.

PR #372, opened 2021-03-02
Attempt to fix mlock error when running in docker

onetechnical:
## Summary

An mlock error may be returned occasionally when running golang in Docker, due to kernel version.

To mitigate this, golang was updated to 1.15, which handles this error better than 1.14.

## Test Plan

Verify Travis tests pass repeatedly.


onetechnical:
Not sure how to replicate the initial issue. I've re-run this build multiple times, but I don't think that's a valid test.

winder (approved):
LGTM

PR #373, opened 2021-03-02
Support databases behind loadbalancers (API handlers)

### LINKED ISSUE #351: Support databases behind loadbalancers
## Summary
Currently, indexer assumes it talks to one database at a time. When connecting to a load balancer that has several readers behind, the API results may be incorrect because sometimes an API call can make more than one database call and these can end up on separate databases which may not be fully synced. 

## Scope/Requirements
There is a need to go over each of the API endpoints and, check which makes more than one database call, and make sure it happens on a single transaction to ensure the use of the same connection (and therefore the same db). 

## Urgency/Relative Priority
It's not time-critical but important to have in order to support that kind of deployment.

tolikzinovyev:
## Summary

Some work towards fixing #351. Rewinding can return an error if the database returns stale transaction list. Accounting when importing a block still needs to be fixed. 

## Test Plan

Unit tests for rewind.go. All other changes are trivial.

### DIFF
--- a/accounting/rewind_test.go
+++ b/accounting/rewind_test.go
@@ -1,6 +1,7 @@
 package accounting
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/algorand/go-algorand-sdk/encoding/msgpack"
@@ -57,8 +58,12 @@ func TestBasic(t *testing.T) {
 
 // Test that when idb.Transactions() returns stale data the first time, we return an error.
 func TestStaleTransactions1(t *testing.T) {
+	var a types.Address
+	a[0] = 'a'
+
 	account := models.Account{
-		Round: 8,
+		Address: a.String(),
+		Round:   8,
 	}
 
 	var outCh <-chan idb.TxnRow
@@ -68,12 +73,16 @@ func TestStaleTransactions1(t *testing.T) {
 	db.On("Transactions", mock.Anything, mock.Anything).Return(outCh, uint64(7)).Once()
 
 	account, err := AccountAtRound(account, 6, db)
-	assert.Error(t, err)
+	assert.True(t, errors.As(err, &ConsistencyError{}), "err: %v", err)
 }
 
 // Test that when idb.Transactions() returns stale data the second time, we return an error.
 func TestStaleTransactions2(t *testing.T) {
+	var a types.Address
+	a[0] = 'a'
+
 	account := models.Account{
+		Address:                     a.String(),
 		Amount:                      100,
 		AmountWithoutPendingRewards: 100,
 		Round:                       8,
@@ -102,5 +111,5 @@ func TestStaleTransactions2(t *testing.T) {
 	db.On("Transactions", mock.Anything, mock.Anything).Return(outCh, uint64(5)).Once()
 
 	account, err := AccountAtRound(account, 6, db)
-	assert.Error(t, err)
+	assert.True(t, errors.As(err, &ConsistencyError{}), "err: %v", err)
 }
--- a/accounting/rewind_test.go
+++ b/accounting/rewind_test.go
@@ -18,10 +18,10 @@ func TestBasic(t *testing.T) {
 	a[0] = 'a'
 
 	account := models.Account{
-		Address: a.String(),
-		Amount: 100,
+		Address:                     a.String(),
+		Amount:                      100,
 		AmountWithoutPendingRewards: 100,
-		Round: 8,
+		Round:                       8,
 	}
 
 	txnBytes := msgpack.Encode(types.SignedTxnWithAD{
@@ -36,7 +36,7 @@ func TestBasic(t *testing.T) {
 		},
 	})
 	txnRow := idb.TxnRow{
-		Round: 7,
+		Round:    7,
 		TxnBytes: txnBytes,
 	}
 
@@ -74,9 +74,9 @@ func TestStaleTransactions1(t *testing.T) {
 // Test that when idb.Transactions() returns stale data the second time, we return an error.
 func TestStaleTransactions2(t *testing.T) {
 	account := models.Account{
-		Amount: 100,
+		Amount:                      100,
 		AmountWithoutPendingRewards: 100,
-		Round: 8,
+		Round:                       8,
 	}
 
 	txnBytes := msgpack.Encode(types.SignedTxnWithAD{
@@ -87,7 +87,7 @@ func TestStaleTransactions2(t *testing.T) {
 		},
 	})
 	txnRow := idb.TxnRow{
-		Round: 7,
+		Round:    7,
 		TxnBytes: txnBytes,
 	}
 
--- a/accounting/rewind_test.go
+++ b/accounting/rewind_test.go
@@ -1,0 +1,106 @@
+package accounting
+
+import (
+	"testing"
+
+	"github.com/algorand/go-algorand-sdk/encoding/msgpack"
+	"github.com/algorand/go-algorand-sdk/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+
+	models "github.com/algorand/indexer/api/generated/v2"
+	"github.com/algorand/indexer/idb"
+	"github.com/algorand/indexer/idb/mocks"
+)
+
+func TestBasic(t *testing.T) {
+	var a types.Address
+	a[0] = 'a'
+
+	account := models.Account{
+		Address: a.String(),
+		Amount: 100,
+		AmountWithoutPendingRewards: 100,
+		Round: 8,
+	}
+
+	txnBytes := msgpack.Encode(types.SignedTxnWithAD{
+		SignedTxn: types.SignedTxn{
+			Txn: types.Transaction{
+				Type: types.PaymentTx,
+				PaymentTxnFields: types.PaymentTxnFields{
+					Receiver: a,
+					Amount:   2,
+				},
+			},
+		},
+	})
+	txnRow := idb.TxnRow{
+		Round: 7,
+		TxnBytes: txnBytes,
+	}
+
+	ch := make(chan idb.TxnRow, 1)
+	ch <- txnRow
+	close(ch)
+	var outCh <-chan idb.TxnRow = ch
+
+	db := &mocks.IndexerDb{}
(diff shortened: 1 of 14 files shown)


idb/dummy.go
```diff
@@ -217,29 +235,34 @@ type IndexerDb interface {
...
-	GetProto(version string) (proto types.ConsensusParams, err error)
+	GetProtoTx(tx *sql.Tx, version string) (proto types.ConsensusParams, err error)
 
 	GetImportState() (is *ImportState, err error)
 	SetImportState(ImportState) (err error)
-	GetMaxRoundAccounted() (round uint64, err error)
+	GetMaxRoundAccountedTx(tx *sql.Tx) (round uint64, err error)
 	GetMaxRoundLoaded() (round uint64, err error)
-	GetSpecialAccounts() (SpecialAccounts, error)
+	GetSpecialAccountsTx(ctx context.Context, tx *sql.Tx) (SpecialAccounts, error)
```
winder, line 244: The context is for operations which utilize a channel for some potentially long operation, I don't think you need to add that here or to the Health endpoint.

idb/dummy.go
```diff
@@ -217,29 +235,34 @@ type IndexerDb interface {
...
 
-	Transactions(ctx context.Context, tf TransactionFilter) <-chan TxnRow
-	GetAccounts(ctx context.Context, opts AccountQueryOptions) <-chan AccountRow
-	Assets(ctx context.Context, filter AssetsQuery) <-chan AssetRow
-	AssetBalances(ctx context.Context, abq AssetBalanceQuery) <-chan AssetBalanceRow
-	Applications(ctx context.Context, filter *models.SearchForApplicationsParams) <-chan ApplicationRow
+	Health(ctx context.Context) (status Health, err error)
 
-	Health() (status Health, err error)
+	MakeTx(ctx context.Context) (*sql.Tx, error)
```
winder, line 265: Ah, you're right we need some sort of interface around `Begin` in order for the DB interface to work... I hadn't considered that.

We would also need a wrapper around `sql.Tx` to finish the abstraction or else we can only support sql databases.

Given this problem what do you think about changing the interface like this?
```diff
-	Transactions(ctx context.Context, tf TransactionFilter) <-chan TxnRow
+	Transactions(ctx context.Context, tf TransactionFilter) (<-chan TxnRow, currentRound)
```

The API layer can return an error (or retry) if it needs multiple calls and the round doesn't match.

idb/dummy.go
```diff
@@ -217,29 +235,34 @@ type IndexerDb interface {
...
 	// YieldTxns returns a channel that produces the whole transaction stream after some round forward
 	YieldTxns(ctx context.Context, prevRound int64) <-chan TxnRow
 
 	CommitRoundAccounting(updates RoundUpdates, round uint64, blockPtr *types.Block) (err error)
 
-	GetBlock(ctx context.Context, round uint64, options GetBlockOptions) (block types.Block, transactions []TxnRow, err error)
+	GetBlock(ctx context.Context, round uint64,
+		options GetBlockOptions) (block types.Block, transactions []TxnRow, err error)
+	GetBlockTx(ctx context.Context, tx *sql.Tx, round uint64,
+		options GetBlockOptions) (block types.Block, transactions []TxnRow, err error)
```
winder, line 255: Is there a way to avoid needing 2 versions of each function? Maybe we can check for `tx==nil` and automatically initialize it when needed?

idb/postgres/postgres.go
```diff
@@ -2362,13 +2419,26 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
...
 		out <- idb.AccountRow{Error: err}
 		close(out)
-		tx.Rollback()
 		return out
 	}
 	go db.yieldAccountsThread(req)
 	return out
 }
 
+func (db *IndexerDb) getAccounts(opts idb.AccountQueryOptions) <-chan idb.AccountRow {
```
winder, line 2428: We should update the old code to use the new signature instead of adding wrapper functions.

winder (commented):
Looks like injecting the transaction is getting pretty complicated already and doubling the size of the interface. Maybe we can find something easier to implement.

winder (commented):
Looks good

idb/postgres/postgres.go
```diff
@@ -423,9 +423,20 @@ func (db *IndexerDb) SetImportState(state idb.ImportState) (err error) {
 	return db.setMetastate(stateMetastateKey, string(json.Encode(state)))
 }
 
+func (db *IndexerDb) getMaxRoundAccountedTx(tx *sql.Tx) (round uint64, err error) {
```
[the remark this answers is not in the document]
tolikzinovyev: That's a good idea.

idb/postgres/postgres.go
```diff
@@ -1527,28 +1538,47 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 }
 
 // Transactions is part of idb.IndexerDB
-func (db *IndexerDb) Transactions(ctx context.Context, tf idb.TransactionFilter) <-chan idb.TxnRow {
+func (db *IndexerDb) Transactions(ctx context.Context,
+	tf idb.TransactionFilter) (<-chan idb.TxnRow, uint64) {
 	out := make(chan idb.TxnRow, 1)
+
+	tx, err := db.db.BeginTx(ctx, &readonlySerializable)
```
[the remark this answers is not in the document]
tolikzinovyev: I think we need to see the read only flag. See https://github.com/algorand/indexer/issues/181 . I haven't thought about a weaker isolation level yet. Maybe, we could change it in a different pull request.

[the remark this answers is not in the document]
winder: I ask about the isolation level mainly because you changed it in this PR by removing `readOnlyTx`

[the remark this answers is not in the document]
tolikzinovyev: readOnlyTx, which is the lowest isolation level, is not enough I think.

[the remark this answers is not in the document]
tolikzinovyev: Actually, it's not the lowest, it's the default. In either case, we can set it to serializable and then think if we can lower it later.

api/handlers.go
```diff
@@ -639,7 +639,7 @@ func (si *ServerImplementation) fetchAccounts(ctx context.Context, options idb.A
 
 		// Compute for a given round if requested.
 		var account generated.Account
-		if atRound != nil {
+		if (atRound != nil) && (*atRound < row.Account.Round) {
```
winder, line 642: requesting a round in the future could return an error if `atRound > round` at line 623

tolikzinovyev: Yes, let's do it in a different pull request.

idb/postgres/postgres.go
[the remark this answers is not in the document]
winder: Could you explain? I think that `Read committed` is the default isolation level and we just need `Repeatable read`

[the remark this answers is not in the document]
winder: @brianolson have you considered the isolation level for ReadOnly queries? :point_up:

api/handlers.go
```diff
@@ -531,17 +487,17 @@ func notFound(ctx echo.Context, err string) error {
 ///////////////////////
 
 // fetchAssets fetches all results and converts them into generated.Asset objects
-func (si *ServerImplementation) fetchAssets(ctx context.Context, options idb.AssetsQuery) ([]generated.Asset, error) {
-	assetchan := si.db.Assets(ctx, options)
+func (si *ServerImplementation) fetchAssets(ctx context.Context, options idb.AssetsQuery) ([]generated.Asset, uint64, error) {
```
brianolson, line 490: it would be nicer to have `round uint64` named in the return values (here and a few places below)

idb/postgres/postgres.go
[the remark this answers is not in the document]
brianolson: For Transactions it probably doesn't matter. It's possible that TxnRow.Extra could get updated, but the core data is always read-only after the block is initially loaded. For everything else readonlySerializeable is probably a good idea.

idb/postgres/postgres.go
```diff
@@ -1628,7 +1657,7 @@ func (db *IndexerDb) txnsWithNext(ctx context.Context, tf idb.TransactionFilter,
 		close(out)
 		return
 	}
-	rows, err := db.db.QueryContext(ctx, query, whereArgs...)
+	rows, err := tx.Query(query, whereArgs...)
```
brianolson, line 1660: why lose the context?

tolikzinovyev: I passed the context when I created `tx`. Isn't it used when I query?

api/handlers.go
tolikzinovyev: I'm not a fan of named return values, but I added comments. Hope this is fine.

idb/postgres/postgres.go
brianolson: oh yep, that's right

accounting/rewind_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	var outCh <-chan idb.TxnRow
+
+	db := &mocks.IndexerDb{}
+	db.On("GetSpecialAccounts").Return(idb.SpecialAccounts{}, nil)
+	db.On("Transactions", mock.Anything, mock.Anything).Return(outCh, uint64(7)).Once()
+
+	account, err := AccountAtRound(account, 6, db)
+	assert.Error(t, err)
```
winder, line 71: Could you add a constant for the inconsistent data error and check for it here?

accounting/rewind_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/stretchr/testify/mock"
+
+	models "github.com/algorand/indexer/api/generated/v2"
+	"github.com/algorand/indexer/idb"
+	"github.com/algorand/indexer/idb/mocks"
+)
+
+func TestBasic(t *testing.T) {
+	var a types.Address
+	a[0] = 'a'
```
winder, line 18: I like this trick a lot.

winder (approved):
Looks good!

PR #399, opened 2021-03-23
Hotfix/2.2.2

winder:
**THIS PR IS FOR REVIEW ONLY**
**MERGE WITH: git flow hotfix finish 2.2.2**

## Summary

If there are many creatables, the account lookup becomes slow. Add missing indices.

## Test Plan

Run migration from 2.2.1 to 2.2.2 and see that the 4 new indices are created and account lookups become fast again.


onetechnical:
Blocking migrations finish in about 30s, reducing accounts endpoint from 1.2s to 0.17s in my test. Block import speeds seem consistent for 900 tx/block, at 1.3s.

PR #401, opened 2021-03-23
Use repeatable read transaction isolation level

### LINKED ISSUE #391: Cannot use serializable mode in a hot standby
## Summary
After the betanet indexer switched live on March 3rd, we started seeing "Serializable Read on hot standby" errors in the all-mobile-blue PostgreSQL logs (about every second):

```
2021-03-05 14:44:44 UTC:192.168.103.224(38268):indexer@betanet_green:[13388]:STATEMENT:  BEGIN ISOLATION LEVEL SERIALIZABLE READ ONLY
2021-03-05 14:44:44 UTC:192.168.103.224(38268):indexer@betanet_green:[13388]:HINT:  You can use REPEATABLE READ instead.
2021-03-05 14:44:44 UTC:192.168.103.224(38268):indexer@betanet_green:[13388]:ERROR:  cannot use serializable mode in a hot standby
```
John created a custom build of develop (cf93e3a) to build/deploy to betanet on March 3rd.
It may have to do with some of the indexer serializable code: https://github.com/algorand/indexer/search?q=serializable

[Cloudwatch Link
](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logsV2:log-groups/log-group/$252Faws$252Frds$252Fcluster$252Findexer-all-mobile-blue-db$252Fpostgresql/log-events$3Fstart$3D-3600000)
[Datadog Link](https://app.datadoghq.com/logs?agg_m=count&agg_t=count&from_ts=1614880354000&index=%2A&live=false&query=service%3Ards%20host%3A%2Amobile%2A%20%28ERROR%20OR%20FATAL%29%20-%22canceling%3Fstatement%3Fdue%3Fto%3Fuser%3Frequest%22%20-%22terminating%3Fconnection%3Fdue%3Fto%3Fadministrator%3Fcommand%22%20-%22invalid%3Finput%3Fsyntax%22&to_ts=1614881254000&cols=host%2Cservice&stream_sort=desc&messageDisplay=inline)
[Datadog Monitor](https://app.datadoghq.com/monitors/30897994?q=total&from_ts=1614875946000&to_ts=1614881406000&eval_ts=1614881406000)

## Scope/Requirements
* Troubleshoot and resolve the serializable errors
* Check that we no longer see similar betanet errors in the PostgreSQL logs (datadog or CloudWatch in the links above) or create a ticket with next steps.

## Urgency/Relative Priority
Medium - betanet seems to have an issue and is logging/alerting errors.

tolikzinovyev:
Use database connections instead of transactions to support multiple databases behind a loadbalancer (#351). Also fixes #391.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -43,6 +43,7 @@ const specialAccountsMetastateKey = "accounts"
 
 // Be a real ACID database
 var serializable = sql.TxOptions{Isolation: sql.LevelSerializable}
+var readonlySerializable = sql.TxOptions{Isolation: sql.LevelSerializable, ReadOnly: true}
 
 // OpenPostgres is available for creating test instances of postgres.IndexerDb
 func OpenPostgres(connection string, opts *idb.IndexerDbOptions, log *log.Logger) (pdb *IndexerDb, err error) {
@@ -420,14 +421,14 @@ func (db *IndexerDb) SetImportState(state idb.ImportState) (err error) {
 }
 
 // If `tx` is null, make a standalone query.
-func (db *IndexerDb) getMaxRoundAccounted(ctx context.Context, conn *sql.Conn) (round uint64, err error) {
+func (db *IndexerDb) getMaxRoundAccounted(tx *sql.Tx) (round uint64, err error) {
 	query := `select coalesce((v->>'account_round')::bigint, 0) from metastate where k = 'state'`
 
 	var row *sql.Row
-	if conn == nil {
-		row = db.db.QueryRowContext(ctx, query)
+	if tx == nil {
+		row = db.db.QueryRow(query)
 	} else {
-		row = conn.QueryRowContext(ctx, query)
+		row = tx.QueryRow(query)
 	}
 
 	err = row.Scan(&round)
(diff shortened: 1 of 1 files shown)


brianolson:
This might be fine but I'm not sure it solves the previous problem. A connection doesn't have higher default transaction serializeability guarantee, only an explicit transaction has that (in the Go database/sql API, apparently). So I think this leaves us open to the previously seen skew of different queries within an Indexer-API potentially having inconsistent results. But, maybe we're giving up because that's inevitable because Postgres read-only-replicas don't allow transactions anyway?

tolikzinovyev:
All we need is receiving consistent results for individual queries and receiving database state version in the increasing order. For example, if I now that round x is accounted, then I should be able to get transactions for round x. Doesn't a read-only replica process updates atomically and in order? I would expect so.

winder:
We have other types of queries with race conditions. For example:
1. Query the accounting round, receive round N
2. Accounting completes, updates the account and accounting round.
3. Query account data, receive data for round N+1.

I think you're right in that we could find a way to consistently use connections to fetch transactions, but then we would have to use the max round filter for all queries (or remove transactions at round N+1).

What is the motivation to using connections over transactions?

tolikzinovyev:
@winder you are right. I kind of took for granted that the code before we added transactions worked correctly with a single database. You provided a counter-example. I changed connections to repeatable read transactions, they provide "snapshot-like" read consistency, so that should be sufficient.

tolikzinovyev:
And I don't know if transactions impact performance here really. When you query blockchain transactions, the database needs to keep a copy of them, but they never change, so there is probably no cost.

winder (approved):
The repeatable read change looks good

PR #404, opened 2021-03-24
Better error handling with sql.Rows

tolikzinovyev:


idb/postgres/postgres.go
```diff
@@ -1758,6 +1766,18 @@ func (db *IndexerDb) yieldTxnsThreadSimple(ctx context.Context, rows *sql.Rows,
 			count++
 		}
 	}
+	if err := rows.Err(); err != nil {
+		select {
```
winder, line 1770: No need for a select here, just set errp if it hasn't already been set.

winder (changes requested):
Looks good, one minor comment

brianolson (approved):
good catch. this is definitely more proper error checking

tolikzinovyev:
Thanks!

PR #407, opened 2021-03-30
Make the defaultFrozen cache is immutable

winder:
## Summary

DefaultFrozen is immutable but the cache was being updated in a mutable way.

## Test Plan

Add a unit test that exploits the hole, it now works correctly.

### DIFF
--- a/util/test/account_testutil.go
+++ b/util/test/account_testutil.go
@@ -58,7 +58,7 @@ func DecodeAddressOrPanic(addr string) types.Address {
 }
 
 // MakeAssetConfigOrPanic is a helper to ensure test asset config are initialized.
-func MakeAssetConfigOrPanic(round, assetid, total, decimals uint64, defaultFrozen bool, unitName, assetName, url string, addr types.Address) (*types.SignedTxnWithAD, *idb.TxnRow) {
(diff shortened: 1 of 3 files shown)


util/test/account_testutil.go
```diff
@@ -58,7 +58,7 @@ func DecodeAddressOrPanic(addr string) types.Address {
 }
 
 // MakeAssetConfigOrPanic is a helper to ensure test asset config are initialized.
-func MakeAssetConfigOrPanic(round, assetid, total, decimals uint64, defaultFrozen bool, unitName, assetName, url string, addr types.Address) (*types.SignedTxnWithAD, *idb.TxnRow) {
+func MakeAssetConfigOrPanic(round, configid, assetid, total, decimals uint64, defaultFrozen bool, unitName, assetName, url string, addr types.Address) (*types.SignedTxnWithAD, *idb.TxnRow) {
```
tolikzinovyev, line 61: I think this function is only used for creating an asset, so we don't need to make it more general. Maybe renaming is a good idea.

winder: It's technically correct, but maybe it can be improved. The transaction type is `AssetConfig`, so it would have to be read as `Make` `AssetConfig` `OrPanic` rather than `MakeAsset` `ConfigOrPanic`

tolikzinovyev: I would just name it `MakeCreateAssetTxn()` as in "make transaction that creates an asset". What do you think?

PR #409, opened 2021-03-31
Remove creator's asset holding when total supply is 0 on asset destruction

### LINKED ISSUE #412: Fix accounting logic
When an asset is destroyed, creator's asset holding is not deleted if the total supply is 0. This is incorrect.

tolikzinovyev:
## Summary
When an asset is destroyed, creator's asset holding was not deleted if the total supply is 0.

Section 9.2 of the ledger specification says that the total supply is allowed to be 0.

Closes #412.

## Test Plan
Added tests that exploit the bug.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -1077,7 +1077,7 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
 		// Note! leaves `asset` and `account_asset` rows present for historical reference, but deletes all holdings from all accounts
 		any = true
 		// Update any account_asset holdings which were not previously closed. By now the amount should already be 0.
-		ads, err := tx.Prepare(`UPDATE account_asset SET amount = 0, closed_at = $1, deleted = true WHERE assetid = $2 AND amount != 0`)
+		ads, err := tx.Prepare(`UPDATE account_asset SET amount = 0, closed_at = $1, deleted = true WHERE addr = (SELECT creator_addr FROM asset WHERE index = $2) AND assetid = $2`)
 		if err != nil {
 			return fmt.Errorf("prepare asset destroy, %v", err)
 		}
--- a/idb/postgres/postgres_integration_test.go
+++ b/idb/postgres/postgres_integration_test.go
@@ -705,6 +705,38 @@ func TestZeroTotalAssetCreate(t *testing.T) {
 	assertAccountAsset(t, db, test.AccountA, assetid, false, 0)
 }
 
+func assertAssetDates(t *testing.T, db *sql.DB, assetID uint64, deleted sql.NullBool, createdAt sql.NullInt64, closedAt sql.NullInt64) {
+	row := db.QueryRow(
+		"SELECT deleted, created_at, closed_at FROM asset WHERE index = $1", int64(assetID))
+
+	var retDeleted sql.NullBool
+	var retCreatedAt sql.NullInt64
+	var retClosedAt sql.NullInt64
+	err := row.Scan(&retDeleted, &retCreatedAt, &retClosedAt)
+	assert.NoError(t, err)
+
+	assert.Equal(t, deleted, retDeleted)
+	assert.Equal(t, createdAt, retCreatedAt)
+	assert.Equal(t, closedAt, retClosedAt)
+}
+
+func assertAssetHoldingDates(t *testing.T, db *sql.DB, address types.Address, assetID uint64, deleted sql.NullBool, createdAt sql.NullInt64, closedAt sql.NullInt64) {
+	row := db.QueryRow(
+		"SELECT deleted, created_at, closed_at FROM account_asset WHERE " +
+		"addr = $1 AND assetid = $2",
+		address[:], assetID)
+
+	var retDeleted sql.NullBool
+	var retCreatedAt sql.NullInt64
+	var retClosedAt sql.NullInt64
+	err := row.Scan(&retDeleted, &retCreatedAt, &retClosedAt)
+	assert.NoError(t, err)
+
+	assert.Equal(t, deleted, retDeleted)
+	assert.Equal(t, createdAt, retCreatedAt)
+	assert.Equal(t, closedAt, retClosedAt)
+}
+
 func TestDestroyAssetBasic(t *testing.T) {
 	db, connStr, shutdownFunc := setupPostgres(t)
 	defer shutdownFunc()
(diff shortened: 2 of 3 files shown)


idb/postgres/postgres_integration_test.go
```diff
@@ -668,4 +669,247 @@ func TestIgnoreDefaultFrozenConfigUpdate(t *testing.T) {
...
+		state := getAccounting(test.Round + 1, cache)
+		err := state.AddTransaction(txnRow)
+		assert.NoError(t, err)
+
+		err = pdb.CommitRoundAccounting(state.RoundUpdates, test.Round + 1, &itypes.Block{})
+		assert.NoError(t, err, "failed to commit")
+	}
+
+	// Check that the asset is deleted.
+	{
```
winder, line 711: You could extend/use `assertAccountAsset` or make a new similar function for these checks.

idb/postgres/postgres_integration_test.go
```diff
@@ -668,4 +669,247 @@ func TestIgnoreDefaultFrozenConfigUpdate(t *testing.T) {
...
+					Type: "acfg",
+					Header: types.Header{
+						Sender:     test.AccountA,
+					},
+					AssetConfigTxnFields: types.AssetConfigTxnFields{
+						AssetParams: types.AssetParams{
+							Manager:       test.AccountB,
+							Reserve:       test.AccountB,
+							Freeze:        test.AccountB,
+							Clawback:      test.AccountB,
```
winder, line 841: These are all AccountB so the test isn't working quite as designed according to the assertions.

idb/postgres/postgres.go
```diff
@@ -1077,7 +1077,7 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
 		// Note! leaves `asset` and `account_asset` rows present for historical reference, but deletes all holdings from all accounts
 		any = true
 		// Update any account_asset holdings which were not previously closed. By now the amount should already be 0.
-		ads, err := tx.Prepare(`UPDATE account_asset SET amount = 0, closed_at = $1, deleted = true WHERE assetid = $2 AND amount != 0`)
+		ads, err := tx.Prepare(`UPDATE account_asset SET amount = 0, closed_at = $1, deleted = true WHERE addr = (SELECT creator_addr FROM asset WHERE index = $2) AND assetid = $2`)
```
winder, line 1080: Elsewhere we use a different pattern for subselects, which I find a lot more readable than the subqueries:
```
WITH lookup_creator_addr AS (SELECT creator_addr FROM asset WHERE index = $2)
UPDATE account_asset SET amount = 0, closed_at = $1, deleted = true WHERE addr = lookup_creator_addr AND assetid = $2
```

On the other hand, it looks like the Postgres optimizer doesn't work well with  "WITH" queries. There is some discussion about it in the doc: [WITH Queries](https://www.postgresql.org/docs/13/queries-with.html).

@brianolson do you know about the performance implications of WITH queries? What do you think about this specific case and using a subquery vs CTE?

brianolson: this changed statement now _only_ affects the creator account, I don't think that's right. Maybe it wants to be simpler and more broad? Maybe just drop the `AND amount != 0` ?

winder: @brianolson I was hoping we could do that, but unfortunately algod supports orphaned asset holdings :(

brianolson: Now I remember that the asset-close operation was supposed to be forbidden by algod unless the creator has reclaimed all the units

idb/postgres/postgres_integration_test.go
tolikzinovyev: Why? We test that creator's asset holding is deleted and manager's is not.

idb/postgres/postgres_integration_test.go
tolikzinovyev: done

idb/postgres/postgres.go
tolikzinovyev: @winder this doesn't work for some reason: `column "lookup_creator_addr" does not exist`

winder: @brianolson it's true, the creator must have a balance equal to the asset total. But it's possible other accounts opted into the asset without receiving any, or to send (instead of close-to) their holdings, or to have their holdings collected via clawback. Those holdings still exist after the asset has been destroyed.

brianolson: maybe the logic should be `WHERE assetid = $2 AND deleted <> false`

winder: From your first comment this morning - we're only supposed to update the creator. Here is a test case that shows why:
```
# Create asset 1
~/a/sandbox$ ./sandbox goal asset create --creator VQSYKAUHMLFZIV3CQJOEHCQIORCEMNL4GX76EVDDE7XNEYIW5NE7II64Q4 --name t=0 --unitname amisc --total 0 --defaultfrozen
Issued transaction from account VQSYKAUHMLFZIV3CQJOEHCQIORCEMNL4GX76EVDDE7XNEYIW5NE7II64Q4, txid VVVNDD7YGDJUYGE2FVDUUHGCEL3M6RM4BXKJSPUQCGE2OWYIY4NQ (fee 1000)
Transaction VVVNDD7YGDJUYGE2FVDUUHGCEL3M6RM4BXKJSPUQCGE2OWYIY4NQ still pending as of round 14
Transaction VVVNDD7YGDJUYGE2FVDUUHGCEL3M6RM4BXKJSPUQCGE2OWYIY4NQ committed in round 16
Created asset with asset index 1

# Optin account ZRVH
 ~/a/sandbox$ ./sandbox goal asset send -a 0 --assetid 1 -f ZRVH7UH2H7JQNGO3BD3VBBH25N64CW3JZMUTTV667E5INJ2DOWE2ECMQNE -t ZRVH7UH2H7JQNGO3BD3VBBH25N64CW3JZMUTTV667E5INJ2DOWE2ECMQNE
Issued transaction from account ZRVH7UH2H7JQNGO3BD3VBBH25N64CW3JZMUTTV667E5INJ2DOWE2ECMQNE, txid X34NTL3ET5F2GTQFG52LFTR44M4ITJH4OH7CNUCMS4GX5E5EUUCA (fee 1000)
Transaction X34NTL3ET5F2GTQFG52LFTR44M4ITJH4OH7CNUCMS4GX5E5EUUCA still pending as of round 21
Transaction X34NTL3ET5F2GTQFG52LFTR44M4ITJH4OH7CNUCMS4GX5E5EUUCA committed in round 23

# Destroy asset 1
 ~/a/sandbox$ ./sandbox goal asset destroy --creator VQSYKAUHMLFZIV3CQJOEHCQIORCEMNL4GX76EVDDE7XNEYIW5NE7II64Q4 --assetid 1
Issued transaction from account VQSYKAUHMLFZIV3CQJOEHCQIORCEMNL4GX76EVDDE7XNEYIW5NE7II64Q4, txid 64CBUU7B6XIQICQ2M77IUQLCBQZDN64NI7RACI2343Z5T54EYTXQ (fee 1000)
Transaction 64CBUU7B6XIQICQ2M77IUQLCBQZDN64NI7RACI2343Z5T54EYTXQ still pending as of round 26
Transaction 64CBUU7B6XIQICQ2M77IUQLCBQZDN64NI7RACI2343Z5T54EYTXQ committed in round 28

# Account ZRVH still has the asset holding even though the asset is deleted
 ~/a/sandbox$ curl "localhost:4001/v2/accounts/ZRVH7UH2H7JQNGO3BD3VBBH25N64CW3JZMUTTV667E5INJ2DOWE2ECMQNE?pretty" -H "X-Algo-API-Token: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
{
  "address": "ZRVH7UH2H7JQNGO3BD3VBBH25N64CW3JZMUTTV667E5INJ2DOWE2ECMQNE",
  "amount": 3999999999999000,
  "amount-without-pending-rewards": 3999999999999000,
  "apps-local-state": [],
  "apps-total-schema": {
    "num-byte-slice": 0,
    "num-uint": 0
  },
  "assets": [
    {
      "amount": 0,
      "asset-id": 1,
      "creator": "",
      "is-frozen": true
    }
  ],
  "created-apps": [],
  "created-assets": [],
  "pending-rewards": 0,
  "reward-base": 0,
  "rewards": 0,
  "round": 32,
  "status": "Offline"
}
```

idb/postgres/postgres_integration_test.go
winder: Ok that makes sense. I think I missed that AccountC opts-in below and you check for that as well.

winder (approved):

PR #414, opened 2021-04-01
Migration to fix default-frozen = true accounts.

winder:
## Summary

Fix asset holdings for assets with default-frozen = true.

## Test Plan

There is a new unit test, aside from that there are a number of assets/holdings on mainnet/testnet/betanet with default-frozen = true which our validator tool had previously identified.


idb/postgres/postgres_migrations.go
```diff
@@ -46,6 +47,8 @@ func init() {
 		{m7StaleClosedAccounts, false, "clear some stale data from closed accounts"},
 		{m8TxnJSONEncoding, false, "some txn JSON encodings need app keys base64 encoded"},
 		{m9SpecialAccountCleanup, false, "The initial m6 implementation would miss special accounts."},
+
```
brianolson, line 50: is there an m10 hiding in another PR?

brianolson (approved):
LGTM

idb/postgres/postgres_migrations.go
winder: yeah exactly,  I'm hoping to make the merge easier

PR #417, opened 2021-04-02
Update version to 2.3.2

tolikzinovyev:


winder (changes requested):
This happens during the release process

PR #420, opened 2021-04-06
Warn the user during m7 if it might use too much memory.

tolikzinovyev:
## Summary
Adds a warning and asks to pass `--force-m7` flag to override it.

## Test Plan
Replaced the if statement to warn unconditionally and ran `make integrate` with and without `--force-m7` to verify the behavior.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -26,6 +26,7 @@ var (
 	noAlgod          bool
 	developerMode    bool
 	tokenString      string
+	forceM7          bool
 )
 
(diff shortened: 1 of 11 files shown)


winder (commented):
Some concerns:
1) m7 is specific to Postgres migrations, so this breaks the `IndexerDb` abstraction.
2) having the option visible in `-h` make indexer slightly harder to use, especially in the future since the option becomes less and less relevant to existing deployments.
3) this warning seems better suited for release notes. What user do we have in mind who will see this message and find it useful?

If we really need a warning like this, perhaps we should use an environment variable. That way it can be checked directly from the Postgres migration.

PR #421, opened 2021-04-06
Do not touch asset closeto address if close amount = 0

tolikzinovyev:
## Summary
This is what algod does: https://github.com/algorand/go-algorand/blob/master/ledger/apply/asset.go#L225.

## Test Plan
Will run validation.


winder (approved):
Looks good. I was worried that this was going to be more complicated.

PR #423, opened 2021-04-06
Warn the user during m7 if it might use too much memory.

tolikzinovyev:
## Summary
Adds a warning and asks to set `FORCEM7` env variable to override it.

## Test Plan
Replaced the if statement to warn unconditionally and ran `make integrate` with and without `FORCEM7` to verify the behavior.

### DIFF
--- a/idb/postgres/postgres_migrations.go
+++ b/idb/postgres/postgres_migrations.go
@@ -821,6 +821,28 @@ func m7RewardsAndDatesPart2UpdateAccounts(db *IndexerDb, accountData []addressAc
 	return nil
 }
 
+func warnUser(db *IndexerDb, maxRound uint32) error {
+	query := "SELECT COUNT(*) FROM account WHERE (created_at IS NULL) OR (created_at <= $1)"
+	row := db.db.QueryRow(query, maxRound)
+
+	var count uint64
+	err := row.Scan(&count)
+	if err != nil {
+		return fmt.Errorf("m7: unable to query the number of rows: %v", err)
+	}
+
+	if count > 10000000/3*4 {
+		db.log.Print("the migration m7 is likely to use more than 4GB of RAM")
+
+		envVar := "FORCEM7"
+		if _, ok := os.LookupEnv(envVar); !ok {
+			return fmt.Errorf("m7: set %s environment variable to force the migration", envVar)
+		}
+	}
+
+	return nil
+}
+
 func getAccountsFirstUsed(db *IndexerDb, maxRound uint32, specialAccounts idb.SpecialAccounts) (map[sdk_types.Address]txnID, error) {
 	res := make(map[sdk_types.Address]txnID)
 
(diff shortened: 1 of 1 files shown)


idb/postgres/postgres_migrations.go
```diff
@@ -821,6 +821,28 @@ func m7RewardsAndDatesPart2UpdateAccounts(db *IndexerDb, accountData []addressAc
...
+	row := db.db.QueryRow(query, maxRound)
+
+	var count uint64
+	err := row.Scan(&count)
+	if err != nil {
+		return fmt.Errorf("m7: unable to query the number of rows: %v", err)
+	}
+
+	if count > 10000000/3*4 {
+		db.log.Print("the migration m7 is likely to use more than 4GB of RAM")
```
algorotem, line 835: We should probably make this comment more explanatory - "The current migration (m7) is likely to use more than 4GB of RAM. The avoid overuse, to process stopped automatically"

idb/postgres/postgres_migrations.go
```diff
@@ -821,6 +821,28 @@ func m7RewardsAndDatesPart2UpdateAccounts(db *IndexerDb, accountData []addressAc
...
+
+		envVar := "FORCEM7"
+		if _, ok := os.LookupEnv(envVar); !ok {
+			return fmt.Errorf("m7: set %s environment variable to force the migration", envVar)
```
algorotem, line 839: It comes right after so - "To force start the migration, please set the environment variable %s to TRUE". I know that the work doesn't matter but it's better to be explicit with the user.

algorotem (changes requested):

idb/postgres/postgres_migrations.go
```diff
@@ -821,6 +821,28 @@ func m7RewardsAndDatesPart2UpdateAccounts(db *IndexerDb, accountData []addressAc
...
+	query := "SELECT COUNT(*) FROM account WHERE (created_at IS NULL) OR (created_at <= $1)"
+	row := db.db.QueryRow(query, maxRound)
+
+	var count uint64
+	err := row.Scan(&count)
+	if err != nil {
+		return fmt.Errorf("m7: unable to query the number of rows: %v", err)
+	}
+
+	if count > 10000000/3*4 {
```
algorotem, line 834: This is a magic number. Consider putting it in a constant.

algorotem (changes requested):

idb/postgres/postgres_migrations.go
tolikzinovyev: Where should I put it?

winder: I think defining it above the condition is fine along with a comment is fine, no need to spread this logic around:
```
// This many accounts needs about 4GB of memory
maxAccountsWithoutWarning := 10000000/3*4
```

algorotem (approved):

PR #426, opened 2021-04-08
Fix repeatable read tx options.

winder:
## Summary

We forgot to change the isolation level.


tolikzinovyev:
Constants is not my thing.

tolikzinovyev (approved):

PR #427, opened 2021-04-08
migration: clear account data

tolikzinovyev:
## Summary
A new migration that clears account data for accounts that have been closed.

## Test Plan
Will run validation.


idb/postgres/postgres_migrations.go
```diff
@@ -1936,3 +1938,104 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+		res = append(res, m12Account{addr, uint64(closedAt.Int64)})
+	}
+	if err := rows.Err(); err != nil {
+		return []m12Account{}, err
+	}
+
+	return res, nil
+}
+
+func m12FixAuthAddr(db *sql.DB, account m12Account) error {
```
winder, line 1978: Not sure it's worth it, but I think we can do this for all accounts with a query something like

```
UPDATE account SET account_data - 'spend' WHERE addr = $1 AND 0 < (SELECT count(*) FROM txn JOIN txn_participation ON round=round AND intra=intra WHERE round > $2 AND (txn -> 'txn' ->> 'rekey') IS NOT NULL AND addr = $1 LIMIT 1)
```

idb/postgres/postgres_migrations.go
```diff
@@ -1936,3 +1938,104 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+	// This will not work properly if the account was closed and reopened in the same round
+	// but that's unlikely to happen.
+	query := "SELECT decode(txn -> 'txn' ->> 'rekey', 'base64') FROM txn WHERE " +
+		"round > $1 AND (txn -> 'txn' ->> 'rekey') IS NOT NULL ORDER BY round DESC LIMIT 1"
```
winder, line 1982: This probably needs to join in txn_participation, and include a filter by sender = account?

idb/postgres/postgres_migrations.go
```diff
@@ -1936,3 +1938,104 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+		query = "UPDATE account SET account_data = account_data - 'spend' WHERE addr = $1"
+		_, err := db.Exec(query, account.address[:])
+		return err
+	}
+
+	return err
+}
+
+func m12FixKeyreg(db *sql.DB, account m12Account) error {
+	query := "SELECT txnbytes FROM txn WHERE round > $1 AND typeenum = $2 ORDER BY round DESC LIMIT 1"
```
winder, line 1999: Filter by account?

idb/postgres/postgres_migrations.go
```diff
@@ -1936,3 +1938,104 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+	row := db.QueryRow(query, account.closedAt)
+
+	var txnbytes []byte
+	err := row.Scan(&txnbytes)
+
+	if err == sql.ErrNoRows {
+		// Delete keyreg fields.
+		query = "UPDATE account SET account_data = account_data - 'vote' - 'sel' - 'onl' - " +
+			"'voteFst' - 'voteLst' - 'voteKD'"
```
winder, line 2008: Where `addr = $1`

winder (changes requested):
This doesn't look quite right

tolikzinovyev:
Thanks, I fixed the queries. Could you check them again?

idb/postgres/postgres_migrations.go
```diff
@@ -1936,3 +1938,104 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+		_, err := db.Exec(query, account.address[:])
+		return err
+	}
+
+	return err
+}
+
+func m12ClearAccountData(db *IndexerDb, state *MigrationState) error {
+	db.accountingLock.Lock()
+	defer db.accountingLock.Unlock()
```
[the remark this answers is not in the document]
tolikzinovyev: I will use a retry here are #431 is merged.

idb/postgres/postgres_migrations.go
```diff
@@ -1935,3 +1938,108 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+}
+
+func m12FixAuthAddr(db *sql.DB, account m12Account) error {
+	// This will not work properly if the account was closed and reopened in the same round
+	// but that's unlikely to happen.
+	query := "SELECT decode(txn -> 'txn' ->> 'rekey', 'base64') FROM txn WHERE " +
+		"(round, intra) IN (SELECT round, intra FROM txn_participation WHERE addr = $1 " +
+		"AND round > $2 ORDER BY round DESC) AND" +
+		"(txn -> 'txn' ->> 'snd') = $1 AND (txn -> 'txn' ->> 'rekey') IS NOT NULL " +
+		"ORDER BY round DESC LIMIT 1"
```
winder, line 1985: I'm not sure if this query is correct, it looks a little different than how we join in txn_participation elsewhere (i.e. it's missing `JOIN ... ON`, not sure how that differs from `IN`)

I'm now realizing that we can use the transactions helper instead of worrying about that:
```
		rows, _ = db.Transactions(
                    context.Background(),
                    idb.TransactionFilter {
                        Address: account.address[:],
                        RekeyTo: boolPtr(true),
                        MinRound: account.closedAt,
                        Limit: 1, // depending on the results order we might not be able to limit, but I think it is descending when you search account txns
		})
```

idb/postgres/postgres_migrations.go
```diff
@@ -1935,3 +1938,108 @@ func m11AssetHoldingFrozen(db *IndexerDb, state *MigrationState) error {
...
+		_, err := db.Exec(query, account.address[:])
+		return err
+	}
+
+	return err
+}
+
+func m12FixKeyreg(db *sql.DB, account m12Account) error {
+	query := "SELECT txnbytes FROM txn WHERE round > $2 AND typeenum = $3 AND " +
+		"(txn -> 'txn' ->> 'snd') = $1 ORDER BY round DESC LIMIT 1"
```
winder, line 2003: This query still isn't right. Lets try to use the `Transactions` function.

winder (changes requested):
Some more comments for you.

Another thing missing is a test. The current `make integration` test is pretty basic, but it's all we have at the moment. Basically, you look for a problematic account in `test/migrations/create_delete.2.2.1.txt` and add a test in `test/postgres_migration_test.sh`.

If there are no problematic accounts I've been manually modifying `create_delete.2.2.1.txt`

idb/postgres/postgres_migrations.go
tolikzinovyev: I think we need to query transactions and do the update below in one database transaction if we don't use the mutex. Should I create `db.transactions(ctx, tx, tf)`?

tolikzinovyev:
`postgres_migrations.go` is getting bloated. I would like to move the migration to a different file and not sure yet if it should be done in this PR.

rotemh:
> `postgres_migrations.go` is getting bloated. I would like to move the migration to a different file and not sure yet if it should be done in this PR.

I would consider doing it in another one. While low risk, the refactoring may introduce new bugs and it would be probably easier to deal with in a separate one.

idb/postgres/postgres.go
```diff
@@ -1428,7 +1428,10 @@ func (db *IndexerDb) GetBlock(ctx context.Context, round uint64, options idb.Get
 			return types.Block{}, nil, err
 		}
 
-		go db.yieldTxnsThreadSimple(ctx, rows, out, true, nil, nil)
+		go func() {
+			db.yieldTxnsThreadSimple(ctx, rows, out, nil, nil)
+			close(out)
```
winder, line 1433: I'm not sure about this pattern, seems like an accident waiting to happen when someone calls the function but forgets to call close afterwards.

idb/postgres/postgres_migrations_integration_test.go
```diff
@@ -0,0 +1,294 @@
...
+	sdk_types "github.com/algorand/go-algorand-sdk/types"
+
+	"github.com/algorand/indexer/accounting"
+	"github.com/algorand/indexer/idb"
+	"github.com/algorand/indexer/importer"
+	"github.com/algorand/indexer/types"
+	"github.com/algorand/indexer/util/test"
+)
+
+func setupDb(t *testing.T) (*IndexerDb /*db*/, func() /*shutdownFunc*/) {
```
winder, line 22: Could you reuse `setupPostgres` instead? Having a single entrypoint for this would make #471 easier

idb/postgres/postgres_migrations_integration_test.go
```diff
@@ -0,0 +1,294 @@
...
+	db, err := OpenPostgres(connStr, idb.IndexerDbOptions{}, nil)
+	assert.NoError(t, err)
+
+	err = db.SetProto(test.Proto, types.ConsensusParams{})
+	assert.NoError(t, err)
+
+	return db, shutdownFunc
+}
+
+func importTxns(t *testing.T, db *IndexerDb, round uint64, txns ...*sdk_types.SignedTxnWithAD) {
```
winder, line 50: Nice trick!

idb/postgres/postgres_migrations.go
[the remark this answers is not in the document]
winder: Could you move the lock into the `fixAuthAddr` / `fixKeyreg` functions?

I think it's still required to ensure no race conditions where a transaction is committed while we're processing transactions.

winder (commented):
This looks really nice, just a few small comments and questions before approving.

idb/postgres/postgres_migrations_integration_test.go
tolikzinovyev: I was going to move it and other functions to `postgres_integration_test_common.go` and replace `setupPostgres` with it. Should I do it in this PR?

idb/postgres/postgres.go
tolikzinovyev: I think the worst thing that can happen is, you will get stuck. Tests would catch it.

tolikzinovyev:
Thanks. I actually removed the use of accounting lock. Accounting code and this code both use the retry functionality. Do you think the lock is still needed here?

idb/postgres/postgres_migrations_integration_test.go
winder: No, that sounds good.

winder:
@tolikzinovyev I'm pretty sure the lock is still required to ensure there are no race conditions. Here is one example that I don't think is resolved with retrying:

1) We query transactions for account A with no rekeying.
2) A round is committed where account A rekeys to B
3) We finish processing transactions and remove the rekey from step 2.

tolikzinovyev:
@winder, but we do 1 and 3 in the same transaction.

winder:
@tolikzinovyev exactly, 1 & 3 are in the same postgres tx with RepeatableRead. That means that if more transactions are added to the DB the initial query wouldn't fetch them and we could wipe out valid changes thinking they were stale.

tolikzinovyev:
I don't think that's the case. All transactions here use the serializable isolation level which means their execution is equivalent to some sequential execution.

winder:
@tolikzinovyev I did some more reading about the serializable isolation level and it looks like you are right, the guarantees are a lot more robust than I had realized. The documentation mentions a lot of caveats and warnings that I'm still processing.

Considering the complexity of Serializable, I think it's still a good idea to use it as a safety net and continue managing serialization ourselves with the mutex.

@brianolson curious what you think about this.

tolikzinovyev:
I think the point of Serializable is to let you forget about all complexity. On the other hand, preventing conflicts on the client side is always more efficient and robust.

brianolson:
I think Serializeable is a good idea. I think the accounting lock so that we don't try to do things that are likely to cause serializeable-conflicts is also good.

PR #431, opened 2021-04-12
Retry db transactions on a serialization error

### LINKED ISSUE #435: Serialization error with postgres
## Summary
Indexer crashes due to a serialization error during a normal catchup.

## Urgency/Relative Priority
Medium

tolikzinovyev:
Closes #435.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -17,6 +17,7 @@ import (
 	"math/big"
 	"os"
 	"strings"
+	"sync"
 	"time"
 	"unicode/utf8"
 
@@ -113,6 +114,8 @@ type IndexerDb struct {
 	protoCache map[string]types.ConsensusParams
 
 	migration *migration.Migration
+
+	accountingLock sync.Mutex
 }
 
 func (db *IndexerDb) init() (err error) {
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -836,6 +836,9 @@ func setDirtyAppLocalState(dirty []inmemAppLocalState, x inmemAppLocalState) []i
 }
 
 func (db *IndexerDb) commitRoundAccounting(updates idb.RoundUpdates, round uint64, blockPtr *types.Block) (err error) {
+	db.accountingLock.Lock()
+	defer db.accountingLock.Unlock()
+
 	any := false
 	tx, err := db.db.BeginTx(context.Background(), &serializable)
 	if err != nil {
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"math/rand"
 	"os"
 	"strings"
 	"sync"
@@ -27,7 +28,7 @@ import (
 	atypes "github.com/algorand/go-algorand-sdk/types"
 
 	// Load the postgres sql.DB implementation
-	_ "github.com/lib/pq"
+	"github.com/lib/pq"
 	log "github.com/sirupsen/logrus"
 
 	models "github.com/algorand/indexer/api/generated/v2"
@@ -83,6 +84,30 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
 		err = pdb.init()
 	}
 	return
+}
+
+func txnWithRetry(f func() error) error {
+	maxDelay := time.Minute
+	delay := time.Millisecond
+
+	for {
+		err := f()
+
+		// If not serialization error.
+		var pqerr *pq.Error
+		if !errors.As(err, &pqerr) || (pqerr.Code != "40001") {
+			return err
+		}
+
+		fmt.Println("retrying transaction")
+		// Randomize delay to avoid conflicts.
+		time.Sleep(time.Duration(rand.Int63n(int64(delay))))
+
+		delay *= 2
+		if delay > maxDelay {
+			maxDelay = delay
+		}
+	}
 }
 
 // IndexerDb is an idb.IndexerDB implementation
@@ -188,8 +213,7 @@ func (db *IndexerDb) AddTransaction(round uint64, intra int, txtypeenum int, ass
 	return nil
 }
 
-// CommitBlock is part of idb.IndexerDB
-func (db *IndexerDb) CommitBlock(round uint64, timestamp int64, rewardslevel uint64, headerbytes []byte) error {
+func (db *IndexerDb) commitBlock(round uint64, timestamp int64, rewardslevel uint64, headerbytes []byte) error {
 	tx, err := db.db.BeginTx(context.Background(), &serializable)
 	if err != nil {
 		return fmt.Errorf("BeginTx %v", err)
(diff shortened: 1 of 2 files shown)


idb/postgres/postgres.go
```diff
@@ -85,6 +86,30 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
...
+			return err
+		}
+
+		fmt.Println("retrying transaction")
+		// Randomize delay to avoid conflicts.
+		time.Sleep(time.Duration(rand.Int63n(int64(delay))))
+
+		delay *= 2
+		if delay > maxDelay {
+			maxDelay = delay
```
winder, line 108: return serialize error when delay > maxDelay?

idb/postgres/postgres.go
```diff
@@ -85,6 +86,30 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
...
 
+func txnWithRetry(f func() error) error {
+	maxDelay := time.Minute
+	delay := time.Millisecond
+
+	for {
+		err := f()
+
+		// If not serialization error.
+		var pqerr *pq.Error
```
winder, line 97: Is there any alternative to exposing pq here? As discussed we probably should replace this library and this makes it a little more difficult.

idb/postgres/postgres.go
```diff
@@ -268,12 +292,19 @@ func (db *IndexerDb) CommitBlock(round uint64, timestamp int64, rewardslevel uin
...
+
+// CommitBlock is part of idb.IndexerDB
+func (db *IndexerDb) CommitBlock(round uint64, timestamp int64, rewardslevel uint64, headerbytes []byte) error {
+	f := func() error {
+		return db.commitBlock(round, timestamp, rewardslevel, headerbytes)
+	}
+	err := txnWithRetry(f)
+
 	db.txrows = nil
 	db.txprows = nil
```
winder, line 306: Why not put these into `commitBlock` so that you can return txnWithRetry like the other wrapper?

idb/postgres/postgres.go
```diff
@@ -85,6 +86,30 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
...
+
+		// If not serialization error.
+		var pqerr *pq.Error
+		if !errors.As(err, &pqerr) || (pqerr.Code != "40001") {
+			return err
+		}
+
+		fmt.Println("retrying transaction")
+		// Randomize delay to avoid conflicts.
+		time.Sleep(time.Duration(rand.Int63n(int64(delay))))
```
winder, line 104: Not sure we need a delay, do both transactions fail? I would think the other had to succeed in order for this one to fail.

winder (commented):
Looks nice.

idb/postgres/postgres.go
```diff
@@ -85,6 +86,30 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
...
+	for {
+		err := f()
+
+		// If not serialization error.
+		var pqerr *pq.Error
+		if !errors.As(err, &pqerr) || (pqerr.Code != "40001") {
+			return err
+		}
+
+		fmt.Println("retrying transaction")
```
winder, line 102: Use the log

idb/postgres/postgres.go
winder: and why delay *= 2?

tolikzinovyev: You probably right, we don't need a delay.

idb/postgres/postgres.go
tolikzinovyev: I don't want to modify state in a function that might be repeated.

idb/postgres/postgres.go
tolikzinovyev: this was for debugging

idb/postgres/postgres.go
tolikzinovyev: I think the only alternative is too compare the error message. Not pretty.

winder:
I think you were right to leave mutex to begin with. Lets leave the old migrations alone, the next time we add a blocking migration we can safely remove the mutex.

tolikzinovyev:
Ok :)

tolikzinovyev:
@winder LGTY?

idb/postgres/postgres.go
```diff
@@ -85,6 +85,21 @@ func openPostgres(db *sql.DB, opts *idb.IndexerDbOptions, logger *log.Logger) (p
 	return
 }
 
+// A helper function that retries the function `f` in case the database transaction in it
+// fails due to a serialization error. `f` must return an error which contains the error returned
+// by sql.Tx.Commit(). The easiest way is to just return the result of sql.Tx.Commit().
+func txnWithRetry(f func() error) error {
+	for {
+		err := f()
+
```
winder, line 94: We should write a log message indicating that we had to retry, and maybe  return the number of attempts it took to commit.

winder (changes requested):
Have you reproduced a problem where this happens and verify that it is fixed?

With earlier builds, I was able to reproduce the serialization error with multiple indexers in sandbox. You can start a 2nd instance like this:
```
diff --git a/docker-compose.yml b/docker-compose.yml
index cbf6a83..94a8246 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,6 +43,29 @@ services:
       - indexer-db
       - algod
 
+  indexer2:
+    container_name: "algorand-sandbox-indexer2"
+    build:
+      context: .
+      dockerfile: ./images/indexer/Dockerfile
+      args:
+        URL: "${INDEXER_URL}"
+        BRANCH: "${INDEXER_BRANCH}"
+        SHA: "${INDEXER_SHA}"
+    ports:
+      - 8981:8981
+    restart: unless-stopped
+    environment:
+      DISABLED: "${INDEXER_DISABLED}"
+      PORT: "8981"
+      SNAPSHOT: ""
+      CONNECTION_STRING: "host=indexer-db port=5432 user=algorand password=algorand dbname=indexer_db sslmode=disable"
+      ALGOD_ADDR: "algod:4001"
+      ALGOD_TOKEN: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+    depends_on:
+      - indexer-db
+      - algod
+
   indexer-db:
     image: "postgres:13-alpine"
     container_name: "algorand-sandbox-postgres"
diff --git a/sandbox b/sandbox
index d9c04fd..d2f8c21 100755
--- a/sandbox
+++ b/sandbox
@@ -292,6 +292,11 @@ sandbox () {
         dc exec indexer /bin/bash
         return
         ;;
+      indexer2)
+        statusline "Entering /bin/bash session in the indexer2 container..."
+        dc exec indexer2 /bin/bash
+        return
+        ;;
       indexer-db)
         statusline "Entering /bin/bash session in the indexer-db container..."
         dc exec indexer-db /bin/bash
```

tolikzinovyev:
@winder my two DBs have fully caught up but still no success in reproducing the behavior. I think we should just merge this.

brianolson:
I'm surprised that 'serialization error' is an ephemeral thing that goes away if we try again. Isn't that the error we got from things that PostgreSQL's version of JSON couldn't understand?
Otherwise this looks like fine retry logic that's fair to add.

PR #433, opened 2021-04-12
fetcher: protect `err` with a mutex

tolikzinovyev:


winder (approved):
Probably unlikely to happen, but definitely correct to have. Thanks.

PR #454, opened 2021-04-20
Hotfix/2.3.3

winder:
**FOR REVIEW ONLY DO NOT MERGE**

## Summary

This is a fix for the testnet failure to fetch 13527623, 13527623 and 13527806

## Test Plan

Verified that the 3 rounds are now returned.

### DIFF
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -420,10 +420,13 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 		d := make([]generated.AccountStateDelta, 0)
 		for _, k := range keys {
 			v := stxn.ApplyData.EvalDelta.LocalDeltas[k.key]
-			d = append(d, generated.AccountStateDelta{
-				Address: k.address.String(),
-				Delta:   *(stateDeltaToStateDelta(v)),
-			})
+			delta := stateDeltaToStateDelta(v)
+			if delta != nil {
+				d = append(d, generated.AccountStateDelta{
+					Address: k.address.String(),
+					Delta:   *delta,
+				})
+			}
 		}
 		localStateDelta = &d
 	}


api/converter_utils.go
```diff
@@ -420,10 +420,13 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 		d := make([]generated.AccountStateDelta, 0)
 		for _, k := range keys {
 			v := stxn.ApplyData.EvalDelta.LocalDeltas[k.key]
```
winder, line 422: For this round there were no deltas for the sender. Because this is added implicitly we were crashing when it gets dereferenced.

algorandskiy:
The fix is good, the situation when the sender gets empty delta - I'll double check if it happens only on key removal.

onetechnical:
Should this target 'master', not 'develop'?

winder:
@algojohnlee yes. I created the branch with `git flow hotfix start 2.3.3`. PRs probably default to develop, `git flow hotfix finish` should do the right thing.

winder:
@algorandskiy it might just be indexer making incorrect assumptions about how to use this data. A few lines above the change I implicitly add the sender with an implicit 0 key. Is that correct? The other keys are regular values.

algorandskiy (approved):
this empty delta exist in both original 2.4.1 and new version.
so the fix is necessary but it needs to be consistent with block delta:
```
        "dt": {
          "ld": {
            "0": {},
            "1": {
              "Status": {
                "at": 1,
                "bs": "Unconfirmed"
              },
              "VoteState": {
                "at": 1,
                "bs": "Add"
              },
              "Votes": {
                "at": 2,
                "ui": 1
              }
            }
          }
        },
```
It is up to FDE but we might want to preserve empty sender delta as delta in the block.

PR #455, opened 2021-04-21
Reset command

brianolson:
## Summary

Productionize resetting the database. A 'reset' deletes all accounting and derived data while retaining original block_header and txn data (faster than re-loading block by block from the archival algod). When Indexer is again started as a daemon it would re-calculate all accounting from round 0 up to the latest block.

Alternative implementation vs #338 . This version drops tables and waits for them to be re-created at db .init() time.

## Test Plan


winder (dismissed):
LGTM

cmd/algorand-indexer/main.go
```diff
@@ -78,13 +78,18 @@ var (
 	cpuProfile     string
 	pidFilePath    string
 	db             idb.IndexerDb
```
tolikzinovyev, line 80: Why do `db` and `dbOpts` need to be global variables?

cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to configure logger: %v", err)
+			os.Exit(1)
+		}
+		opts := idb.IndexerDbOptions{NoMigrate: true}
+		db := globalIndexerDb(&opts)
+		timeout := time.Now().Add(5 * time.Second)
+		health, err := db.Health()
+		maybeFail(err, "could not get db health, %v", err)
+		for health.IsMigrating || !health.DBAvailable {
```
tolikzinovyev, line 37: I think we should just exit.

idb/postgres/reset_sql.go
```diff
@@ -0,0 +1,12 @@
+// Code generated from source reset.sql via go generate. DO NOT EDIT.
+
+package postgres
+
+const reset_sql = `DROP TABLE IF EXISTS account;
```
tolikzinovyev, line 5: I think there is a discrepancy here. We don't drop `txn_participation` table, but we do drop `metastate`. However, one of Will's recent migrations fixes `txn_participation`. So, if I were to run the reset function, `txn_participation` will never be fixed. I think the proper solution is to generate `txn_participation` during accounting phase instead of block import. @winder any thoughts?

tolikzinovyev:
This version is much simpler!

cmd/algorand-indexer/main.go
brianolson: yeah, `globalIndexerDb()` didn't turn out to be used in the pattern I first imagined, it is now `indexerDbFromFlags()` and fits better.

idb/postgres/reset_sql.go
brianolson: Aside from that bug, `txn_participation` isn't accounting, it's written at block load time and then we're done with it. `reset` is about re-doing everything that gets lumped into the 'accounting' pass.

winder: Stuff like this is why we should never document this command. In my opinion this is a development tool.

tolikzinovyev: @winder if you are fine with it, I'm also fine

cmd/algorand-indexer/reset.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	andRebuild bool
+)
+
+var resetCmd = &cobra.Command{
+	Use:   "reset",
+	Short: "run indexer reset",
+	Long:  "run indexer reset. Serve api on HTTP.",
```
tolikzinovyev, line 23: could you maybe add that this is a developer's tool only?

tolikzinovyev (commented):
Looks good overall. However, is it really necessary to sleep after opening a database? I would really prefer to remove that logic.

cmd/algorand-indexer/reset.go
brianolson: This was partly copy-paste-o from the daemon command. Long text is now explanatory and cautionary.

winder (approved):
LGTM

PR #464, opened 2021-04-26
Fix conflict between initial import refactoring and reset command.

winder:
## Summary

The PRs we merged today had a conflict.


egieseke (approved):
Looks good.

Issue #461, opened 2021-04-26
Support databases behind loadbalancers (2)


tolikzinovyev:
## Summary
We need indexer code to work correctly when used with a replicated database, particularly Amazon Aurora.

## Scope/Requirements
- ~Fix most API endpoints~ #373
- Fix account rewinding (accounting/rewind.go)
- ~Fix accounting for importing blocks (importer/helper.go: updateAccounting()); should be partially fixed by #331, `YieldTxns()` can still return inconsistent results~ will be fixed by https://github.com/algorand/indexer/pull/618
- Fix transaction paging (idb/postgres.go: Transactions()); it is possible that the user gets only some transactions for a round because of the limit, but the next Transactions() call will not return the rest of the transactions for this round

## Urgency/Relative Priority
?

Blackglade:
+1 on this. Having issues with this using Aurora.

winder:
@Blackglade what sort of issues are you seeing?

Blackglade:
@winder Deleted my previous comment. My issues with Aurora were not related to the indexer. Aurora has been working great so far :)

winder:
@Blackglade great! Thanks for the feedback.

tsachizehub:
@tolikzinovyev - could you please break this one into two, and we will size these next week ?

tolikzinovyev:
Closing in favor of #619 and #620.

PR #465, opened 2021-04-26
Fix paging inconsistencies when using amazon aurora

tolikzinovyev:
## Summary
It is possible that the user gets only some transactions for a round because of the limit, but the next `Transactions()` call will not return the rest of the transactions for this round because a follower DB replica returned stale data.

Fixes one item in #461.

## Test Plan
Will add a test.

### DIFF
--- a/accounting/rewind.go
+++ b/accounting/rewind.go
@@ -12,15 +12,6 @@ import (
 	"github.com/algorand/indexer/types"
 )
 
-// ConsistencyError is returned when the database returns inconsistent (stale) results.
-type ConsistencyError struct {
-	msg string
-}
-
-func (e ConsistencyError) Error() string {
-	return e.msg
-}
-
 func assetUpdate(account *models.Account, assetid uint64, add, sub uint64) {
 	if account.Assets == nil {
 		account.Assets = new([]models.AssetHolding)
@@ -101,7 +92,7 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 	}
 	txns, r := db.Transactions(context.Background(), tf)
 	if r < account.Round {
-		err = ConsistencyError{fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round)}
+		err = types.MakeConsistencyError(fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round))
 		return
 	}
 	txcount := 0
@@ -179,8 +170,8 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 		tf.Limit = 1
 		txns, r = db.Transactions(context.Background(), tf)
 		if r < round {
-			err = ConsistencyError{
-				fmt.Sprintf("queried round r: %d < requested round round: %d", r, round)}
+			err = types.MakeConsistencyError(
+				fmt.Sprintf("queried round r: %d < requested round round: %d", r, round))
 			return
 		}
 		for txnrow := range txns {
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -446,11 +446,15 @@ func (db *IndexerDb) GetMaxRoundAccounted() (round uint64, err error) {
 	return db.getMaxRoundAccounted(nil)
 }
 
-// GetMaxRoundLoaded is part of idb.IndexerDB
-func (db *IndexerDb) GetMaxRoundLoaded() (round uint64, err error) {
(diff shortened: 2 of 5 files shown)


brianolson:
I think the indexer _should_ be able to return txns that haven't been accounted yet. Part of the two phase import (load data / do accounting) is that just having the data loaded makes it available and useful. Someone can query their txn history and doesn't need Indexer to do the accounting for them to do that query.

tolikzinovyev:
Thanks, I meant to use the max loaded round.

idb/postgres/postgres.go
```diff
@@ -1675,15 +1685,26 @@ func (db *IndexerDb) txnsWithNext(ctx context.Context, tx *sql.Tx, tf idb.Transa
...
 		out <- idb.TxnRow{Error: err}
-		close(out)
+		return
+	}
+	maxRoundLoaded, err := db.getMaxRoundLoaded(tx)
+	if err != nil {
+		out <- idb.TxnRow{Error: err}
+		return
+	}
+	if maxRoundLoaded < nextround {
```
winder, line 1695: I think the condition needs to change slightly to check if an address filter is provided. Those results are returned oldest to newest so we wouldn't have to worry about the order.

brianolson:
So, the ultimate problem is like this?
1. do an ascending query (e.g. all txns by asset)
2. get to approximately _now_ and get a next-token
3. do a query against that next token, but this time served by a different db backend that hasn't progressed as far, so we don't return data when we might otherwise be able to?

The problem is that Indexer might invisibly query two different database backends A and B with different views of the most recent txns? But the problem is only with txns in the last 0-10 seconds or so? **I think we shouldn't try to fix this problem**. I think we shouldn't add complexity to Indexer to try and workaround this strange setup. Just let the query return no data because it isn't there yet. If someone really-really wants the latest transactions they should be following with Algobot.

winder:
I agree with Brian on this one, if absolute consistency on results paging becomes a requirement we really need to use proper DB level pagination.

tolikzinovyev:
@brianolson, yes that's the issue. The problem is, everyone uses this setup and I think correctness is important. Right now we might return some transactions for a round and not be able to return the rest. @algorotem, @winder and I discussed this and this was the solution we agreed on.

idb/postgres/postgres.go
tolikzinovyev: I think this problem applies to both situations.

accounting/rewind.go
```diff
@@ -101,7 +92,7 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 	}
 	txns, r := db.Transactions(context.Background(), tf)
 	if r < account.Round {
-		err = ConsistencyError{fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round)}
+		err = types.MakeConsistencyError(fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round))
```
tsachiherman, line 95: given that you've already did the hard work of having this error object, why don't you want to take advantage of it and place the two rounds inside that object and let the object's `Error()` method to format the string ?

tolikzinovyev: The line you pointed is actually that "elsewhere". It was there before. Now I use this error for transaction paging.

idb/postgres/postgres.go
tsachiherman: not sure if I'm following - does both queries are using the same transaction ? If so, it shouldn't be inconsistent.

tolikzinovyev: I'm also not sure what you mean. :)  Will is talking about two modes of reading transactions. If address filter is not given, transactions are returned from oldest to newest. With address filter, transactions are returned from newest to oldest.

winder (changes requested):
I think we should close this.

The proposed consistency check does seems slightly better than the current one, but I think there might be other inconsistencies. If we want to improve paging in general and add new guarantees, I think that is a bigger feature that we need to design.

Also, I think this particular inconsistency is also obsolete after the accounting refactoring.

mxmauro:
The sooner or later, users should expect some inconsistencies in order to be able to scale. Probably getting all the transactions of a block can be handled "atomically" but also expecting, for e.g., accounts to be up-to-date at the same time would be unrealistic.

Creating a big db transaction to keep consistency every time a new block is processed may become a bottleneck mainly if the number of transactions raises in the futre and a lot of indexes are added to tables in order to make queries faster.

tolikzinovyev:
Closing in favor of #683.

PR #467, opened 2021-04-27
Pass IndexerDbOptions by value

tolikzinovyev:
## Summary
I think there is no reason to have a pointer.

## Test Plan
The change is trivial.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -70,7 +70,7 @@ var daemonCmd = &cobra.Command{
 		if noAlgod && !allowMigration {
 			opts.ReadOnly = true
 		}
-		db := indexerDbFromFlags(&opts)
+		db := indexerDbFromFlags(opts)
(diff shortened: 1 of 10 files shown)


brianolson (changes requested):
Why do you like by-value better? Isn't `nil` a better signal of 'nothing set'?
By value copies the whole contents of the struct every time, pointer doesn't.

tolikzinovyev:
Mainly for the sake of readability. Then you don't need wonder if `nil` is somehow different from `&IndexerOptions{}`. Also the user of `opts` doesn't need to check for `nil` every time a field needs to be read. It's true that pointer might be more efficient, but it's just initialization code, so doesn't matter.

winder (approved):
The change seems fine to me. I guess this would be a more functional approach, it's nice that it avoids some weird nil checks during initialization.

tolikzinovyev:
@winder I rebased it, please approve

winder (approved):

PR #470, opened 2021-04-28
Update consensus params to support v27

tsachiherman:
## Summary

Update consensus params to support protocol v27

## Test Plan

Standard testing.


winder (approved):
LGTM

Issue #443, opened 2021-04-14
Block Import Test Driver


winder:
Create a tool to drive the performance tests.

It needs to do the following:
1) Select a config file (i.e. from the scenarios or user specified)
2) Create a postgres database.
3) Launch the generator with the config file.
4) Launch a specific version of indexer (defined by repo / branch / optional commit-hash)
5) Run for a specified duration.
6) Record performance metrics.
7) Cleanup.
8) Continue from step 1.

It should allow the test to run for a specific duration, this doesn't fit nicely in the config file so maybe it is an option to this tool.

PR #474, opened 2021-04-28
Block Generator Prototype

winder:
## Summary

Working prototype for a block generator performance tool. It generates pseudorandom transactions and fulls up blocks. By imitating algod's `/genesis` and `/v2/blocks/<>` endpoints it interoperates with indexer.

## Test Plan

For now there is a simple driver script which starts postgres in a docker container, starts generator, and connects indexer to them.


tolikzinovyev:
Would it make more sense to have it in a different repository and have the code depend on types in go-algorand? Since this intends to mimic algod, we shouldn't depend on indexer but on go-algorand.

winder:
@tolikzinovyev These blocks aren't useful to anything except Indexer because they don't have real signatures and there is no block certificate, so I don't think it makes sense to put in go-algorand. If all of the types were available in the go SDK then maybe it could be a standalone project. At the moment I think this repo is the best spot for it.

cmd/block-generator/main.go
```diff
@@ -0,0 +1,117 @@
...
+		var f *os.File
+		f, err = os.Open(configFile)
+		if err != nil {
+			return err
+		}
+
+		return viper.ReadConfig(f)
+	}
+
+	return viper.ReadInConfig()
```
tolikzinovyev, line 45: could you add the default config name in the help text / flag description?

cmd/block-generator/main.go
```diff
@@ -0,0 +1,117 @@
...
+		TxnPerBlock:                  15000,
+		NewAccountFrequency:          100,
+		Protocol:                     "future",
+		NumGenesisAccounts:           10,
+		GenesisAccountInitialBalance: 1000000000,
+		GenesisID:                    "blockgen-test",
+		GenesisHash:                  [32]byte{},
+	})
+
+	portStr := fmt.Sprintf(":%d", port)
```
tolikzinovyev, line 64: could you move this down?

cmd/block-generator/main.go
```diff
@@ -0,0 +1,117 @@
...
+	fmt.Fprintf(w, "Use /v2/blocks/:blocknum: to get a block.")
+}
+
+func handleGenesis(w http.ResponseWriter, r *http.Request) {
+	gen.WriteGenesis(w)
+}
+
+func handleBlock(w http.ResponseWriter, r *http.Request) {
+	// The generator doesn't actually care about the block...
+	//block, err := parseBlock(r.URL.Path)
```
tolikzinovyev, line 84: please delete

cmd/block-generator/main.go
```diff
@@ -0,0 +1,117 @@
...
+	//	return
+	//}
+
+	gen.WriteBlock(w)
+}
+
+const blockQueryPrefix = "/v2/blocks/"
+const blockQueryBlockIdx = len(blockQueryPrefix)
+
+func parseBlock(path string) (uint64, error) {
```
tolikzinovyev, line 96: I think this is not used. So are the two constants above.

cmd/block-generator/run_generator.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Demonstrate how to run the generator and connect it to indexer.
+
+POSTGRES_CONTAINER=generator-test-container
+POSTGRES_PORT=15432
+POSTGRES_DATABASE=generator_db
+
+function start_postgres() {
+   # Start postgres container...
```
tolikzinovyev, line 10: indentation is funky

cmd/block-generator/run_generator.sh
```diff
@@ -0,0 +1,46 @@
...
+   docker run \
+     -d \
+     --name $POSTGRES_CONTAINER \
+     -e POSTGRES_USER=algorand \
+     -e POSTGRES_PASSWORD=algorand \
+     -e PGPASSWORD=algorand \
+     -p $POSTGRES_PORT:5432 \
```
tolikzinovyev, line 17: could you make a constant for 5432? what is it for btw?

cmd/block-generator/run_generator.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "Starting postgres container."
+start_postgres
+echo "Starting block generator (see generator.log)"
+./cmd/block-generator/block-generator -port 11111 -config cmd/block-generator/config.yml > generator.log 2>&1 &
+GENERATOR_PID=$!
+./cmd/algorand-indexer/algorand-indexer daemon \
+              -S localhost:8980 \
+              --algod-net localhost:11111 \
+              --algod-token security-is-our-number-one-priority \
+              -P "host=localhost user=algorand password=algorand dbname=generator_db port=15432 sslmode=disable"
```
tolikzinovyev, line 46: We have constants for dbname and port.

cmd/block-generator/run_generator.sh
```diff
@@ -0,0 +1,46 @@
...
+
+echo "Building generator."
+go build
+pushd ../.. > /dev/null
+echo "Building indexer."
+make
+echo "Starting postgres container."
+start_postgres
+echo "Starting block generator (see generator.log)"
+./cmd/block-generator/block-generator -port 11111 -config cmd/block-generator/config.yml > generator.log 2>&1 &
```
tolikzinovyev, line 40: could you make a constant for 11111?

tolikzinovyev:
Looks good overall! One small concern I have is that we don't account for the fee. If the initial amount is too small, we could make some balance negative in indexer. Also, could you rebase?

cmd/block-generator/run_generator.sh
winder: This script is just an MVP so I don't think we need to worry about cleaning it up, a more advanced test driver is going to be needed: https://github.com/algorand/indexer/issues/443

cmd/block-generator/main.go
winder: I passed it into the interface in case it becomes useful later.

cmd/block-generator/main.go
winder: done

cmd/block-generator/generator/generate.go
```diff
@@ -0,0 +1,275 @@
...
+
+	stxn := sdk_types.SignedTxn{
+		Sig:      sdk_types.Signature{},
+		Msig:     sdk_types.MultisigSig{},
+		Lsig:     sdk_types.LogicSig{},
+		Txn:      txn,
+		AuthAddr: sdk_types.Address{},
+	}
+	stxn.Sig[32] = 50
+	/*
```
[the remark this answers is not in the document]
winder: I'd like to leave it for now since the signature handling should probably be further considered.

cmd/block-generator/generator/generate.go
```diff
@@ -218,13 +210,14 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (types.Signed
 	receiver := indexToAccount(receiveIndex)
 
 	amount := g.balances[sendIndex] / 2
-	g.balances[sendIndex] -= amount
-	g.balances[receiveIndex] += amount
 
-	if g.balances[sendIndex] < 0 {
-		panic(fmt.Sprintf("the balance fell below zero for idx %d", sendIndex))
+	if g.balances[sendIndex] < amount {
```
tolikzinovyev, line 214: This obviously cannot happen. Why are you adding this?

cmd/block-generator/generator/generate.go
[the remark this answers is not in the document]
tolikzinovyev: Indexer doesn't look at signatures. What's the use case? And you can always add it later.

cmd/block-generator/main.go
```diff
@@ -0,0 +1,114 @@
...
+
+	result := uint64(0)
+	pathlen := len(path)
+
+	if pathlen == blockQueryBlockIdx {
+		return 0, fmt.Errorf("no block in path")
+	}
+
+	for i := blockQueryBlockIdx; i < pathlen; i++ {
+		if path[i] < '0' || path[i] > '9' {
```
tolikzinovyev, line 106: This can be moved outside of the loop. Also, isn't there some function in go that converts strings to integers?

cmd/block-generator/generator/generate.go
winder: I changed the comparison based on your feedback, it is checking for < amount to detect an underflow.

cmd/block-generator/main.go
winder: What part can move outside of the loop?

cmd/block-generator/generator/generate.go
```diff
@@ -0,0 +1,275 @@
...
+// GenerationConfig defines the tunable parameters for block generation.
+type GenerationConfig struct {
+	TxnPerBlock         uint64
+	NewAccountFrequency uint64
+
+	Protocol                     string
+	NumGenesisAccounts           uint64
+	GenesisAccountInitialBalance uint64
+	GenesisID                    string
+	GenesisHash                  [32]byte
```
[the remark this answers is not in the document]
tolikzinovyev: could you change this too?

cmd/block-generator/generator/generate.go
tolikzinovyev: a < a/2 is always false

cmd/block-generator/main.go
tolikzinovyev: The line I commented on - 106. It only does anything if `i = blockQueryBlockIdx`.

winder: oh, it's also the exit condition in case there are query parameters like `?pretty`

tolikzinovyev: Well, it only exits when `i = blockQueryBlockIdx`. Then it's a bug.

tolikzinovyev: can you write some tests?

cmd/block-generator/generator/generate.go
winder: What I'm trying to do with this check is make sure the program exits if the account balance is too small to make a useful transaction. I see now that the logic isn't correct. I guess I could `g.balances[sendIndex]` > 1, which would also avoid the possible divide-by-zero crash.

This code is going to have to change anyway since it doesn't currently support closing an account. So I don't think this needs to be fixed.

tolikzinovyev: Ok, just remove the check.

tolikzinovyev (approved):

PR #487, opened 2021-05-04
Delete file import state from db schema.

tolikzinovyev:


winder (approved):
LGTM. The only other thing I would consider adding is to mark the `import` sub-command as hidden to further discourage people from using this feature.

PR #488, opened 2021-05-04
Delete consensus params from DB schema.

tolikzinovyev:


winder (approved):
So the downside here is that we load a (currently) 65K JSON file into memory. Seems like a small cost to simplifying the code. If we get to a point where this is important we could find a way to solve it inside `types.Protocol`

Maybe storing it in program memory with the new `embed` package.

PR #492, opened 2021-05-05
Refactor database code.

tolikzinovyev:
## Summary
I split `idb/dummy.go` into several files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ idb/postgres/setup_postgres_sql.go idb/postgres/reset_sql.go:	idb/postgres/setup
 types/protocols_json.go:	types/protocols.json types/consensus.go
 	cd types && go generate
 
-idb/mocks/IndexerDb.go:	idb/dummy/dummy.go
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ idb/postgres/setup_postgres_sql.go idb/postgres/reset_sql.go:	idb/postgres/setup
 types/protocols_json.go:	types/protocols.json types/consensus.go
 	cd types && go generate
 
-idb/mocks/IndexerDb.go:	idb/dummy.go
+idb/mocks/IndexerDb.go:	idb/dummy/dummy.go
```
winder, line 24: This should probably be `idb/idb.go`

winder (changes requested):
Looks pretty good but I think the mock generation should look at the interface, not the dummy implementation

Makefile
tolikzinovyev: Thanks!

winder (approved):

PR #494, opened 2021-05-05
Delete fake migration.

tolikzinovyev:


winder:
I was kinda proud when I came up with this hack, it's saved us a couple of times when we forgot to update the next migration index (though I suppose it will have to change when we start deprecating them).

tolikzinovyev:
I will make sure each migration has a test that checks the index is incremented. :)

winder:
I think it would be better to replace this with some sort of auto-increment method

tolikzinovyev:
Would you like to contribute?

PR #497, opened 2021-05-07
Delete unused import block and round limits.

tolikzinovyev:


idb/idb.go
```diff
@@ -498,9 +498,6 @@ type UpdateFilter struct {
 	// StartRound only include transactions confirmed at this round or later.
 	StartRound uint64
 
-	// RoundLimit only process this many rounds of transactions.
-	RoundLimit *int
```
winder, line 502: I think we should keep this. I'd like to use it along with MaxRound to make the startup process more deterministic. On start we would run accounting with `MaxRound = MaxTxLoaded`, then the block handler would run with `RoundLimit = 1`.

tolikzinovyev: I thought the plan was to do import and accounting at the same time instead of separately.

brianolson: There's still a lot of value in doing it in two phases. 'most recent txns for my account' is a query that doesn't need accounting, so having that data always be immediately available is worthwhile. There could even be a future mode where an Indexer is configurable to not do accounting at all (no forward accounting, no rewind) and _just_ be for searching txns. (There has been some worry that Indexer+Postgres won't be able to do accounting at sufficient speed to keep up with 10kTPS algod or more, so not doing accounting could sidestep that.)

tolikzinovyev: Why is performance a worry? It shouldn't be slower than algod if we read accounts and write accounts in 1 RTT each.

tolikzinovyev: Also having "no accounting mode" doesn't prevent us from doing import and accounting in one phase.

tolikzinovyev: Actually, I think one phase will be better for performance, because then we don't have to read transactions from DB.

winder: We'll have some perf tests soon to help guide these sorts of design changes.

PR #500, opened 2021-05-10
make check

tolikzinovyev:
## Summary

Add a make target that checks that all go code (except tests) can be compiled and use this rule in the CI script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,10 @@
 # Build artifacts
 cmd/algorand-indexer/algorand-indexer
+cmd/block-generator/block-generator
+cmd/e2equeries/e2equeries
+cmd/idbtest/idbtest
+cmd/texttosource/texttosource
+cmd/validator/validator
 api/.3tmp.json
 tmp/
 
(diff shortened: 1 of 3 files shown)


brianolson:
If we want control there is always `go build -o`. Could be:
```make
bin/algorand-indexer: idb/postgres/setup_postgres_sql.go idb/postgres/reset_sql.go types/protocols_json.go
        cd cmd/algorand-indexer && CGO_ENABLED=0 go build -o ../../bin/algorand-indexer -ldflags="${GOLDFLAGS}"
```

tsachiherman:
Yes, I'm ok with going with this for now, but I have a feeling that we'll hit various "blockers" down the road which would force us to make changes here as well.

winder:
@tsachiherman you may be right, but the release process doesn't use the Makefile, so that's true regardless of any changes we make here. I've heard that the devops team wants to rewrite the mule build/deploy scripts. So hopefully they'll solve this problem for us in a way thats consistent with go-algorand.

winder (approved):

PR #509, opened 2021-05-21
Block Generator Reporting

winder:
## Summary

Add a basic reporting endpoint for verifying the expected transaction distribution, and seeing the performance impact from generating transactions.

Technically this should have a mutex but I don't want to add that to the generation path. That can be addressed later if it ever becomes a real problem.

## Test Plan

Run with scenario configs, and check `/report`


cmd/block-generator/generator/generate.go
```diff
@@ -18,19 +19,24 @@ import (
 
 var errOutOfRange = fmt.Errorf("selection is out of weighted range")
 
+type txID string
```
tolikzinovyev, line 22: I think `txType` would be more appropriate. "Transation ID" often refers to the transaction hash.

tolikzinovyev (approved):

cmd/block-generator/generator/generate.go
```diff
@@ -607,7 +608,7 @@ func (g *generator) generateAssetTxnInternalHint(txType txID, sp sdk_types.Sugge
 
 	txn.Fee = 1000
 	if g.balances[senderIndex] < uint64(txn.Fee) {
-		fmt.Printf(fmt.Sprintf("\n\nthe sender account does not have enough algos for the transfer. idx %d, payment number %d\n\n", senderIndex, g.numPayments))
+		fmt.Printf(fmt.Sprintf("\n\nthe sender account does not have enough algos for the transfer. idx %d, asset type %v, num %d\n\n", senderIndex, actual, g.reportData[actual].GenerationCount))
```
tolikzinovyev, line 611: asset type -> transaction type?

winder: asset transaction type might be a little more accurate, but not transaction type. It's because asset transactions have multiple configurations to do different things (i.e. an asset xfer is used to optin/closeout/transfer).

winder: Fixed

tolikzinovyev (approved):

PR #518, opened 2021-06-01
Prometheus Metrics Endpoint

winder:
## Summary

The development branch of `echo-contrib` has changes to their Prometheus metrics endpoint making it suitable for our needs.

## Test Plan

A docker-compose environment has been created to demonstrate how everything should be configured. You can use the `/metrics` endpoint without all of that stuff, but histogram metrics are difficult to read without additional processing.

I put together a simple dashboard for testing:
image

Note: the block import time looks bad because it's bad, not because the metrics are wrong. It's possible this is because of the block-generator, more testing is needed.


tolikzinovyev:
One note and some questions.

1. The last chart title in the screenshot has a typo: "Byes".
2. What is "request size bytes" in the screenshot? The length of the url?
3. What is "all queries" in the screenshot?
4. Could you add units of measurements to each chart?
5. I'm concerned about storing too much data in memory. What does Prometheus store exactly? Does it delete old info?

winder:
> One note and some questions.
> 
> 1. The last chart title in the screenshot has a typo: "Byes".

Fixed
> 2. What is "request size bytes" in the screenshot? The length of the url?

The HTTP request sizes in bytes.
> 3. What is "all queries" in the screenshot?

95th percentile of the total sum of requests per 5 minutes. It's an out of the box expression for histograms in grafana:
`histogram_quantile(0.95, sum(rate(indexer_request_duration_seconds_bucket[5m])) by (le))`
> 4. Could you add units of measurements to each chart?

This is just for testing, not a real dashboard.
> 5. I'm concerned about storing too much data in memory. What does Prometheus store exactly? Does it delete old info?

Indexer only maintains real-time stats and counters, there is no old data for Indexer to worry about. Prometheus is a time series database that is configured to poll the new `/metrics` endpoint at some interval. Every 15 seconds is the default interval.

The only unusual thing I've done is to include query parameters in the path, this is to allow seeing the impact of certain query parameters (i.e. timestamp searches, rollback, etc). This means that instead of maintaining a single histogram per endpoint it's possible to have a histogram for every combination of parameters.

There is some overhead though, so I just added a `--disable-metrics` option in case people want to disable metrics.

cmd/algorand-indexer/daemon.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
```
brianolson, line 10: a little confusing that this is different than what's used in api/server.go

cmd/algorand-indexer/daemon.go
```diff
@@ -25,17 +27,26 @@ var (
 	noAlgod          bool
 	developerMode    bool
 	allowMigration   bool
+	metricsMode      string
 	tokenString      string
 )
 
+var importTimeHistogramSeconds = prometheus.NewSummary(
```
brianolson, line 34: Comment on what this does? It registers something into a package level global and then disappears.

api/server.go
```diff
@@ -16,24 +17,49 @@ import (
...
+}
+
 // Serve starts an http server for the indexer API. This call blocks.
-func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError error, log *log.Logger, tokens []string, developerMode bool) {
+func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError error, log *log.Logger, options ExtraOptions) {
 	e := echo.New()
 	e.HideBanner = true
 
+	if options.MetricsEndpoint {
+		p := prometheus.NewPrometheus("indexer", nil, nil)
```
brianolson, line 41: I'm guessing from context that labstack/echo-contrib/prometheus is providing the handler for `/metrics` and it will provide per-URL-path counts, like `/metrics: 123`, but it would be nice to see a comment describing what's going on. Also I don't see where any handler is specifically bound to `/metrics`. Where does that happen?

winder: That all happens in `p.Use(e)`, which is weird since middleware is normally registered with `e.Use(p)`. I guess the reason for this is because `p.Use` calls `e.Use` for stat collection and also registers the `/metrics` handler.

I added some comments to the code.

cmd/algorand-indexer/daemon.go
winder: done

cmd/algorand-indexer/daemon.go
winder: I changed it so that the echo module uses a different package name. This way only the prometheus client code will use that package.

brianolson:
LGTM

brianolson (approved):

PR #520, opened 2021-06-03
Make `deleted` columns NOT NULL

tolikzinovyev:
## Test Plan

Added a migration test.

### DIFF
--- a/idb/postgres/postgres_migrations.go
+++ b/idb/postgres/postgres_migrations.go
@@ -52,6 +52,7 @@ func init() {
 		// Migrations for a next release
 		{FixFreezeLookupMigration, false, "Fix search by asset freeze address."},
 		{ClearAccountDataMigration, false, "clear account data for accounts that have been closed"},
+		{MakeDeletedNotNullMigration, false, "make all \"deleted\" columns NOT NULL"},
 	}
 
 	// Verify ensure the constant is pointing to the right index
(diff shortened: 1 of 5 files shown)


idb/postgres/postgres_migrations.go
```diff
@@ -2204,3 +2205,32 @@ func ClearAccountDataMigration(db *IndexerDb, state *MigrationState) error {
...
+		"UPDATE account_app SET deleted = false WHERE deleted is NULL",
+		"ALTER TABLE account_app ALTER COLUMN deleted SET NOT NULL",
+	}
+
+	for _, query := range queries {
+		_, err := db.db.Exec(query)
+		if err != nil {
+			return fmt.Errorf("query: \"%s\" err: %w", query, err)
+		}
+	}
```
brianolson, line 2233: use `sqlMigration()`

idb/postgres/postgres_migrations.go
```diff
@@ -2204,3 +2205,32 @@ func ClearAccountDataMigration(db *IndexerDb, state *MigrationState) error {
...
+
+// MakeDeletedNotNullMigration makes "deleted" columns NOT NULL in tables
+// account, account_asset, asset, app, account_app.
+func MakeDeletedNotNullMigration(db *IndexerDb, state *MigrationState) error {
+	db.accountingLock.Lock()
+	defer db.accountingLock.Unlock()
+
+	queries := []string{
+		"UPDATE account SET deleted = false WHERE deleted is NULL",
+		"ALTER TABLE account ALTER COLUMN deleted SET NOT NULL",
```
brianolson, line 2217: I think this isn't quite safe. There could be a moment between `UPDATE` and `ALTER TABLE` where a new row is added but it doesn't have a value set so is NULL which violates NOT NULL.
But, we can
```sql
ALTER TABLE account ALTER COLUMN deleted SET DEFAULT false
```
Then new rows will get the default value written. _Then_ `UPDATE`, then `ALTER ... SET NOT NULL`

idb/postgres/postgres_migrations.go
brianolson: hmm, but that wraps all the changes in one big transaction block, and it might be better not to, _if_ we make sure it's a fully safe sequence (see other comment)

idb/postgres/postgres_migrations.go
tolikzinovyev: Do you mean somebody can connect to DB and manually insert a row? We do use the accounting mutex, so the accounting code will not create conflicts.

brianolson: I guess that makes it safe.

winder: In addition to that, it would require (a) multiple writers, and (b) one of the writers would have to be an old version.

I think our bases are covered here even if we didn't have the mutex.

idb/postgres/postgres_migrations.go
```diff
@@ -2204,3 +2205,21 @@ func ClearAccountDataMigration(db *IndexerDb, state *MigrationState) error {
 
 	return nil
 }
+
+// MakeDeletedNotNullMigration makes "deleted" columns NOT NULL in tables
+// account, account_asset, asset, app, account_app.
+func MakeDeletedNotNullMigration(db *IndexerDb, state *MigrationState) error {
+	queries := []string{
+		"UPDATE account SET deleted = false WHERE deleted is NULL",
```
winder, line 2213: Have you seen that there are cases with `deleted = NULL` after the m7 migration? If there were null values that could indicate corruption, maybe we should allow the migration to fail?

winder, line 2213: Have you seen that there are cases with deleted = NULL after the m7 migration? If there were null values that could indicate corruption, maybe we should allow the migration to fail?

winder (commented):
One small question about the UPDATEs. Looks good otherwise.

idb/postgres/postgres_migrations.go
tolikzinovyev: No, m7 should make all `deleted` either `true` or `false`. But just in case I'm wrong about it, let's do `UPDATE`s.

idb/postgres/postgres_migrations.go
tolikzinovyev: why does it matter?

winder (approved):

PR #522, opened 2021-06-03
Retry transactions in sqlMigration().

tolikzinovyev:


idb/postgres/postgres_migrations.go
```diff
@@ -1165,32 +1165,28 @@ func sqlMigration(db *IndexerDb, state *MigrationState, sqlLines []string) error
...
 	thisMigration := state.NextMigration
-	tx, err := db.db.BeginTx(context.Background(), &serializable)
-	if err != nil {
-		db.log.WithError(err).Errorf("migration %d tx err", thisMigration)
-		return err
-	}
-	defer tx.Rollback() // ignored if .Commit() first
-	for i, cmd := range sqlLines {
-		_, err = tx.Exec(cmd)
+	state.NextMigration++
```
winder, line 1168: Because `state` is passed by reference, please wait to update it until after the migration is successful

winder (changes requested):
Just a small request.

Looking at this code, I'm also wondering about the `tx.Rollback()` which is no longer called. We should add that to `txWithRetry` to ensure the transactions are properly closed when there is an error.

idb/postgres/postgres_migrations.go
tolikzinovyev: `upsertMigrationStateTx()` updates it before talking to DB. I don't think there is a good solution, especially here because `f()` can be called multiple time.

tolikzinovyev (commented):
Good catch about `tx.Rollback()`. I will make a new PR now.

tolikzinovyev:
Actually, the description of `txWithRetry()` says the function must call `Rollback()`.

idb/postgres/postgres_migrations.go
brianolson:
```go
nextState := *state
nextState.NextMigration++
...
*state = nextState
```

winder (approved):
LGTM, thanks for the changes!

PR #525, opened 2021-06-08
Creatable ID is wrong on round 0.

winder:
## Summary

Something about assets created on round 0 was leading to negative asset balances. Since it isn't possible to have transactions on round zero anyway, just skip it to avoid the problem.

## Test Plan

New unit tests.


brianolson (approved):

tolikzinovyev (approved):
I was just having the same issue with round 0 when testing new accounting.

PR #526, opened 2021-06-08
Update comment in setup_postgres.sql.

tolikzinovyev:


brianolson (approved):
I guess what it keeps in is more concise than what it leaves out

PR #528, opened 2021-06-08
Delete duplicate ImportState from importer

tolikzinovyev:


brianolson (approved):

winder (dismissed):
It looks like `ImportState` is now a postgres implementation detail. Can you move it from `idb` to `postgres`?

PR #536, opened 2021-06-11
Return an error in `CommitRoundAccounting()` when import state is uninitialized

tolikzinovyev:
## Summary

`CommitRoundAccounting()` currently proceeds to commit even if the import state is uninitialized. In `daemon.go` we, however, make sure to initialize the import state before starting the fetcher.

## Test Plan

Some existing tests have to do the import state initialization now by calling `db.LoadGenesis()`.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -1255,15 +1255,12 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
 	staterow := tx.QueryRow(`SELECT v FROM metastate WHERE k = 'state'`)
 	var stateJSONStr string
 	err = staterow.Scan(&stateJSONStr)
-	if err == sql.ErrNoRows {
-		// ok
-	} else if err != nil {
+	if err != nil {
 		return
-	} else {
-		err = encoding.DecodeJSON([]byte(stateJSONStr), &importState)
-		if err != nil {
-			return
-		}
+	}
+	err = encoding.DecodeJSON([]byte(stateJSONStr), &importState)
+	if err != nil {
+		return
 	}
 	if importState.AccountRound >= int64(round) {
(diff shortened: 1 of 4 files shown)


idb/postgres/postgres.go
```diff
@@ -1255,15 +1255,12 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
 	staterow := tx.QueryRow(`SELECT v FROM metastate WHERE k = 'state'`)
 	var stateJSONStr string
 	err = staterow.Scan(&stateJSONStr)
-	if err == sql.ErrNoRows {
```
winder, line 1258: If we're going to make changes here, maybe we should add `GetImportState` and `SetImportState` back. Then the error would be the `ErrorNotInitialized` error.

idb/postgres/postgres_integration_test.go
```diff
@@ -471,6 +486,9 @@ func TestRekeyBasic(t *testing.T) {
 	pdb, err := idb.IndexerDbByName("postgres", connStr, idb.IndexerDbOptions{}, nil)
 	assert.NoError(t, err)
 
+	err = pdb.LoadGenesis(test.MakeGenesis())
```
winder, line 489: Are there any tests that don't need the is? Maybe worth wrapping setupPostgres + IndexerDbByName + LoadGenesis in a helper?

idb/postgres/postgres.go
tolikzinovyev: Yes, I'm going to do it in another PR.

idb/postgres/postgres_integration_test.go
tolikzinovyev: It's probably a good idea. In another PR?

winder (approved):

PR #537, opened 2021-06-11
Refactoring in postgres

tolikzinovyev:
## Summary

- Re-add `getImportState()` and `setImportState()`
- Use them in `getMaxRoundAccounting()`
- Use `getMaxRoundAccounting()` in `commitRoundAccounting()`

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -37,7 +37,7 @@ import (
 
 type importState struct {
 	// Last accounted round.
-	AccountRound int64 `codec:"account_round"`
+	AccountRound *int64 `codec:"account_round"`
 }
 
 const stateMetastateKey = "state"
@@ -132,9 +132,9 @@ func (db *IndexerDb) txWithRetry(ctx context.Context, opts sql.TxOptions, f func
 }
 
 func (db *IndexerDb) init(opts idb.IndexerDbOptions) (err error) {
-	accountingStateJSON, _ := db.getMetastate(stateMetastateKey)
+	accountingStateJSON, _ := db.getMetastate(nil, stateMetastateKey)
 	hasAccounting := len(accountingStateJSON) > 0
-	migrationStateJSON, _ := db.getMetastate(migrationMetastateKey)
+	migrationStateJSON, _ := db.getMetastate(nil, migrationMetastateKey)
 	hasMigration := len(migrationStateJSON) > 0
 
 	db.GetSpecialAccounts()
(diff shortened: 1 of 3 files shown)


idb/postgres/postgres.go
```diff
@@ -358,43 +371,65 @@ func (db *IndexerDb) getMetastate(key string) (jsonStrValue string, err error) {
...
+		_, err = tx.Exec(setMetastateUpsert, key, jsonStrValue)
 	}
+	return
+}
 
-	var nullableRound sql.NullInt64
-	err = row.Scan(&nullableRound)
+// Returns idb.ErrorNotInitialized if uninitialized.
+// If `tx` is nil, use a normal query.
+func (db *IndexerDb) getImportState(tx *sql.Tx) (importState, error) {
```
winder, line 386: nit: this is pretty much the best-case scenario for when to name variables `(state importState, err error)`. It would let you avoid all of the `importState{}` and  `var state importState`.

idb/postgres/postgres.go
```diff
@@ -1251,30 +1286,27 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
 	if !any {
 		db.log.Debugf("empty round %d", round)
 	}
-	var importState importState
-	staterow := tx.QueryRow(`SELECT v FROM metastate WHERE k = 'state'`)
-	var stateJSONStr string
-	err = staterow.Scan(&stateJSONStr)
+
+	maxRoundAccounted, err := db.getMaxRoundAccounted(tx)
```
winder, line 1290: Need to use `getImportState` here. If we were to add another value to `importState` the logic here would clear it out each time we update accounting.

winder (changes requested):
Looks fine, one small change needed

idb/postgres/postgres.go
tolikzinovyev: I think it's nice to return an empty `importState{}` in case of an error, so the caller will behave weirdly if it doesn't check for errors, and tests can detect that.

idb/postgres/postgres.go
tolikzinovyev: fair

idb/postgres/postgres.go
winder: I'm not sure what you mean, the named variable would be initialized to `importState{}`

idb/postgres/postgres.go
```diff
@@ -1251,30 +1286,28 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
...
+	importstate, err := db.getImportState(tx)
 	if err != nil {
-		return
+		return err
 	}
-	err = encoding.DecodeJSON([]byte(stateJSONStr), &importState)
-	if err != nil {
-		return
+
+	if importstate.AccountRound == nil {
```
winder, line 1295: If I'm not mistaken, this is where `importState` should be initialized. So we need to allow the nil case on round 0:
```suggestion
	if importstate.AccountRound == nil && round != 0 {
```

winder: This would effect the other errors as well, `err.ErrorNotInitialized` is allowed, and `>= round` below needs to account for the first round.

idb/postgres/postgres.go
tolikzinovyev: Actually, I'm wrong about this case. But look at the previous implementation.
```
// GetImportState is part of idb.IndexerDB
func (db *IndexerDb) GetImportState() (state idb.ImportState, err error) {
	var importStateJSON string
	importStateJSON, err = db.getMetastate(stateMetastateKey)
	if err == sql.ErrNoRows || importStateJSON == "" {
		// no previous state, ok
		err = idb.ErrorNotInitialized
		return
	} else if err != nil {
		err = fmt.Errorf("unable to get import state: %v", err)
		return
	}

	err = encoding.DecodeJSON([]byte(importStateJSON), &state)
	if err != nil {
		err = fmt.Errorf("unable to parse import state: %v", err)
	}
	return
}
```

If `DecodeJSON` fails, it might have already written something to `state`. Then we would return an error with an initialized `state`.

idb/postgres/postgres.go
tolikzinovyev: The import state is initialized before the fetcher starts. Here it should already be initialized.

winder: I see, so this code only works because there are no transactions in round 0. Maybe that's always been the case.

winder (approved):

idb/postgres/postgres.go
tolikzinovyev: Yes, `commitRoundAccounting()` is not called with `round = 0`.

PR #540, opened 2021-06-11
Refactoring to make it easier to have subcommands.

winder:
## Summary

I'm planning to add a test runner subcommand to manage running tests. To make that easier this PR moves the current `block-generator` command to a cobra subcommand named `daemon`. There is also some minor refactoring that moves the  HTTP Server code into a server file to ensure no global values are used (i.e. config file / port).

With these changes when I add the test runner it should be fairly isolated.

## Test Plan

Made sure `run_generator.sh` and `docker-compose up` still work.


cmd/block-generator/generator/server.go
```diff
@@ -0,0 +1,109 @@
...
+	"os"
+	"strings"
+
+	"github.com/spf13/viper"
+
+	"github.com/algorand/indexer/util"
+)
+
+func initializeConfigFile(configFile string) (config GenerationConfig, err error) {
+	var f *os.File
```
tolikzinovyev, line 15: nit: An example where named return variables are not useful. You need to declare `f` separately. It's also very easy to create a bug by writing `f, err := os.Open(configFile)`, and you might return a non-`nil` `f` in case of an error.

cmd/block-generator/generator/server_test.go
```diff
@@ -0,0 +1,77 @@
...
+			expectedRound: 1234,
+			err:           "",
+		},
+	}
+
+	for _, testcase := range testcases {
+		t.Run(testcase.name, func(t *testing.T) {
+			round, err := parseRound(testcase.url)
+			if len(testcase.err) == 0 {
+				require.NoError(t, err, fmt.Sprintf("Unexpected error parsing '%s', expected round '%d' received error: %v", testcase.url, testcase.expectedRound, err))
```
tolikzinovyev, line 69: could you make this line shorter?

tolikzinovyev (commented):
What tests is the test runner going to run?

cmd/block-generator/generator/server.go
winder: you're right, the var definition isn't necessary because colon-equal does not shadow variables at the top level scope. Fixed.

cmd/block-generator/docker/Dockerfile-indexer
```diff
@@ -11,7 +11,7 @@ RUN go build
 CMD ["/bin/sh", "-c", "\
   echo $ALGOD_NET && \
   echo $CONNECTION_STRING &&\
-  sleep 10 && \
+  sleep 5 && \
```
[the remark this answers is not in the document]
winder: This sort of thing is a common problem with docker. The framework only concerns itself with the container starting, not the service within the container. This is a solution for Postgres, but [sandbox doesn't even do it properly](https://github.com/algorand/sandbox/issues/31):
```
# Wait for postgres to start.
while ! curl http://${DB_ADDR}/ 2>&1 | grep '52'
do
  sleep 1
done
```

There are also popular bash scripts that some people like to include in their deployments, [like wait-for-it](https://github.com/vishnubob/wait-for-it).

For our purposes, I think it's better to use sleep and overshoot the startup time rather than introduce something that might break.

cmd/block-generator/generator/server_test.go
winder: done

cmd/block-generator/runner.go
```diff
@@ -0,0 +1,25 @@
+package main
```
tolikzinovyev, line 1: let's not include this file

winder: Oops, this wasn't supposed to be included yet.

cmd/block-generator/daemon.go
```diff
@@ -0,0 +1,36 @@
...
+
+var configFile string
+var port uint64
+
+const configFileName = "block_generator_config"
+
+func init() {
+	rand.Seed(12345)
+
+	daemonCmd.Flags().StringVarP(&configFile, "config", "c", configFileName, fmt.Sprintf("Override default config file from '%s'.", configFileName))
```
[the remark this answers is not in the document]
winder: done

tolikzinovyev (approved):

PR #541, opened 2021-06-11
Replace `GetMaxRoundLoaded()` with `GetNextRoundToLoad()`

tolikzinovyev:
Also fixes a bug in `GetMaxRoundLoaded()`.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -399,24 +399,23 @@ func (db *IndexerDb) GetMaxRoundAccounted() (round uint64, err error) {
 	return db.getMaxRoundAccounted(nil)
 }
 
-// GetMaxRoundLoaded is part of idb.IndexerDB
-func (db *IndexerDb) GetMaxRoundLoaded() (round uint64, err error) {
-	var nullableRound sql.NullInt64
-	round = 0
+// GetNextRoundToLoad is part of idb.IndexerDB
+func (db *IndexerDb) GetNextRoundToLoad() (uint64, error) {
 	row := db.db.QueryRow(`SELECT max(round) FROM block_header`)
-	err = row.Scan(&nullableRound)
 
-	if err == sql.ErrNoRows || !nullableRound.Valid {
-		err = idb.ErrorNotInitialized
-		return
+	var nullableRound sql.NullInt64
+	err := row.Scan(&nullableRound)
+	if err == sql.ErrNoRows {
+		return 0, nil
 	}
-
 	if err != nil {
-		return
+		return 0, err
 	}
 
-	round = uint64(nullableRound.Int64)
-	return
+	if !nullableRound.Valid {
+		return 0, nil
+	}
+	return uint64(nullableRound.Int64), nil
 }
 
 // Break the read query so that PostgreSQL doesn't get bogged down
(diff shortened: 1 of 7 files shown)


idb/postgres/postgres.go
```diff
@@ -399,24 +399,23 @@ func (db *IndexerDb) GetMaxRoundAccounted() (round uint64, err error) {
...
-		return
+		return 0, err
 	}
 
-	round = uint64(nullableRound.Int64)
-	return
+	if !nullableRound.Valid {
+		return 0, nil
+	}
+	return uint64(nullableRound.Int64), nil
```
winder, line 418:
```suggestion
	return uint64(nullableRound.Int64) + 1, nil
```

idb/postgres/postgres.go
```diff
@@ -399,24 +399,23 @@ func (db *IndexerDb) GetMaxRoundAccounted() (round uint64, err error) {
...
 	row := db.db.QueryRow(`SELECT max(round) FROM block_header`)
-	err = row.Scan(&nullableRound)
 
-	if err == sql.ErrNoRows || !nullableRound.Valid {
-		err = idb.ErrorNotInitialized
-		return
+	var nullableRound sql.NullInt64
+	err := row.Scan(&nullableRound)
+	if err == sql.ErrNoRows {
+		return 0, nil
```
winder, line 409: Return  `idb.ErrorNotInitialized`?

idb/postgres/postgres_integration_test.go
```diff
@@ -44,14 +44,15 @@ func TestMaxRoundOnUninitializedDB(t *testing.T) {
...
+	roundL, errL := db.GetNextRoundToLoad()
 
 	//////////
 	// Then // The error message should be set.
 	//////////
 	assert.Equal(t, errA, idb.ErrorNotInitialized)
-	assert.Equal(t, errL, idb.ErrorNotInitialized)
 	assert.Equal(t, uint64(0), roundA)
+
+	require.NoError(t, errL)
```
winder, line 55: Why is this changing?

idb/postgres/postgres_integration_test.go
```diff
@@ -95,14 +96,14 @@ func TestMaxRound(t *testing.T) {
...
-	roundL, err := pdb.GetMaxRoundLoaded()
+	roundL, err := pdb.GetNextRoundToLoad()
 	assert.NoError(t, err)
 
 	//////////
 	// Then // There should be no error and we return that there are zero rounds.
 	//////////
 	assert.Equal(t, uint64(123454321), roundA)
-	assert.Equal(t, uint64(543212345), roundL)
+	assert.Equal(t, uint64(543212346), roundL)
```
winder, line 106: I'm confused about how this test is passing, where does the +1 happen?

winder (changes requested):
I'm not sure about these changes, seems like there could be a regression.

idb/postgres/postgres.go
tolikzinovyev: I think it shouldn't be a thing for importing blocks. With accounting it's genesis that needs to be initialized, here it's nothing.

idb/postgres/postgres.go
tolikzinovyev: right, thanks

idb/postgres/postgres_integration_test.go
tolikzinovyev: It was not. After fixing the bug you pointed out, it is.

winder (approved):

PR #542, opened 2021-06-14
Replace GetMaxRoundAccounted() with GetNextRoundToAccount()

tolikzinovyev:
## Summary

Currently the import state stores the last accounted round which creates ambiguities: we cannot distinguish between the state before accounting round 0 and after. This works because block 0 contains no transactions, and we don't actually do accounting for it. However, with one phase import it will be a problem.

This PR changes the import state to store the next round to account instead, `GetMaxRoundAccounted()` is replaced with `GetNextRoundToAccount()`, and a migration is added. Since older migrations run before that, `GetNextRoundToAccount()` still supports the old import state format. In most cases, however, `CommitRoundAccounting()` will upgrade import state to the new format before the migration.

## Test Plan

Wrote additional tests for `GetNextRoundToAccount()` to cover the old import state format, and added tests for the new migration.

### DIFF
--- a/accounting/rewind.go
+++ b/accounting/rewind.go
@@ -100,7 +100,7 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 		MaxRound: account.Round,
 	}
 	txns, r := db.Transactions(context.Background(), tf)
-	if r < account.Round {
+	if r <= account.Round {
 		err = ConsistencyError{fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round)}
 		return
 	}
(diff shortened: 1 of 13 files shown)


winder (changes requested):
Why do we need this change? I'm having a hard time seeing how it improves things, and it also seems pretty risky.

Since we have a lot of code here which would like to know `GetMaxRoundAccounted`, why not keep that function and add 1 in the new accounting code?

If it's too difficult for the other changes coming we could add a temporary `GetNextRoundToAccount` function which calls `GetMaxRoundAccounted()` and adds 1 (and overrides the error with 0). After the dust settles we can consolidate `GetMax` and `GetNext`.

tolikzinovyev:
Closing in favor of #559.

PR #544, opened 2021-06-16
Import performance test runner

winder:
## Summary

Adds a new subcommand to the block generator to run multiple scenarios and generate reports based on the results.

For testing postgres can be started in a local docker container with:
```
~$ docker run -it --rm --name some-postgres -e POSTGRES_PASSWORD=algorand -e POSTGRES_USER=algorand -e PGPASSWORD=algorand -e POSTGRES_DB=algorand -p 5432:5432 postgres
```

With all binaries built, the new command can be used from the `cmd/block-generator` directory with the following command:
```
~$ ./block-generator runner \
            -d 5m \
            -i ./../algorand-indexer/algorand-indexer \
            -c "host=localhost user=algorand password=algorand dbname=algorand port=5432 sslmode=disable" \
            -r results \
            -s scenarios
```

## Test Plan

Run the command and verify no errors, and that reports are generated.
```
~$ go build && rm -rf results && ./block-generator runner -d 5m -i ./../algorand-indexer/algorand-indexer -c "host=localhost user=algorand password=algorand dbname=algorand port=5432 sslmode=disable" -r results -s scenarios
Running test for configuration 'scenarios/config.asset.close.yml'
Running test for configuration 'scenarios/config.asset.destroy.yml'
Running test for configuration 'scenarios/config.asset.xfer.yml'
Running test for configuration 'scenarios/config.mixed.jumbo.yml'
Running test for configuration 'scenarios/config.mixed.yml'
Running test for configuration 'scenarios/config.payment.full.yml'
Running test for configuration 'scenarios/config.payment.jumbo.yml'
Running test for configuration 'scenarios/config.payment.small.yml'

~$ cat results/config.payment.jumbo.report 
test_duration_seconds: 300
transaction_count_pay:1999323
transaction_count_pay_create:40575
starting_block_import_duration_average_seconds:2.111194
final_import_duration_average_second:2.692307

~$ cat results/config.mixed.jumbo.report 
test_duration_seconds: 300
transaction_count_pay:294032
transaction_count_pay_create:5989
transaction_count_asset_close:34964
transaction_count_asset_create:712
transaction_count_asset_optin:71071
transaction_count_asset_xfer:593182
starting_block_import_duration_average_seconds:4.945975
final_import_duration_average_second:5.672932
```

### DIFF
--- a/cmd/block-generator/runner/metrics_collector.go
+++ b/cmd/block-generator/runner/metrics_collector.go
@@ -24,14 +24,14 @@ type Entry struct {
 
 // Collect fetches the metrics.
 func (r *MetricsCollector) Collect(prefix string) error {
-	metricsStr, err := r.getMetrics(prefix)
+	metrics, err := r.getMetrics(prefix)
 	if err != nil {
 		return err
 	}
 
-	if len(metricsStr) > 0 {
+	if len(metrics) > 0 {
 		entry := Entry{Timestamp: time.Now()}
-		for _, str := range metricsStr {
+		for _, str := range metrics {
 			entry.Data = append(entry.Data, str)
 		}
 		r.Data = append(r.Data, entry)
--- a/cmd/block-generator/runner/metrics_collector.go
+++ b/cmd/block-generator/runner/metrics_collector.go
@@ -30,9 +30,9 @@ func (r *MetricsCollector) Collect(prefix string) error {
 	}
 
 	if len(metrics) > 0 {
-		entry := Entry{Timestamp: time.Now()}
-		for _, str := range metrics {
-			entry.Data = append(entry.Data, str)
+		entry := Entry{
+			Timestamp: time.Now(),
+			Data: metrics,
 		}
 		r.Data = append(r.Data, entry)
 	}
--- a/cmd/block-generator/runner/metrics_collector.go
+++ b/cmd/block-generator/runner/metrics_collector.go
@@ -1,0 +1,69 @@
+package runner
+
+import (
+	"bufio"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// MetricsCollector queries a /metrics endpoint for prometheus style metrics and saves metrics matching a pattern.
+type MetricsCollector struct {
+	// MetricsURL where metrics can be queried.
+	MetricsURL string
+	// Data is all of the results.
+	Data []Entry
+}
+
+// Entry is the raw data pulled from the endpoint along with a timestamp.
+type Entry struct {
+	Timestamp time.Time
+	Data      []string
+}
+
+// Collect fetches the metrics.
+func (r *MetricsCollector) Collect(prefix string) error {
+	metricsStr, err := r.getMetrics(prefix)
+	if err != nil {
+		return err
+	}
+
+	if len(metricsStr) > 0 {
+		entry := Entry{Timestamp: time.Now()}
+		for _, str := range metricsStr {
+			entry.Data = append(entry.Data, str)
+		}
+		r.Data = append(r.Data, entry)
+	}
+
+	return nil
+}
+
+func (r MetricsCollector) getMetrics(prefix string) (result []string, err error) {
+	resp, err := http.Get(r.MetricsURL)
+	if err != nil {
+		err = fmt.Errorf("unable to read metrics url '%s'", r.MetricsURL)
+		return
+	}
+	defer resp.Body.Close()
+
+	scanner := bufio.NewScanner(resp.Body)
+	for scanner.Scan() {
+		str := scanner.Text()
+
+		if strings.HasPrefix(str, "#") {
+			continue
+		}
+
+		if strings.HasPrefix(str, prefix) {
+			result = append(result, str)
+		}
+	}
+
+	if scanner.Err() != nil {
+		err = fmt.Errorf("problem reading metrics response: %w", scanner.Err())
+	}
+
+	return
+}
--- a/cmd/block-generator/runner/metrics_collector.go
+++ b/cmd/block-generator/runner/metrics_collector.go
@@ -32,7 +32,7 @@ func (r *MetricsCollector) Collect(prefix string) error {
 	if len(metrics) > 0 {
 		entry := Entry{
 			Timestamp: time.Now(),
-			Data: metrics,
+			Data:      metrics,
 		}
 		r.Data = append(r.Data, entry)
 	}
--- a/cmd/block-generator/runner/run.go
+++ b/cmd/block-generator/runner/run.go
@@ -1,13 +1,20 @@
 package runner
 
 import (
+	"bytes"
 	"context"
+	"database/sql"
 	"fmt"
-	"github.com/algorand/indexer/cmd/block-generator/generator"
+	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"time"
 
+	// Load the postgres sql.DB implementation
+	_ "github.com/lib/pq"
+
+	"github.com/algorand/indexer/cmd/block-generator/generator"
 	"github.com/algorand/indexer/util"
 )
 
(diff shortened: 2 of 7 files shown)


cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	sum := 0.0
+	count := 0.0
+
+	for _, metric := range entry.Data {
+		var err error
+		if strings.HasPrefix(metric, "indexer_daemon_import_time_sec_sum") {
+			val := strings.Split(metric, " ")[1]
+			sum, err = strconv.ParseFloat(val, 64)
+		}
+		if strings.HasPrefix(metric, "indexer_daemon_import_time_sec_count") {
```
brianolson, line 104: else if

brianolson (commented):
there's a lot here but it all looks good. I could only find one tiny style quibble.

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Collect results.
+	durationStr := fmt.Sprintf("test_duration_seconds:%d\n", uint64(r.RunDuration.Seconds()))
+	if _, err := report.WriteString(durationStr); err != nil {
+		return fmt.Errorf("unable to write duration metric: %w", err)
+	}
+
+	resp, err := http.Get(fmt.Sprintf("http://%s/report", generatorURL))
+	if err != nil {
+		return fmt.Errorf("the process failed to start properly, health endpoint query failed")
```
winder, line 155: Wrong comment
```suggestion
		return fmt.Errorf("generator report query failed")
```

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+
+// startIndexer resets the postgres database and executes the indexer binary. It performs some simple verification to
+// ensure that the service has started properly.
+func startIndexer(indexerBinary string, algodNet string, indexerNet string, postgresConnectionString string) (func() error, error) {
+	{
+		db, err := sql.Open("postgres", postgresConnectionString)
+		if err != nil {
+			return nil, fmt.Errorf("postgres connection string did not work: %w", err)
+		}
+		db.Exec(`DROP SCHEMA public CASCADE; CREATE SCHEMA public;`)
```
tolikzinovyev, line 217: Could you comment what this does?

tolikzinovyev: we should probably check the returned error

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+		db.Close()
```
tolikzinovyev, line 218: we should probably check the returned error

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	time.Sleep(250 * time.Millisecond)
```
tolikzinovyev, line 221: what does this do?

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, fmt.Errorf("the process failed to start properly, health endpoint query failed")
+	}
+	resp.Body.Close()
+
+	return func() error {
+		if err := cmd.Process.Kill(); err != nil {
+			return fmt.Errorf("failed to kill indexer process: %w", err)
+		}
+
+		// Clear postgres DB
```
tolikzinovyev, line 257: ?

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+		if err := collector.Collect("indexer_daemon_import_time_sec"); err != nil {
+			return fmt.Errorf("problem collecting metrics: %w", err)
+		}
+	}
+	if err := collector.Collect("indexer_daemon_import_time_sec"); err != nil {
+		return fmt.Errorf("problem collecting metrics: %w", err)
+	}
+
+	// Collect results.
+	durationStr := fmt.Sprintf("test_duration_seconds:%d\n", uint64(r.RunDuration.Seconds()))
```
tolikzinovyev, line 148: we should probably report the actual elapsed time

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	return nil
+}
+
+// startGenerator starts the generator server.
+func startGenerator(configFile string, addr string) func() error {
+	// Start generator.
+	server, done := generator.StartServer(configFile, addr)
+
+	return func() error {
+		if err := server.Shutdown(context.Background()); err != nil {
```
tolikzinovyev, line 194: From the description, `Shutdown()` blocks until the server is shut down and `ListenAndServe()` returns immediately. So, `done` will be closed immediately. On a related note, I think we shouldn't worry about shutting down the generator gracefully.

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	if err := indexerShutdownFunc(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Helper to record the import rate.
+func recordDataToFile(entry Entry, key string, out *os.File) error {
+	sum := 0.0
```
tolikzinovyev, line 95: Could you add units to the variable name? Maybe `totalSeconds`?

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Record a rate from one of the first data points.
+	if len(collector.Data) > 5 {
+		if err := recordDataToFile(collector.Data[2], "starting_block_import_duration_average_seconds", report); err != nil {
+			return err
+		}
+	}
+
+	// Also record the final one.
+	if err := recordDataToFile(collector.Data[len(collector.Data)-1], "final_import_duration_average_second", report); err != nil {
```
tolikzinovyev, line 181: `final_import_duration_average_seconds`

cmd/block-generator/runner/run.go
winder: The units are in the key, this function just divides the two numbers.

cmd/block-generator/runner/run.go
winder: Both these functions block. I happen to be reusing the same port for each run, so it's important that the generator and indexer both shut down before starting the next test.

cmd/block-generator/runner/run.go
winder: I moved this to the beginning. Removed comment.

cmd/block-generator/runner/run.go
winder: Looks like this isn't necessary. Removed.

cmd/block-generator/runner/run.go
winder: added

cmd/block-generator/runner/run.go
winder: Resets the DB. The returned error describes what it's trying to do now

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+			continue
+		}
+		str := fmt.Sprintf("transaction_%s_total:%d\n", metric, entry.GenerationCount)
+		if _, err := report.WriteString(str); err != nil {
+			return fmt.Errorf("unable to write transaction_count metric: %w", err)
+		}
+	}
+
+	// Record a rate from one of the first data points.
+	if len(collector.Data) > 5 {
```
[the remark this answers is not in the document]
winder: No, the metrics are aggregated. When we call `importTimeHistogramSeconds.Observe(dt.Seconds())` it increments a counter and adds `dt.Seconds()` to a sum.

In this code I happen to be collecting the aggregated metrics 11 times during the course of the test, but if we wanted graphs it would be better to hook up a Prometheus DB and query it directly.

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	msg := fmt.Sprintf("%s:%f\n", key, rate)
+	if _, err := out.WriteString(msg); err != nil {
+		return fmt.Errorf("unable to write metric '%s': %w", key, err)
+	}
+	return nil
+}
+
+// Run the test for 'RunDuration', collect metrics and write them to the 'ReportDirectory'
+func (r *Args) runTest(indexerURL string, generatorURL string) error {
+	collector := &MetricsCollector{MetricsURL: fmt.Sprintf("http://%s/metrics", indexerURL)}
```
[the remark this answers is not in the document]
winder: The data field is an array so it doesn't need to be initialized.

cmd/block-generator/runner/metrics_collector.go
```diff
@@ -0,0 +1,69 @@
...
+// Collect fetches the metrics.
+func (r *MetricsCollector) Collect(prefix string) error {
+	metricsStr, err := r.getMetrics(prefix)
+	if err != nil {
+		return err
+	}
+
+	if len(metricsStr) > 0 {
+		entry := Entry{Timestamp: time.Now()}
+		for _, str := range metricsStr {
```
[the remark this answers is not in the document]
winder: This code changed and I didn't update the variable name. This is a clice not a string.

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+			}
+			runnerArgs := args
+			runnerArgs.Path = path
+			fmt.Printf("Running test for configuration '%s'\n", path)
+			return runnerArgs.run()
+		})
+	}
+
+	// Single file mode
+	return args.run()
```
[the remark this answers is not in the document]
winder: Nice :+1:

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	if _, err := os.Stat(args.ReportDirectory); !os.IsNotExist(err) {
+		return fmt.Errorf("report directory '%s' already exists", args.ReportDirectory)
+	}
+	os.Mkdir(args.ReportDirectory, os.ModeDir|os.ModePerm)
+
+	// Batch mode
+	pathStat, err := os.Stat(args.Path)
+	util.MaybeFail(err, "Unable to check path.")
+	if pathStat.IsDir() {
+		return filepath.Walk(args.Path, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
winder: I don't think it's necessary, the intent is a "directory containing scenarios". If it contains other directories with incompatible files the program will fail with a useful error message.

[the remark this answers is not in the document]
winder: Added the check for err in `run()`

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	// Single file mode
+	return args.run()
+}
+
+func (r *Args) run() error {
+	// Start services
+	algodNet := fmt.Sprintf("localhost:%d", 11112)
+	indexerNet := fmt.Sprintf("localhost:%d", r.IndexerPort)
+	generatorShutdownFunc := startGenerator(r.Path, algodNet)
+	indexerShutdownFunc, err := startIndexer(r.IndexerBinary, algodNet, indexerNet, r.PostgresConnectionString)
```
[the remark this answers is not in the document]
winder: I don't think adding new lines would help with readability. Almost every line in this function has an error to handle, so I intentionally organized the top-level run function so that all of the vertical space/indents are used for the errors.

cmd/block-generator/runner/metrics_collector.go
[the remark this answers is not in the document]
tolikzinovyev: Right. Both `metrics` and `entry.Data` are `[]string`, so we can do `entry.Data = mertics`.

cmd/block-generator/runner/run.go
winder: Yes, this should exit immediately if everything is working correctly. If it doesn't there's a 10 second timeout to make sure nothing hangs.

cmd/block-generator/runner/run.go
```diff
@@ -0,0 +1,260 @@
...
+	ReportDirectory          string
+
+	indexerPort uint64
+}
+
+// Run is a public helper run the tests.
+// The test will run against the generator configuration file specified by 'args.Path'.
+// If 'args.Path' is a directory it should contain generator configuration files, a test will run using each file.
+func Run(args Args) error {
+	if _, err := os.Stat(args.ReportDirectory); !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
winder: Turns out, I'm not sure the point of `os.IsExist`. This interface is not very good for checking that a directory exists because the check would need to be `err == nil || os.IsExist(err)`!

I'll grant that the double negative is ugly, but the alternative is too!

cmd/block-generator/runner/metrics_collector.go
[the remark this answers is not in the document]
winder: Nice, updated!

cmd/block-generator/runner/run.go
winder: done

cmd/block-generator/runner/run.go
winder: Fixed, thanks!

tolikzinovyev (approved):
Great!

PR #547, opened 2021-06-18
Update protocols.json to v28

tolikzinovyev:
Closes https://github.com/algorand/go-algorand/issues/2257 .


tsachiherman (dismissed):
I think that these changes are good, but we also need the corresponding changes in types/types.go. See previous update as a reference : https://github.com/algorand/indexer/commit/a014e8410a85773831a3fcd23c8bdaa8d70c139e

winder (approved):

brianolson (approved):

PR #549, opened 2021-06-21
fix usage of e2e_client_runner.py which changed out from under us

brianolson:
## Summary

Fix the `misc/buildtestdata.sh` script for new usage of `go-algorand/test/scripts/e2e_client_runner.py`

## Test Plan

This is mostly a developer tool. Checked in my environment.


winder (approved):
LGTM, I guess the client runner changed its CWD requirement?

PR #557, opened 2021-07-01
Add devmode support to indexer

tsachiherman:
## Summary

Add the `DevMode` to the genesis definition to support devmode nodes.

## Test Plan

Manual testing.


tsachiherman:
@tolikzinovyev  - do we really need this with your upcoming changes ? ( I think the answer is "no", but I just want to make sure )

tolikzinovyev (approved):
The plan is to remove this file, but I think it doesn't hurt to update it now. Thanks.

PR #558, opened 2021-07-02
Delete old migrations

tolikzinovyev:
## Summary

Delete migration that were introduced before version 2.4.0.

## Test Plan

Added a test for the scenario when a new database is initialized.
Manually tested upgrading (unsuccessfully) from version 2.2.1.


idb/postgres/postgres_migrations.go
```diff
@@ -4,65 +4,25 @@
...
-		// Migrations for 2.3.1 release
-		{m5MarkTxnJSONSplit, true, "record round at which txn json recording changes, for future migration to fixup prior records"},
-		{m6RewardsAndDatesPart1, true, "Update DB Schema for cumulative account reward support and creation dates."},
-		{m7RewardsAndDatesPart2, false, "Compute cumulative account rewards for all accounts."},
-
-		// Migrations for 2.3.2 release
-		{m8StaleClosedAccounts, false, "clear some stale data from closed accounts"},
-		{m9TxnJSONEncoding, false, "some txn JSON encodings need app keys base64 encoded"},
-		{m10SpecialAccountCleanup, false, "The initial m7 implementation would miss special accounts."},
-		{m11AssetHoldingFrozen, false, "Fix asset holding freeze states."},
```
winder, line 50: What do you think about keeping placeholders? There are a few nice benefits:
1) no need to worry about the changing the index?
2) we could replace the existing migration functions with a special `unsupportedMigration` function that errors unconditionally. The comment field could be the last version supported like "2.5.0" to go in a message like "Please downgrade to 2.5.0 and complete migration before installing this version."
3) easier to deprecate future migrations, just delete the function and change the placeholder function to `unsupportedMigration` and note the last supported version.

winder (commented):
Looks great, just one comment

idb/postgres/postgres_migrations.go
tolikzinovyev: Thanks, I did some something similar.

idb/postgres/postgres_migrations.go
```diff
@@ -4,29 +4,17 @@
...
 	"github.com/algorand/indexer/idb/migration"
 	"github.com/algorand/indexer/idb/postgres/internal/encoding"
-	"github.com/algorand/indexer/types"
 )
 
-// rewardsMigrationIndex is the index of m7RewardsAndDatesPart2.
-const rewardsMigrationIndex = 7
-
 func init() {
 	migrations = []migrationStruct{
```
winder, line 19:
```suggestion
	// To deprecate old migrations change the functions to return a `unsupportedMigrationErrorMsg` error.
	// Make sure you set the blocking flag to true to avoid possible consistency issues during startup.
	migrations = []migrationStruct{
```

idb/postgres/postgres_migrations.go
tolikzinovyev: The only issue now is that indexer doesn't exit if can't run a migration, but keeps hanging instead. It might be worth improving the migration engine. What do you think?

winder (approved):
Looks good, just one small suggestion left.

PR #559, opened 2021-07-02
Store next round in import state

tolikzinovyev:
## Summary

Currently the import state stores the last accounted round which creates ambiguities: we cannot distinguish between the state before accounting round 0 and after. This works because block 0 contains no transactions, and we don't actually do accounting for it. However, with one phase import it will be a problem.

This PR changes the import state to store the next round to account instead, `GetMaxRoundAccounted()` is replaced with `GetNextRoundToAccount()`, and a blocking migration is added.

## Test Plan

Added tests for the migration, and manually tested the migration when max round accounted is 0 and >0.

### DIFF
--- a/idb/postgres/postgres_migrations.go
+++ b/idb/postgres/postgres_migrations.go
@@ -42,6 +42,7 @@ func init() {
 		{FixFreezeLookupMigration, false, "Fix search by asset freeze address."},
 		{ClearAccountDataMigration, false, "clear account data for accounts that have been closed"},
 		{MakeDeletedNotNullMigration, false, "make all \"deleted\" columns NOT NULL"},
+		{MaxRoundAccountedMigration, true, "change import state format"},
 	}
 }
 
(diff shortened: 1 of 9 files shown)


idb/postgres/postgres_migrations.go
```diff
@@ -99,27 +100,16 @@ func needsMigration(state MigrationState) bool {
...
-// upsertMigrationState updates the migration state, and optionally increments the next counter.
-func upsertMigrationState(db *IndexerDb, state *MigrationState, incrementNextMigration bool) (err error) {
-	if incrementNextMigration {
-		state.NextMigration++
-	}
-	migrationStateJSON := encoding.EncodeJSON(state)
-	_, err = db.db.Exec(setMetastateUpsert, migrationMetastateKey, migrationStateJSON)
-
-	return err
+	return db.setMetastate(tx, migrationMetastateKey, string(migrationStateJSON))
```
winder, line 112: This needs the weird post-success-next-migration-update thing because it might be used inside a retry function.

idb/postgres/postgres_migrations.go
```diff
@@ -538,3 +528,56 @@ func MakeDeletedNotNullMigration(db *IndexerDb, state *MigrationState) error {
...
 	return sqlMigration(db, state, queries)
 }
+
+// MaxRoundAccountedMigration converts the import state.
+func MaxRoundAccountedMigration(db *IndexerDb, migrationState *MigrationState) error {
+	db.accountingLock.Lock()
+	defer db.accountingLock.Unlock()
+
+	nextMigrationState := *migrationState
+	nextMigrationState.NextMigration++
```
winder, line 538: Ok, I see you've handled it here instead. This is better (since Commit can fail), but it does mean we need to remember to do this for every migration.

Maybe we can find a way to auto-increment this in the future.

idb/postgres/postgres_migrations.go
```diff
@@ -538,3 +528,56 @@ func MakeDeletedNotNullMigration(db *IndexerDb, state *MigrationState) error {
...
+		defer tx.Rollback()
+
+		importstate, err := db.getImportState(tx)
+		if err == idb.ErrorNotInitialized {
+			// Leave uninitialized.
+			db.log.Printf("Import state is not initialized, leaving unchanged.")
+		} else if err != nil {
+			return err
+		} else {
+			if importstate.AccountRound == nil {
```
winder, line 550: I think you can move this to `if err == idb.ErrorNotInitialized || importstate.AccountRound == nil {` and avoid the extra indent. It probably ends up in that block anyway.

idb/postgres/postgres_migrations.go
```diff
@@ -538,3 +528,56 @@ func MakeDeletedNotNullMigration(db *IndexerDb, state *MigrationState) error {
...
+				}
+			}
+		}
+
+		err = upsertMigrationState(db, tx, &nextMigrationState, false)
+		if err != nil {
+			return err
+		}
+
+		return tx.Commit()
```
winder, line 574: The upsert/commit can move into the else block, they're only needed when the `NextRoundToAccount` is actually changed/initialized.

winder (changes requested):
Looks good, thanks for making these changes. Just a couple nits

idb/postgres/postgres_migrations.go
tolikzinovyev: Yes, finding a way to auto-increment would be nice.

idb/postgres/postgres_migrations.go
tolikzinovyev: No, this would be a bug if `err != nil` and `err != idb.ErrorNotInitialized`!

idb/postgres/postgres_migrations.go
tolikzinovyev: No, we need to increment the migration counter in all cases.

winder (approved):
Thank for the explanations, I'm convinced. LGTM.

PR #567, opened 2021-07-12
Support assets with null characters in strings.

winder:
## Summary

It's possible for users to put null characters in the asset name, asset unit, and asset URL. The indexer should handle these valid transactions in some sensible way.
## Test Plan

New unit tests.


idb/postgres/internal/encoding/encode.go
```diff
@@ -53,7 +54,52 @@ func convertEvalDelta(evalDelta types.EvalDelta) types.EvalDelta {
...
+	return str
+}
+
+// SanitizeNullForQuery converts a string into something postgres can store in a jsonb column.
+func SanitizeNullForQuery(str string) string {
+	if str != "" {
+		result := strings.ReplaceAll(str, "\x00", "\\u0000")
+		if len(result) != len(str) {
+			// Escape the escapes if this is going to be used in a query.
+			return strings.ReplaceAll(result, "\\", "\\\\")
```
brianolson, line 70: I think you need to replace existing backslashes before adding backslashes (this needs to happen beforre line 67)

brianolson (commented):
I think we should probably always do like the '...ForQuery' variant and escape backslashes and then zeros; then the desanitize() function will actually do the right thing.

idb/postgres/internal/encoding/encode.go
winder: This is specifically to search for `\u0000`, if you don't escape the slash it doesn't find anything.

I'm simplifying this to be just `return strings.ReplaceAll(str, "\x00", "\\\\u0000").

I think that all other unicode characters would be converted by the json encoder with a single backslash, postgres must convert them into the unicode character.

idb/postgres/internal/encoding/encoding_test.go
```diff
@@ -84,3 +84,52 @@ func TestJSONEncoding(t *testing.T) {
 
 	assert.Equal(t, x, xx)
 }
+
+func TestSanitizeNull(t *testing.T) {
```
brianolson, line 88: I think the test I want to see is a round trip desanitize(sanitize(x)) == x; 
and that will fail for x := "foo\\u0000bar" which already has the escaped-zero code in it which will become actual zero on its way back out through desanitize.

idb/postgres/internal/encoding/encode.go
```diff
@@ -53,7 +54,39 @@ func convertEvalDelta(evalDelta types.EvalDelta) types.EvalDelta {
 	return evalDelta
 }
 
+func desanitizeNull(str string) string {
```
tolikzinovyev, line 57: I think, we shouldn't desanitize. Essentially, we need to choose between the zero byte and `\u0000` in the string. If we don't desanitize, then the zero byte suffers, if we do desanitize, then `\u0000` suffers. I would prefer that the zero byte suffers.

idb/postgres/internal/encoding/encode.go
```diff
@@ -53,7 +54,39 @@ func convertEvalDelta(evalDelta types.EvalDelta) types.EvalDelta {
...
+	return strings.ReplaceAll(str, "\x00", `\\u0000`)
+}
+
+// sanitizeNull converts a string into something postgres can store in a jsonb column.
+func sanitizeNull(str string) string {
+	return strings.ReplaceAll(str, "\x00", `\u0000`)
+}
+
+// SanitizeParams sanitizes all AssetParams that need it.
+func SanitizeParams(params types.AssetParams) types.AssetParams {
```
tolikzinovyev, line 72: The convention I had in mind is: functions like `EncodeSignedTxnWithAD()` which calls `convertSignedTxnWithAD()` and does json encoding in this file, and the code outside that only uses `EncodeSignedTxnWithAD()`. So, `SanitizeParams()` should be renamed to `convertAssetParams()` and we need a function called `EncodeAssetParams()`.

The decoding functions should be in `decode.go` instead, but I think we shouldn't have them. They should have a similar convention.

winder: I think this suggestion breaks things down more than is useful. I'll make some updates to fix the naming conventions.

idb/postgres/internal/encoding/encode.go
```diff
@@ -63,13 +127,123 @@ func EncodeSignedTxnWithAD(stxn types.SignedTxnWithAD) []byte {
...
+			continue
+		}
+		// TODO: unicode.IsPrint(c) ?
+		switch c {
+		case 0:
+			newlen += 6 // \u0000
+		case '\\':
+			newlen += 2
+		default:
+			newlen+= csize
```
tsachiherman, line 148: add space before `+=`

idb/postgres/internal/encoding/encode.go
```diff
@@ -63,13 +127,123 @@ func EncodeSignedTxnWithAD(stxn types.SignedTxnWithAD) []byte {
...
+		i += csize
+	}
+	if start < len(xb) {
+		copy(out[outpos:], xb[start:])
+	}
+	return string(out)
+}
+
+// UnescapeNulls is the inverse function of EscapeNulls.
+// UnescapeNulls converts \\ and \uXXXX back into their unescaped form but may not be fully general for input not generated by EscapeNulls().
```
tsachiherman, line 202: I believe the method is correct - but relying on the fact that it was correctly generated by EscapeNulls() is an issue. It need to be refactored to return an error in case it's not possible to unescape.

idb/postgres/internal/encoding/encode.go
```diff
@@ -63,13 +128,122 @@ func EncodeSignedTxnWithAD(stxn types.SignedTxnWithAD) []byte {
...
+
+func EscapeNulls(x string) string {
+	if nothingToEscape(x) {
+		return x
+	}
+
+	xb := []byte(x)
+
+	escapenull := []byte("\\u0000")
+	var out strings.Builder
```
brianolson, line 159: This is worse. The prior implementation will produce less memory garbage by doing exactly one allocation of exactly the right size. Why replace the better implementation with a worse implementation? (I know, readability, but the better implementation already existed so moving backwards is frustrating.)

idb/postgres/internal/encoding/encode.go
brianolson: If this is all only used internally to store a few fields and read them back, then there is no problem?

winder:
Replaced by https://github.com/algorand/indexer/pull/577

Issue #566, opened 2021-07-11
non utf8 support for asset strings


jeapostrophe:
I got some txns committed to an algod round and got this error from the indexer trying to import them:

```
INDEX:{"level":"info","msg":"Starting block importer.","time":"2021-07-11T04:20:18Z"}

INDEX:{"level":"error","msg":"CommitBlock failed: pq: unsupported Unicode escape sequence (&pq.Error{Severity:\"ERROR\", Code:\"22P05\", Message:\"unsupported Unicode escape sequence\", Detail:\"\\\\u0000 cannot be converted to text.\", Hint:\"\", Position:\"\", InternalPosition:\"\", InternalQuery:\"\", Where:\"JSON data, line 1: ...QncyBzaGlueSEAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\\\",\\\"an\\\":...\\nCOPY txn, line 2, column txn: \\\"{\\\"lsig\\\":{\\\"l\\\":\\\"BCABATIEIglJOBCBBhJEOBiBBBJDIg==\\\"},\\\"txn\\\":{\\\"apar\\\":{\\\"am\\\":\\\"SXQncyBzaGlueSEAAAAAAAAAAAAAAA...\\\"\", Schema:\"\", Table:\"\", Column:\"\", DataTypeName:\"\", Constraint:\"\", File:\"json.c\", Line:\"878\", Routine:\"json_lex_string\"})","time":"2021-07-11T04:20:18Z"}

[...]

PR #568, opened 2021-07-13
backslash escape bad utf8 and nulls

brianolson:
## Summary

PostgreSQL can't handle zero bytes in strings in jsonb, and _might_ do weird things with invalid utf8, so backslash escape those things, and have a function for reversing that.
Alternate implementation to #567 

## Test Plan

Adds unit test.


idb/postgres/internal/encoding/encode.go
```diff
@@ -53,7 +55,30 @@ func convertEvalDelta(evalDelta types.EvalDelta) types.EvalDelta {
 	return evalDelta
 }
 
+// SanitizeNullForQuery converts a string for putting into a SELECT ... WHERE
+func SanitizeNullForQuery(str string) string {
+	return EscapeNulls(str) // TODO: does Postgres query need double-escaping?
```
winder, line 60: Alas, the answer is yes. One of the failures tests exactly this case:
```
--- FAIL: TestEmbeddedNullString (87.44s)
    postgres_integration_test.go:76: 
        	Error Trace:	postgres_integration_test.go:76
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestEmbeddedNullString
```

Maybe we need to triple escape now?

idb/postgres/internal/encoding/encode.go
```diff
@@ -73,3 +98,125 @@ func init() {
 	jsonCodecHandle.Indent = 0
 	jsonCodecHandle.MapKeyAsString = true
 }
+
+// EscapeNulls converts null bytes, invalid utf8, and backslashes into backslash escape sequences \uXXXX
```
winder, line 102: Is there anything we can use in `%q`? It seems very similar, we would have to strip the wrapped quotes and add an extra layer of escapes.
```
func TestQuote(t *testing.T) {
	invalid := "\xc0"
	quoted := fmt.Sprintf("%q", invalid)
	require.Equal(t, `"\xc0"`, quoted)
	unquoted, e := strconv.Unquote(quoted)
	require.NoError(t, e)
	require.Equal(t, invalid, unquoted)
}
```

brianolson:
fix for #566

idb/postgres/postgres_integration_test.go
```diff
@@ -31,6 +31,79 @@ func getAccounting(round uint64, cache map[uint64]bool) *accounting.State {
...
+	for asset := range assets {
+		require.NoError(t, asset.Error)
+		require.Equal(t, nameWithNull, asset.Params.AssetName)
+		require.Equal(t, unitWithNull, asset.Params.UnitName)
+		require.Equal(t, urlWithNull, asset.Params.URL)
+		copy(creator[:], asset.Creator[:])
+		num++
+	}
+	require.Equal(t, 1, num)
+
```
winder, line 77: I had one more test in my branch that throws a wrench in things a little bit. The user isn't actually going to be able to search for `nameWithNull`, or invalid UTF8, so the main search parameter would be something like this:

```
	assets, _ = db.Assets(context.Background(), idb.AssetsQuery{Name: encoding.EncodeStringForQuery(nameWithNull)})
	num = 0
	for asset := range assets {
		require.NoError(t, asset.Error)
		require.Equal(t, encodedNameWithNull, asset.Params.AssetName)
		require.Equal(t, encodedUnitWithNull, asset.Params.UnitName)
		require.Equal(t, encodedUrlWithNull, asset.Params.URL)
		num++
	}
	require.Equal(t, 1, num)
```

I'm not sure we can support this and `idb.AssetsQuery{Name: nameWithNull}`

idb/postgres/postgres.go
```diff
@@ -2589,8 +2589,9 @@ func (db *IndexerDb) Assets(ctx context.Context, filter idb.AssetsQuery) (<-chan
 		partNumber++
 	}
 	if filter.Name != "" {
+		queryName := encoding.SanitizeNullForQuery(filter.Name)
```
winder, line 2592: To solve the problem I mentioned, we just have to remove this. Now the escaped format will be accepted and if you submit a search with a null character you'll get a malformed query error.

brianolson: Huh? I just got convinced we need to do this. I'd expect we need to do at least the regular forward EscapeNulls(), but then by integration test we can see that lib/pq isn't escaping that right so we need to double escape it; so what chain of events lead to us not needing to do any escaping at all?

winder:
https://github.com/algorand/indexer/pull/577

PR #569, opened 2021-07-13
Validator improvements.

winder:
## Summary
Validator improvements:
- Add retry count to error results.
- Add early exit to retry loop when the rounds are equal.

### DIFF
--- a/cmd/validator/validate_accounting.go
+++ b/cmd/validator/validate_accounting.go
@@ -43,7 +43,8 @@ type Processor interface {
 // Result is the output of ProcessAddress.
 type Result struct {
 	// Error is set if there were errors running the test.
-	Error error
+	Error     error
+	SameRound bool
 
 	Equal   bool
 	Retries int
@@ -235,7 +236,7 @@ func callProcessor(processor Processor, addrInput string, config Params, results
 
 		result, err := processor.ProcessAddress(algodData, indexerData)
 
-		if err == nil && (result.Equal || i >= config.retries) {
+		if err == nil && (result.Equal || result.SameRound || i >= config.retries) {
 			// Return when results are equal, or when finished retrying.
 			result.Retries = i
 			if result.Details != nil {
@@ -318,6 +319,8 @@ func resultsPrinter(config Params, printCurl bool, results <-chan Result) int {
 			errorLog.Printf("%s", time.Now().Format("2006-01-02 3:4:5 PM"))
 			errorLog.Printf("Account: %s", r.Details.address)
 			errorLog.Printf("Error #: %d", numErrors)
+			errorLog.Printf("Retries: %d", r.Retries)
+			errorLog.Printf("Rounds Match: %t", r.SameRound)
 
 			// Print error message if there is one.
 			if r.Error != nil {
--- a/cmd/validator/struct_processor.go
+++ b/cmd/validator/struct_processor.go
@@ -32,8 +32,9 @@ func (gp StructProcessor) ProcessAddress(algodData, indexerData []byte) (Result,
 	differences := equals(indexerAcct, algodAcct)
 	if len(differences) > 0 {
 		return Result{
-			Equal:   false,
-			Retries: 0,
+			Equal:     false,
+			SameRound: indexerAcct.Round == algodAcct.Round,
+			Retries:   0,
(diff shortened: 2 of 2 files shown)


cmd/validator/validate_accounting.go
```diff
@@ -235,7 +236,7 @@ func callProcessor(processor Processor, addrInput string, config Params, results
 
 		result, err := processor.ProcessAddress(algodData, indexerData)
 
-		if err == nil && (result.Equal || i >= config.retries) {
+		if err == nil && (result.Equal || result.SameRound || i >= config.retries) {
```
tolikzinovyev, line 239: I think `result.Equal` is unnecessary.

winder: Since the reward base only changes every 30 rounds or so, most of the time it doesn't matter if we're on the same round. In that case `result.Equal` can be true and `result.SameRound` false.

tolikzinovyev: I would like to stay on the safe side and only return results with the same round. Efficiency shouldn't suffer since most of the time algod and indexer are synchronized.

winder: There isn't any risk because the comparison has to succeed for the test to pass. Why bother waiting and trying again if the test has already passed?

tolikzinovyev: Because two results having same data for different rounds doesn't mean they will have same data for the same round.

winder: If we were off by an entire round, then we would see failures when accounts are checked and indexer/algod are on the same round.

I wouldn't mind a strict mode flag which does this, but considering the test already takes something like 12 hours to run, I think we should let it run as fast as possible rather than worry about this case.

tolikzinovyev: Like I said before, I only saw dots in the output and no numbers, which means there are actually very few retries in practice. So, efficiency shouldn't suffer. But I would like to have more reliability.

winder: IIRC after the 12 hours, there will be hundreds or low thousands of retries.

Let's save this suggestion for a follow-up when we can also measure the impact

winder: I was lamenting the lack of retrys in all the old logs on my test machine, but I found some in nohup.out:
```
Number of errors: [112 / 9969050]
Retry count: 2365
Checks per second: 160.449854
Test duration: 17:15:46
```

tolikzinovyev: Ok, 2365 * 50ms = 2 minutes of additional time. I think, that's reasonable.

winder: By my calculation is was something between 15-39 minutes. The retry delay defaults to 1 second.

tolikzinovyev: You are right. But can we make the retry delay 100ms? We can also make the number of retries very large because now we will not retry that many times if there is a discrepancy.

winder: That doesn't account for the fact that every time a goroutines is waiting we aren't checking `160/<num-threads>` other accounts.

Also, only dots in the output doesn't mean that algod and indexer are synchronized. There's no indicator for matches that ignore that the round is mismatched.

winder: I thought that Indexer normally lags behind algod by several seconds. We would effectively be pausing the checks by that delta for every round. Hypothetically, lets say Indexer is lagging behind algod by 100ms and we're able to hit that time perfectly.

That would put our idle time at 100ms / 4.4s = 2.3%
In the run from above it took 17.25 hours. (17.25 * 1.023) = 17.25, so an extra ~24 minutes of downtime.

I think those numbers are probably pretty generous and the impact could be much larger.

tolikzinovyev: Got 79000 retries in my validation, so I guess it's pretty considerable. Can we still decrease retry timeout to 100ms and increase number of retries?

winder: Yeah, you can set those to whatever you want - there are CLI arguments

tolikzinovyev (approved):

PR #573, opened 2021-07-15
An integration test that tests database backward compatibility

tolikzinovyev:
Closes https://github.com/algorand/go-algorand-internal/issues/1340.


winder:
This doesn't actually test the backwards compatibility, just that there are no read errors. It's fine for a manual test but I think it would be a false sense of security if we were to check it in.

PR #575, opened 2021-07-17
Copy the byte array in bytePtr()

tolikzinovyev:
In the new accounting code this function triggered a bug due to calling this function in a loop in `api/converter_utils.go` `msigToTransactionMsig()`.

### DIFF
--- a/api/pointer_utils.go
+++ b/api/pointer_utils.go
@@ -59,7 +59,9 @@ func bytePtr(x []byte) *[]byte {
 	// Don't return if it's all zero.
 	for _, v := range x {
 		if v != 0 {
-			return &x
+			xx := make([]byte, len(x))
+			copy(xx, x)
+			return &xx
 		}
 	}
 
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -2360,9 +2360,10 @@ func baPtr(x []byte) *[]byte {
 	if allzero {
 		return nil
 	}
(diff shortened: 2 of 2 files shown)


api/pointer_utils.go
```diff
@@ -59,7 +59,9 @@ func bytePtr(x []byte) *[]byte {
 	// Don't return if it's all zero.
 	for _, v := range x {
 		if v != 0 {
-			return &x
+			xx := make([]byte, len(x))
+			copy(xx, x)
+			return &xx
```
winder, line 64: Could you also update `postgres.go : baPtr`?
```
func baPtr(x []byte) *[]byte {
	if x == nil || len(x) == 0 {
		return nil
	}
	allzero := true
	for _, b := range x {
		if b != 0 {
			allzero = false
			break
		}
	}
	if allzero {
		return nil
	}
	out := new([]byte)
	*out = x
	return out
}
```

tolikzinovyev: Is there a difference between `bytePtr()` and `baPtr()`? I can just delete the later.

winder: Just that they're private functions in different packages, I think there are a few cases where we have duplicate functions for the pointer utilities.

tolikzinovyev: It would be nice to have only one function in `util/`, but I changed `baPtr()` for now.

winder (approved):

PR #576, opened 2021-07-17
Move signature type logic to `idb`

tolikzinovyev:
## Summary

Create `SigType` type in `idb/` along with some helper functions.

## Test Plan

Verified that postgres doesn't complain about `SigType` in a query.

### DIFF
--- a/idb/sig_type.go
+++ b/idb/sig_type.go
@@ -1,19 +1,44 @@
 package idb
 
 import (
+	"fmt"
+	"strings"
+
 	sdk_types "github.com/algorand/go-algorand-sdk/types"
 )
 
+// SigType is signature type.
 type SigType string
 
+// Possible signature types.
 const (
 	Sig  SigType = "sig"
 	Msig SigType = "msig"
 	Lsig SigType = "lsig"
 )
 
+var sigTypeEnumMap = map[SigType]struct{}{
+	Sig:  {},
+	Msig: {},
+	Lsig: {},
+}
+
+func makeSigTypeEnumString() string {
+	keys := make([]string, 0, len(sigTypeEnumMap))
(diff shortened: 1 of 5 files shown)


idb/sig_type.go
```diff
@@ -0,0 +1,42 @@
...
+	if !stxn.Msig.Blank() {
+		return Msig
+	}
+	if !isZeroSig(stxn.Lsig.Sig) {
+		return Sig
+	}
+	if !stxn.Lsig.Msig.Blank() {
+		return Msig
+	}
+	return Lsig
```
winder, line 41: It was confusing to see this "default value" when written this way. Could you add an "unknown" type (or error) for the default, and put the 3 lsig options inside a statement like before?

idb/sig_type.go
```diff
@@ -0,0 +1,42 @@
...
+type SigType string
+
+const (
+	Sig  SigType = "sig"
+	Msig SigType = "msig"
+	Lsig SigType = "lsig"
+)
+
+func IsSigTypeValid(sigtype SigType) bool {
+	return (sigtype == Sig) || (sigtype == Msig) || (sigtype == Lsig)
```
winder, line 16: The lookup map is faster and the SigTypeEnumString means 1 fewer thing to remember when we add a new signature. Could you add it back?

winder (changes requested):
couple small comments otherwise looks good

idb/sig_type.go
tolikzinovyev: done

idb/sig_type.go
tolikzinovyev: done

winder (approved):
Thanks for the changes, looks good.

PR #577, opened 2021-07-17
non utf8 support for asset strings

winder:
## Summary

If asset strings contain non-printable characters they are stored in a byte array.
Return `-b64` asset strings in most endpoints.

## Test Plan

New unit test


types/types.go
```diff
@@ -906,15 +906,36 @@ const (
 	PaysetCommitMerkle
 )
 
+// AssetParamsWithExtra adds byte arrays to store non UTF8 asset strings.
+type AssetParamsWithExtra struct {
```
tsachiherman, line 910: you are using this only for the database persistence, right ? if so, wouldn't it be better to move this to the database layer ? ( i.e. idb/* )

idb/postgres/internal/encoding/encode.go
```diff
@@ -60,6 +96,11 @@ func convertSignedTxnWithAD(stxn types.SignedTxnWithAD) types.SignedTxnWithAD {
 
 // EncodeSignedTxnWithAD returns a json string where all byte arrays are base64 encoded.
 func EncodeSignedTxnWithAD(stxn types.SignedTxnWithAD) []byte {
+	ap := stxn.Txn.AssetParams
+	ap.AssetName = strings.ReplaceAll(ap.AssetName, "\000", "")
+	ap.UnitName = strings.ReplaceAll(ap.UnitName, "\000", "")
+	ap.URL = strings.ReplaceAll(ap.URL, "\000", "")
```
tsachiherman, line 102: This looks incorrect, but I'm probably not understanding the context : 
going from *imported* AssetParams to persisted AssetParams should base64 the result into the "XXXBytes" variant only when needed. Otherwise, it should remain in the original variable.

idb/postgres/postgres.go
```diff
@@ -1052,24 +1052,26 @@ ON CONFLICT (addr, assetid) DO UPDATE SET amount = account_asset.amount + EXCLUD
...
 							err = encoding.DecodeJSON(paramjson, &old)
 							if err != nil {
 								return fmt.Errorf("bad acgf json %d, %v", au.AssetID, err)
 							}
-							np := types.MergeAssetConfig(old, au.Config.Params)
-							outparams = encoding.EncodeJSON(np)
+							new := au.Config.Params
+							params = types.MergeAssetConfig(old, new)
 						}
+						params = encoding.ConvertAssetParams(params)
```
tsachiherman, line 1073: I think that the conversion here should be from `types.AssetParams` -> `types.AssetParamsWithExtra` and not from `AssetParamsWithExtra` -> `AssetParamsWithExtra`

idb/postgres/internal/encoding/encode.go
```diff
@@ -60,6 +96,11 @@ func convertSignedTxnWithAD(stxn types.SignedTxnWithAD) types.SignedTxnWithAD {
 
 // EncodeSignedTxnWithAD returns a json string where all byte arrays are base64 encoded.
 func EncodeSignedTxnWithAD(stxn types.SignedTxnWithAD) []byte {
+	ap := stxn.Txn.AssetParams
+	ap.AssetName = strings.ReplaceAll(ap.AssetName, "\000", "")
```
brianolson, line 100: leftover from a previous incarnation?

idb/postgres/internal/encoding/encode.go
```diff
@@ -18,6 +21,39 @@ func EncodeJSON(obj interface{}) []byte {
...
+}
+
+// ConvertAssetParams sanitizes asset param string fields before encoding to JSON bytes.
+func ConvertAssetParams(ap types.AssetParamsWithExtra) types.AssetParamsWithExtra {
+	ap.AssetName = printableUTF8OrEmpty(ap.AssetName)
+	ap.UnitName = printableUTF8OrEmpty(ap.UnitName)
+	ap.URL = printableUTF8OrEmpty(ap.URL)
+	// If the string is printable, don't store the encoded version.
+	// This is a nice optimization, and required for backwards compatibility.
+	if len(ap.AssetName) > 0 {
```
brianolson, line 45: this logic is backwards? If *NameBytes has become empty, then *Name must become empty, but *NameBytes can exist while *Name has been squelched.

idb/postgres/postgres.go
```diff
@@ -2117,9 +2120,12 @@ func (db *IndexerDb) yieldAccountsThread(req *getAccountsRequest) {
 						Total:         ap.Total,
 						Decimals:      uint64(ap.Decimals),
 						DefaultFrozen: boolPtr(ap.DefaultFrozen),
-						UnitName:      stringPtr(ap.UnitName),
-						Name:          stringPtr(ap.AssetName),
-						Url:           stringPtr(ap.URL),
+						UnitName:      stringPtr(ap.AssetParams.UnitName),
+						UnitNameB64:   baPtrOrDefault(ap.UnitNameBytes, []byte(ap.AssetParams.UnitName)),
```
brianolson, line 2124: This will result in []byte{} where we would rather have nil? If UnitName is "", `[]byte(ap.AssetParams.UnitName)` will be `[]byte{}`, but wouldn't we rather have nil?

idb/postgres/internal/encoding/encode.go
winder: This part is for the transaction's JSON format, these fields happen to be unused but I left it in like this in case Anya wants to use it.

types/types.go
brianolson: If it's db-only then it's postgres-only, because hopefully another db won't have the problem with storing data that Postgres has which is the cause of all this runaround.

tsachiherman: sorry - I did not mean it needs to move *directly* under `idb`, but rather under the `idb` root.

idb/postgres/postgres.go
```diff
@@ -2117,9 +2120,12 @@ func (db *IndexerDb) yieldAccountsThread(req *getAccountsRequest) {
 						Total:         ap.Total,
 						Decimals:      uint64(ap.Decimals),
 						DefaultFrozen: boolPtr(ap.DefaultFrozen),
-						UnitName:      stringPtr(ap.UnitName),
-						Name:          stringPtr(ap.AssetName),
-						Url:           stringPtr(ap.URL),
+						UnitName:      stringPtr(ap.AssetParams.UnitName),
```
tsachiherman, line 2123: It's not incorrect, but why do you use `ap.AssetParams.UnitName` instead of `ap.UnitName` ? i.e. it's an embedded struct

idb/postgres/internal/encoding/encode.go
tsachiherman: @brianolson I think that given that it's the "sanitize" path, it should be ok, but we should populate the `ap.AssetNameBytes` before the `ap.AssetName` assignment above.

types/types.go
winder: Thanks, I'll move it to postgres

idb/postgres/postgres.go
winder: Thanks, I missed this one. Previously I had `UnitName` defined twice, then I renamed the second one to `UnitNameBytes`.

idb/postgres/internal/encoding/encode.go
winder: Kind of. This is used for encoding the transaction JSON, so these fields are the "convenience" fields, we don't actually use them directly. Since there is no b64 equivalent in this data structure I did the minimal processing. This field would only be used by people poking around in the DB manually.

idb/postgres/internal/encoding/encode.go
winder: Fixed.

idb/postgres/postgres.go
winder: I shifted around the scope a bit to clean this up.

`AssetParamsWithExtra` is now part of the encoding layer.

The idb/api layers share the model object for sharing the string/byte fields (this is already shared in a couple places for other model objects).

idb/postgres/postgres.go
winder: Fixed

api/handlers.go
```diff
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"fmt"
+	"github.com/algorand/indexer/util"
```
tsachiherman, line 6: nit: move this line to the next section

api/handlers.go
```diff
@@ -492,34 +493,26 @@ func (si *ServerImplementation) fetchAssets(ctx context.Context, options idb.Ass
...
-				Reserve:       strPtr(row.Params.Reserve.String()),
-				Freeze:        strPtr(row.Params.Freeze.String()),
-				Manager:       strPtr(row.Params.Manager.String()),
-			},
+			Params:           row.Params,
+		}
+
+		// In case the DB layer filled the name with non-printable utf8
+		if asset.Params.Name != nil {
+			name := util.PrintableUTF8OrEmpty(*asset.Params.Name)
```
tsachiherman, line 506: while it's ok to add this, I would suggest you'll add a logging here to indicate that it happened, as it implies there is an issue elsewhere.

idb/postgres/internal/encoding/encode.go
```diff
@@ -2,8 +2,10 @@ package encoding
 
 import (
 	"encoding/base64"
-
+	sdk_types "github.com/algorand/go-algorand-sdk/types"
 	"github.com/algorand/go-codec/codec"
+	"github.com/algorand/indexer/util"
+	"strings"
```
tsachiherman, line 8: nit: move strings to the first group.

idb/postgres/internal/encoding/encode.go
```diff
@@ -18,6 +20,62 @@ func EncodeJSON(obj interface{}) []byte {
 	return buf
 }
 
+// AssetParamsWithExtra adds byte arrays to store non UTF8 asset strings.
+type AssetParamsWithExtra struct {
+	sdk_types.AssetParams
+	UnitNameBytes  []byte `codec:"un64"`
+	AssetNameBytes []byte `codec:"an64"`
+	URLBytes       []byte `codec:"au64"`
```
tsachiherman, line 28: add `omitempty`

idb/postgres/internal/encoding/encode.go
tsachiherman: Is it really possible to have this ? i.e. the AssetName should not contain these characters, so I'm kind of puzzled why we want to have this code.

api/handlers.go
winder: I was imagining another DB implemented the API. I don't think it's strictly an issue for the DB layer to ignore this presentation requirement. For Postgres, the tests verify that this case would never be triggered.

idb/postgres/postgres_integration_test.go
```diff
@@ -1103,3 +1103,210 @@ func TestOpenDbAgain(t *testing.T) {
...
+	if expected == "" {
+		require.Nil(t, actual)
+	} else {
+		require.NotNil(t, actual)
+		require.Equal(t, expected, *actual)
+	}
+}
+
+// TestEmbeddedNullString make sure we're able to import cheeky assets.
+func TestNonDisplayableUTF8(t *testing.T) {
```
tsachiherman, line 1117: Please add the B64 variation to this test ( or add another test that cover that functionality )

idb/postgres/postgres_migrations_integration_test.go
```diff
@@ -7,7 +7,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/algorand/go-algorand-sdk/encoding/json"
-
```
tsachiherman, line 10: any particular reason ? we usually group by repo, so the separation makes sense.

idb/postgres/postgres_integration_test.go
winder: This is testing the DB layer, so there is no B64 encoding to check here. In some cases there are the "B64" variables, which are actually byte arrays in the model, so I compare those with byte slice equality.

The tests in `test/common.sh` test the HTTP layer, but the test data is limited.

idb/postgres/internal/encoding/encode.go
winder: The strings need to be sanitized in the Asset table and the Transaction table.

Based on a discussion with Tokil I'm modifying the encoding to match some conventions used in the accounting rewrite and this special case is going away.

tolikzinovyev:
Looks good overall. Left some comments in the code. Could you also add tests in `idb/postgres/internal/encoding` similar to https://github.com/algorand/indexer/tree/tolik/acct-rewrite4/idb/postgres/internal/encoding?

idb/postgres/internal/encoding/encode.go
```diff
@@ -2,10 +2,11 @@ package encoding
 
 import (
 	"encoding/base64"
-
+	sdk_types "github.com/algorand/go-algorand-sdk/types"
```
tolikzinovyev, line 5: I think there should be an empty line to separate standard libraries from non-standard imports.

api/handlers.go
```diff
@@ -511,6 +512,9 @@ func (si *ServerImplementation) fetchAssets(ctx context.Context, options idb.Ass
 				Name:          strPtr(row.Params.AssetName),
```
tolikzinovyev, line 512: Don't we need to use `PrintableUTF8OrEmpty()` here?

idb/postgres/internal/encoding/encode.go
```diff
@@ -18,6 +21,38 @@ func EncodeJSON(obj interface{}) []byte {
...
 
+func convertAssetParams(params sdk_types.AssetParams) assetParams {
+	ret := assetParams{
+		AssetParams:      params,
+		AssetNameBytes: []byte(params.AssetName),
+		UnitNameBytes:  []byte(params.UnitName),
+		URLBytes:       []byte(params.URL),
+	}
+
+	ret.AssetName = util.PrintableUTF8OrEmpty(params.AssetName)
```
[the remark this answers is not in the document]
winder: I copied this function from the go-algorand implementation, it lets us skip 16 if statements which is pretty nice. Some discussion about the different versions here if you're curious: https://github.com/algorand/go-algorand/pull/2555#discussion_r670582068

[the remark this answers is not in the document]
tolikzinovyev: Yes, most of the times it is better. But in this function we set `AssetNameBytes` and then conditionally unset it, which took me some effort to understand. I suggest `IsPrintableUTF8()` is factored out of `PrintableUTF8OrEmpty()` and used here.

types/types.go
```diff
@@ -907,17 +907,17 @@ const (
...
-		out = new
-	} else if new == (sdk_types.AssetParams{}) {
+	if old.IsZero() {
 		out = new
+	} else if new.IsZero() {
+		out = old
 	} else {
 		out = old
-		if !old.Manager.IsZero() {
+		if !old.IsZero() {
```
tolikzinovyev, line 920: I'm confused about this change. Could you explain?

test/common.sh
```diff
@@ -709,4 +709,44 @@ function create_delete_tests() {
...
+      200 \
+      true \
+      '"decimals": 0' \
+      '"default-frozen": false' \
+      '"name": "bogocoin"' \
+      '"name-b64": "Ym9nb2NvaW4="' \
+      '"reserve": "EQJSQTOITX64CPGA5VRKBKLEJR57YXVNLTW5DKZMESIPTEOLRDNWQIJCGU"' \
+      '"total": 1000000000000' \
+      '"unit-name": "bogo"' \
+      '"unit-name-b64": "Ym9nbw=="'
```
tolikzinovyev, line 751: I think we also need tests where the strings are not utf-8-printable. I think a go test will suffice.

types/types.go
winder: Thanks, I shouldn't have removed `Manager`. Technically these checks are optional anyway because the transactions have already been validated.

test/common.sh
winder: Unfortunately, I don't have a good way to test the HTTP layer. You're right that this is missing the final check for removing name/unit-name/url if they contain non-utf8 characters.

tolikzinovyev: Mock `IndexerDb` and test the api functions? Ideally, we would test api functions with an implementation (implementations) of `IndexerDb`, but a bit of refactoring will need to be done beforehand.

winder: For now, I have run some tests manually.

I think it would be good to do some refactoring to the `api.Serve`, the pattern used by the block generator is pretty nice and allowed me to start/stop the server multiple times for the performance test runner.

tolikzinovyev:
LGTM. I will let @tsachiherman and @brianolson take a look as well.

idb/postgres/internal/encoding/types.go
```diff
@@ -0,0 +1,24 @@
+package encoding
+
+import (
+	sdk_types "github.com/algorand/go-algorand-sdk/types"
+
+	"github.com/algorand/indexer/types"
+)
+
+type assetParams struct {
```
brianolson, line 9: I don't like the split of types.go/encode.go/decode.go
I'd rather see one type declaration and its encode and decode in one file (OOP style). Easier to make sure encode and decode match that way. Easier to find the type being referred to.

brianolson (commented):
Gave it another read and generally looks good but one quibble over file layout.

idb/postgres/internal/encoding/types.go
winder: This split makes a lot more sense in the context of Tolik's rewrite, this pattern is used to wrap many more objects than I've done here.

I would probably opt for a single file, only because it would let you use simple text searches to find everything. But don't have a strong opinion either way. For now, I'd like to match his PR to make the merge slightly easier, maybe we can postpone this discussion for the rewrite PR?

Reference: https://github.com/algorand/indexer/blob/tolik/acct-rewrite4/idb/postgres/internal/encoding/types.go

brianolson: I won't block over it but I still don't like it. I still want type-encode-decode bundled together in one file. `types.go` makes the most sense when the only thing happening to them is via reflection (e.g. go-codec loading and storing serialization forms).

idb/postgres/internal/encoding/decode.go
```diff
@@ -2,7 +2,81 @@ package encoding
 
 import (
 	"github.com/algorand/go-algorand-sdk/encoding/json"
+	"github.com/algorand/indexer/types"
+
+	sdk_types "github.com/algorand/go-algorand-sdk/types"
```
tsachiherman, line 7: The grouping here is weird; it should have been like this, right ? 
```golang
import(
 	"github.com/algorand/go-algorand-sdk/encoding/json"
 	sdk_types "github.com/algorand/go-algorand-sdk/types" 	

 	"github.com/algorand/indexer/types"
)
```

tsachiherman (approved):
looks good to me.

PR #580, opened 2021-07-19
Run e2e test with nightly channel.

winder:
## Summary

When indexer is implementing a new algod feature, we need to use the nightly channel. This configuration uses a simple docker-compose environment to setup this environment.

## Test Plan

CI

### DIFF
--- a/misc/Dockerfile
+++ b/misc/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.15.4
+FROM golang:1.14.7
 
 # Misc dependencies
 ENV HOME /opt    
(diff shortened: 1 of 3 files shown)


bricerisingalgorand (approved):

misc/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM golang:1.15.4
```
tsachiherman, line 1: most current go version is 1.16.5; go-algorand uses 1.14.7.
Unless we have a good reason to use 1.15.4 here, I'd rather using the same version as in go-algorand for now.

winder: This is only used to build indexer, I'm not sure what version of go we use for the indexer releases.

tsachiherman: yes, I realize that.. using different versions for different products makes it harder to track "which of the known go bugs applies to which product".

PR #585, opened 2021-07-22
Delete reverse app deltas

tolikzinovyev:
## Summary

Reverse app deltas are currently written to the database, but are not used. This PR deprecates it and adds a non-blocking migration that deletes the reverse app deltas from the database.

## Test Plan

Migration integration tests.


brianolson:
Are we sure that we'll _never_ want rewind for apps?

tolikzinovyev:
Maybe. But supporting per-transaction rewinding will require much effort with the new accounting code: go-algorand only gives state changes per block, not per transaction.

winder (changes requested):
Blocked by requirements dependency. @yaovi-a please review.

yaovi-a:
@winder @brianolson If we don't delete the reverse delta data but don't update the logic until a later date, is the data useless?

tolikzinovyev:
@yaovi-a, after talking to @tsachiherman, I'm going to delete the migration and keep the data, just because there is no huge reason to do the migration that might impact Indexer performance.

brianolson:
We are currently generating data that's not used. We could either use it (implement app rewind) OR give up on app rewind and stop generating the reverse-app-deltas and delete them.

winder:
@yaovi-a if we don't update the logic now then the data will become out of date and require a major migration to recompute it later. I think the sort of migration we would require is significantly more expensive with the accounting rewrite than it is with the current custom accounting.

yaovi-a:
@winder, we would have to process every transaction because the rewrite would happen at the block level so excluding app txns wouldn't be possible, correct?

winder:
> @winder, we would have to process every transaction because the rewrite would happen at the block level so excluding app txns wouldn't be possible, correct?

because the rewrite uses algod's accounting which is optimized to work per-block.

yaovi-a:
Ok, unfortunately I don't think we have enough time to update the logic. The data will be useless for now, but we'll just have to revisit rewinding at a later date and decided if we want to update that logic.

yaovi-a (approved):

winder (approved):
LGTM

PR #587, opened 2021-07-23
Add more prometheus metrics.

winder:
## Summary

Added some new metrics to go along with `indexer_daemon_import_time_sec`:
-  block_upload_time_sec
-  imported_tx_per_block
-  imported_round

Also moved the location where metrics are collected, previously they did not represent exactly one round worth of data.

Updated the block generator import test runner to capture the new metrics.

New report output
```
cat cmd/block-generator/results/config.asset.close.report
test_duration_seconds:3600
test_duration_actual_seconds:3600.027172
transaction_asset_create_total:2134
transaction_asset_optin_total:1065726
transaction_asset_xfer_total:208325
transaction_asset_close_total:843815
early_average_import_time_sec:8.17
early_cumulative_import_time_sec:1054.50
early_average_imported_tx_per_block:5000.00
early_cumulative_imported_tx_per_block:645000
early_average_block_upload_time_sec:0.36
early_cumulative_block_upload_time_sec:47.54
early_imported_round:130
early_overall_transactions_per_second:611.67
early_uptime_seconds:3600.03
final_average_import_time_sec:8.42
final_cumulative_import_time_sec:3554.10
final_average_imported_tx_per_block:5000.00
final_cumulative_imported_tx_per_block:2110000
final_average_block_upload_time_sec:0.37
final_cumulative_block_upload_time_sec:155.84
final_imported_round:423
final_overall_transactions_per_second:593.68
final_uptime_seconds:3600.03
```

## Test Plan

Run the test runner and observed that the metrics look about right.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -186,8 +186,7 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	start := time.Now()
 	_, err := bih.imp.ImportDecodedBlock(block)
 	maybeFail(err, "ImportDecodedBlock %d", block.Block.Round)
-	dtUpload := time.Now().Sub(start)
-	metrics.BlockUploadTimeSeconds.Observe(dtUpload.Seconds())
+	metrics.BlockUploadTimeSeconds.Observe(time.Since(start).Seconds())
 	startRound, err := bih.db.GetNextRoundToAccount()
 	maybeFail(err, "failed to get next round to account")
 	// During normal operation StartRound and MaxRound will be the same round.
@@ -196,9 +195,9 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 		MaxRound:   uint64(block.Block.Round),
 	}
 	rounds, txns := importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
-	dt := time.Now().Sub(start)
+	dt := time.Since(start)
 
-	// Ignore calls that update >1 round (sneaky migration) and round 0
+	// Ignore calls that update >1 round (sneaky migration) and round 0 (which is empty)
 	if rounds <= 1 && startRound != 0 {
 		metrics.BlockImportTimeSeconds.Observe(dt.Seconds())
 		metrics.ImportedTxnsPerBlock.Observe(float64(txns))
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -188,7 +188,6 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	maybeFail(err, "ImportDecodedBlock %d", block.Block.Round)
 	dtUpload := time.Now().Sub(start)
 	metrics.BlockUploadTime.Observe(dtUpload.Seconds())
-	metrics.CumulativeBlockUploadTime.Add(dtUpload.Seconds())
 	startRound, err := bih.db.GetNextRoundToAccount()
 	maybeFail(err, "failed to get next round to account")
 	// During normal operation StartRound and MaxRound will be the same round.
@@ -202,10 +201,8 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	// Ignore calls that update >1 round (sneaky migration) and round 0
 	if rounds <= 1 && startRound != 0 {
 		metrics.BlockImportTimeSeconds.Observe(dt.Seconds())
-		metrics.CumulativeImportTime.Add(dt.Seconds())
 		metrics.ImportedTxnsPerBlock.Observe(float64(txns))
-		metrics.CumulativeTxns.Add(float64(txns))
-		metrics.CurrentRoundGauge.Set(float64(startRound))
+		metrics.ImportedRoundGauge.Set(float64(startRound))
 	}
 
 	logger.Infof("round r=%d (%d txn) imported in %s", block.Block.Round, len(block.Block.Payset), dt.String())
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
@@ -31,23 +30,12 @@ var (
 	tokenString      string
 )
 
-// importTimeHistogramSeconds is used to record the block import time metric.
-var importTimeHistogramSeconds = prometheus.NewSummary(
-	prometheus.SummaryOpts{
-		Subsystem: "indexer_daemon",
-		Name:      "import_time_sec",
-		Help:      "Block import and processing time in seconds.",
-	})
-
 var daemonCmd = &cobra.Command{
 	Use:   "daemon",
 	Short: "run indexer daemon",
 	Long:  "run indexer daemon. Serve api on HTTP.",
 	//Args:
 	Run: func(cmd *cobra.Command, args []string) {
-		// register metric with global prometheus metrics handler
-		prometheus.Register(importTimeHistogramSeconds)
-
 		var err error
 		config.BindFlags(cmd)
 		err = configureLogger()
@@ -206,7 +194,6 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	}
 	importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
 	dt := time.Now().Sub(start)
-	// record metric
-	importTimeHistogramSeconds.Observe(dt.Seconds())
+
 	logger.Infof("round r=%d (%d txn) imported in %s", block.Block.Round, len(block.Block.Payset), dt.String())
 }
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -16,6 +16,7 @@ import (
 	"github.com/algorand/indexer/idb"
 	"github.com/algorand/indexer/importer"
 	"github.com/algorand/indexer/types"
+	"github.com/algorand/indexer/util/metrics"
 )
 
 var (
@@ -185,6 +186,9 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	start := time.Now()
 	_, err := bih.imp.ImportDecodedBlock(block)
 	maybeFail(err, "ImportDecodedBlock %d", block.Block.Round)
+	dtUpload := time.Now().Sub(start)
+	metrics.BlockUploadTime.Observe(dtUpload.Seconds())
+	metrics.CumulativeBlockUploadTime.Add(dtUpload.Seconds())
 	startRound, err := bih.db.GetNextRoundToAccount()
 	maybeFail(err, "failed to get next round to account")
 	// During normal operation StartRound and MaxRound will be the same round.
@@ -192,8 +196,17 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 		StartRound: startRound,
 		MaxRound:   uint64(block.Block.Round),
 	}
-	importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
+	rounds, txns := importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
 	dt := time.Now().Sub(start)
+
+	// Ignore calls that update >1 round (sneaky migration) and round 0
+	if rounds <= 1 && startRound != 0 {
+		metrics.BlockImportTimeSeconds.Observe(dt.Seconds())
+		metrics.CumulativeImportTime.Add(dt.Seconds())
+		metrics.ImportedTxnsPerBlock.Observe(float64(txns))
+		metrics.CumulativeTxns.Add(float64(txns))
+		metrics.CurrentRoundGauge.Set(float64(startRound))
+	}
 
 	logger.Infof("round r=%d (%d txn) imported in %s", block.Block.Round, len(block.Block.Payset), dt.String())
 }
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -187,7 +187,7 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
 	_, err := bih.imp.ImportDecodedBlock(block)
 	maybeFail(err, "ImportDecodedBlock %d", block.Block.Round)
 	dtUpload := time.Now().Sub(start)
-	metrics.BlockUploadTime.Observe(dtUpload.Seconds())
+	metrics.BlockUploadTimeSeconds.Observe(dtUpload.Seconds())
 	startRound, err := bih.db.GetNextRoundToAccount()
 	maybeFail(err, "failed to get next round to account")
 	// During normal operation StartRound and MaxRound will be the same round.
--- a/cmd/block-generator/runner/run.go
+++ b/cmd/block-generator/runner/run.go
@@ -277,13 +277,13 @@ func (r *Args) runTest(indexerURL string, generatorURL string) error {
 
 	// Record a rate from one of the first data points.
 	if len(collector.Data) > 5 {
-		if err := recordDataToFile(collector.Data[2], "starting",  report); err != nil {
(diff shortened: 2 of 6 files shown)


util/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	prometheus.Register(BlockImportTimeHistogramSeconds)
+	prometheus.Register(CumulativeImportTimeCounter)
+	prometheus.Register(CumulativeTransactionsCounter)
+	prometheus.Register(ImportedTransactionsPerBlockHistogram)
+	prometheus.Register(CurrentRoundGauge)
+}
+
+const (
+
+	// ImportTimePerBlockHistogramName metric name.
```
tolikzinovyev, line 17: I think, if you put a comment right above `const (`, the linter will not require all these silly comments.

util/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+func RegisterPrometheusMetrics() {
+	prometheus.Register(BlockImportTimeHistogramSeconds)
+	prometheus.Register(CumulativeImportTimeCounter)
+	prometheus.Register(CumulativeTransactionsCounter)
+	prometheus.Register(ImportedTransactionsPerBlockHistogram)
+	prometheus.Register(CurrentRoundGauge)
+}
+
+const (
+
```
tolikzinovyev, line 16: please delete empty line

util/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	// ImportTimePerBlockHistogramName metric name.
+	ImportTimePerBlockHistogramName = "average_import_time_sec"
+	// ImportTimeCounterName metric name.
+	ImportTimeCounterName = "cumulative_import_time_sec"
+	// TransactionsPerBlockHistogramName metric name.
+	TransactionsPerBlockHistogramName = "average_imported_tx_per_block"
```
tolikzinovyev, line 22: nit: we use `txn` throughout indexer to denote blockchain transactions

util/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	// CurrentRoundGaugeName metric name.
+	CurrentRoundGaugeName = "current_round"
+)
+
+var (
+	// BlockImportTimeHistogramSeconds average block import duration in seconds.
+	BlockImportTimeHistogramSeconds = prometheus.NewSummary(
+		prometheus.SummaryOpts{
+			Subsystem: "indexer_daemon",
+			Name:      ImportTimePerBlockHistogramName,
```
tolikzinovyev, line 34: Would be nice to make the name of this variable consistent with the name of the prometheus variable. Right now there are so many variables + string constants, it is confusing.

importer/helper.go
```diff
@@ -259,21 +260,40 @@ func updateAccounting(db idb.IndexerDb, frozenCache map[uint64]bool, filter idb.
...
+			roundStart = now
+			metrics.BlockImportTimeHistogramSeconds.Observe(dt.Seconds())
+			metrics.CumulativeImportTimeCounter.Add(dt.Seconds())
+			metrics.ImportedTransactionsPerBlockHistogram.Observe(float64(txnForRound))
+			metrics.CumulativeTransactionsCounter.Add(float64(txnForRound))
+			metrics.CurrentRoundGauge.Set(float64(currentRound))
+		}
+	}
+	for txn := range txns {
+		fmt.Println("processing a transaction")
```
tolikzinovyev, line 292: Are you sure we want to print this for every transaction?

util/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	// TransactionsPerBlockHistogramName metric name.
+	TransactionsPerBlockHistogramName = "average_imported_tx_per_block"
+	// ImportedTransactionsCounterName metric name.
+	ImportedTransactionsCounterName = "cumulative_imported_tx"
+	// CurrentRoundGaugeName metric name.
+	CurrentRoundGaugeName = "current_round"
+)
+
+var (
+	// BlockImportTimeHistogramSeconds average block import duration in seconds.
```
tolikzinovyev, line 30: Similarly here, a single comment can be on top of `var (`.

importer/helper.go
```diff
@@ -259,21 +260,40 @@ func updateAccounting(db idb.IndexerDb, frozenCache map[uint64]bool, filter idb.
...
+			if txnForRound > 0 {
 				err := db.CommitRoundAccounting(act.RoundUpdates, currentRound, blockHeaderPtr)
 				maybeFail(err, l, "failed to commit round accounting")
 			}
 
+			// record metric
+			now := time.Now()
+			dt := now.Sub(roundStart)
+			roundStart = now
+			metrics.BlockImportTimeHistogramSeconds.Observe(dt.Seconds())
```
tolikzinovyev, line 284: Previously we measured import + accounting time, now we measure only accounting time. What is the reason for the change?

winder: Good point, maybe I should move it back.

My thinking was that it's possible for this function to process multiple rounds. I suppose for the tests I'm running (and healthy deployments) that wouldn't happen.

tolikzinovyev: Yes, in a normal scenario we import one block and account one block.

util/metrics/metrics.go
winder: I tried to shorten these a bit, and make the naming a bit more consistent

cmd/block-generator/runner/run.go
```diff
@@ -82,31 +83,133 @@ func (r *Args) run() error {
...
+}
+
+func getMetric(entry Entry, suffix string, t metricType) (float64, error) {
+	total := 0.0
 	sum := 0.0
 	count := 0.0
+	hasSum := false
+	hasCount := false
 
 	for _, metric := range entry.Data {
```
tolikzinovyev, line 173: Will `entry.Data` contain same key multiple times? Does this loop select the latest measurement then?

tolikzinovyev (commented):
The block generator code seems complicated, and I don't think I understand it fully. I think it can be simplified.

cmd/algorand-indexer/daemon.go
```diff
@@ -197,16 +186,27 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
...
 		MaxRound:   uint64(block.Block.Round),
 	}
-	importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
+	rounds, txns := importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
 	dt := time.Now().Sub(start)
-	// record metric
-	importTimeHistogramSeconds.Observe(dt.Seconds())
+
+	// Ignore calls that update >1 round (sneaky migration) and round 0
+	if rounds <= 1 && startRound != 0 {
```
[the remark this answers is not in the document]
winder: It never had any transactions, so it skews the results

util/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+		})
+
+	CumulativeImportTime = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: "indexer_daemon",
+			Name:      CumulativeImportTimeName,
+			Help:      "Total time in seconds spent uploading and processing blocks.",
+		})
+
+	BlockUploadTime = prometheus.NewSummary(
```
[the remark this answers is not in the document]
winder: Its a subset of the total import + processing time. So if it takes 8 seconds to it import a round with 20k transactions this might tell us that 1.5 seconds we're spent uploading the raw data

cmd/block-generator/runner/run.go
```diff
@@ -82,31 +83,133 @@ func (r *Args) run() error {
...
-			val := strings.Split(metric, " ")[1]
-			sum, err = strconv.ParseFloat(val, 64)
-		} else if strings.HasPrefix(metric, "indexer_daemon_import_time_sec_count") {
-			val := strings.Split(metric, " ")[1]
-			count, err = strconv.ParseFloat(val, 64)
-		}
-		if err != nil {
-			return fmt.Errorf("unable to parse metric '%s': %w", metric, err)
+
+		if strings.Contains(metric, suffix) {
```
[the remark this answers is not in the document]
winder: The Prometheus metrics parser is available as a library. Or Setting up a Prometheus instance and running a proper rate query would also be a more robust solution.

That seemed like overkill for what we're doing so far.

cmd/block-generator/runner/run.go
winder: There is an array of entry objects and 10 samples are taken while the test runs.

cmd/block-generator/runner/run.go
```diff
@@ -127,11 +230,12 @@ func (r *Args) runTest(indexerURL string, generatorURL string) error {
...
 	for time.Since(start) < r.RunDuration {
 		time.Sleep(r.RunDuration / 10)
-		if err := collector.Collect("indexer_daemon_import_time_sec"); err != nil {
+
+		if err := collector.Collect(metrics.AllMetricNames...); err != nil {
 			return fmt.Errorf("problem collecting metrics: %w", err)
 		}
 	}
-	if err := collector.Collect("indexer_daemon_import_time_sec"); err != nil {
+	if err := collector.Collect(metrics.AllMetricNames...); err != nil {
```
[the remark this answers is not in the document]
winder: I'm searching for a substring so a map wouldn't work. The metric prefix should be configurable, we just haven't done it yet. Also, different metrics have different suffixes added automatically.

I don't think we need to do better than this. If we wanted to do better we'd probably replace the whole thing instead of optimizing this code.

cmd/block-generator/runner/run.go
```diff
@@ -82,31 +83,133 @@ func (r *Args) run() error {
...
+		return err
+	}
+
+	importTimeS, err := getMetric(entry, metrics.CumulativeImportTimeName, intTotal)
+	if err != nil {
+		return err
+	}
+	tps := totalTxn / importTimeS
+	tpsKey := "overal_transactions_per_second"
+	msg := fmt.Sprintf("%s_%s:%.2f\n", prefix, tpsKey, tps)
```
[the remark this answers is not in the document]
winder: No, that helper function also looks up the metric.

util/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	prometheus.Register(CumulativeTxns)
+	prometheus.Register(ImportedTxnsPerBlock)
+	prometheus.Register(CurrentRoundGauge)
+	prometheus.Register(CumulativeBlockUploadTime)
+	prometheus.Register(BlockUploadTime)
+}
+
+// Prometheus metric names broken out for reuse.
+const (
+	BlockImportTimeName           = "average_import_time_sec"
```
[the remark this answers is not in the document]
winder: The histogram comes from taking multiple samples, but whats actually happening with "prometheus.NewSummary" type is an average.

util/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	BlockUploadTimeName,
+	CumulativeBlockUploadTimeName,
+	ImportedTxnsPerBlockName,
+	CumulativeTxnsName,
+	CurrentRoundGaugeName,
+}
+
+// Initialize the prometheus objects.
+var (
+	BlockImportTimeSeconds = prometheus.NewSummary(
```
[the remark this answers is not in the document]
winder: I'll add seconds to the other time.

cmd/block-generator/runner/run.go
```diff
@@ -82,33 +83,140 @@ func (r *Args) run() error {
...
+					return sum, nil
+				}
+				if hasTotal {
+					return total, nil
+				}
+			}
+		}
+	}
+
+	return 0.0, fmt.Errorf("metric not found: %s", suffix)
```
algonautshant, line 217: This may not be necessarily correct. 
Does it matter if the metric is there, but not in the form of rateMetric. And because the request is rateMetric, metric not found is returned. Does this mean "rate metric not found"?

algonautshant (approved):
Looks great. 
Just a few comments.

cmd/algorand-indexer/daemon.go
```diff
@@ -197,16 +186,23 @@ func (bih *blockImporterHandler) HandleBlock(block *types.EncodedBlockCert) {
...
 	}
-	importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
-	dt := time.Now().Sub(start)
-	// record metric
-	importTimeHistogramSeconds.Observe(dt.Seconds())
+	rounds, txns := importer.UpdateAccounting(bih.db, bih.cache, filter, logger)
+	dt := time.Since(start)
+
+	// Ignore calls that update >1 round (sneaky migration) and round 0 (which is empty)
+	if rounds <= 1 && startRound != 0 {
```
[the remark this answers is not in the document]
winder: `rounds` is the number of rounds processed in a batch, under normal operation it should be 0 if the round at no transactions, or 1.

Sneaky migration is a way that we can recompute the accounting without reloading transaction/block header data. It's kind of a weird thing, and users shouldn't be doing it, so I felt ok disabling metrics for that case.

cmd/block-generator/runner/run.go
```diff
@@ -82,33 +83,140 @@ func (r *Args) run() error {
...
-			return fmt.Errorf("unable to parse metric '%s': %w", metric, err)
+const (
+	rate metricType = iota
+	intTotal
+	floatTotal
+)
+
+// Helper to record metrics. Supports rates (sum/count) and counters.
+func recordDataToFile(start time.Time, entry Entry, prefix string, out *os.File) error {
+	num := 0
```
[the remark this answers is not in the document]
winder: Oops, it isn't. I was using this earlier but not anymore.

cmd/block-generator/runner/run.go
```diff
@@ -82,33 +83,140 @@ func (r *Args) run() error {
...
+	var writeErrors strings.Builder
+	var writeErr error
+	record := func(prefix2, name string, t metricType) {
+		key := fmt.Sprintf("%s%s_%s", prefix, prefix2, name)
+		if err := recordMetricToFile(entry, key, name, t, out); err != nil {
+			writeErr = err
+			if writeErrors.Len() > 0 {
+				writeErrors.WriteString(", ")
+			}
+			writeErrors.WriteString(name)
```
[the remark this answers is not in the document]
winder: honestly, the main reason was to avoid having to check return values after every call.

I do kinda like accumulating errors like this, if two of the metrics were missing it would report both of them instead of just the first.

cmd/block-generator/runner/run.go
winder: I think it's still correct, but you're right about there being some nuance to it... if the metric wasn't a "summary" then the rate fields wouldn't be there and the error could be more precise. On the other hand if the metric isn't there, it would also hit this  error.

cmd/block-generator/runner/run.go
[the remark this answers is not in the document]
winder: I dunno, I tried changing. Do you think this better?
```
  record := func(prefix2, name string, t metricType) {
    key := fmt.Sprintf("%s%s_%s", prefix, prefix2, name)
    return recordMetricToFile(entry, key, name, t, out)
  }

  if err := record("_average", metrics.BlockImportTimeName, rate); err != nil {    
    return err     
  }    
  if err := record("_cumulative", metrics.BlockImportTimeName, floatTotal); err != nil {    
    return err                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                            
  if err := record("_average", metrics.ImportedTxnsPerBlockName, rate); err != nil {                                                                                                                                                                           
    return err                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                            
  if err := record("_cumulative", metrics.ImportedTxnsPerBlockName, intTotal); err != nil {                                                                                                                                                                    
    return err                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                            
  if err := record("_average", metrics.BlockUploadTimeName, rate); err != nil {                                                                                                                                                                                
    return err                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                            
  if err := record("_cumulative", metrics.BlockUploadTimeName, floatTotal); err != nil {                                                                                                                                                                       
    return err                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                            
  if err := record("", metrics.ImportedRoundGaugeName, intTotal); err != nil {                                                                                                                                                                                 
    return err                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                            
```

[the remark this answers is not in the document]
algonautshant: haha maybe not. This looks more like go :-P 
It's your call. I was curious if there was another reason for it. You can leave it as is if you like.

PR #588, opened 2021-07-24
Update README.md

ipaleka:
## Summary

The deleted paragraph is obsolete as today's difference is 495GB:609GB.

https://forum.algorand.org/t/does-anybody-have-a-link-to-a-pg-dump-of-the-indexer-database/3843


winder:
Rather than removing the comment, could you update it to say the current size?

ipaleka:
Ok, I'm going to do that.

PR #591, opened 2021-07-26
Add missing validation for b64 fields.

winder:
## Summary

Update the validator for new fields.


tolikzinovyev (approved):
LGTM. No change needed for the dynamic processor?

winder:
> LGTM. No change needed for the dynamic processor?

No, that one compares everything dynamically so it would just work.

PR #593, opened 2021-07-26
Add support for running integration tests with an external DB.

winder:
## Summary

Didn't want to lose this functionality from the recent non-utf8 work. We should use it on CI to avoid the gnomock overhead.

### DIFF
--- a/idb/postgres/postgres_integration_common_test.go
+++ b/idb/postgres/postgres_integration_common_test.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"database/sql"
+	"flag"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


idb/postgres/postgres_integration_common_test.go
```diff
@@ -17,8 +18,24 @@ import (
 	"github.com/algorand/indexer/util/test"
 )
 
+var testpg = flag.String("test-pg", "", "postgres connection string")
+
 // setupPostgres starts a gnomock postgres DB then returns the connection string and a shutdown function.
 func setupPostgres(t *testing.T) (*sql.DB, string, func()) {
+	if testpg != nil && *testpg != "" {
```
tolikzinovyev, line 25: One of the checks should be sufficient.

idb/postgres/postgres_integration_common_test.go
```diff
@@ -17,8 +18,24 @@ import (
...
+		// TODO: Drop schema?
```
tolikzinovyev, line 26: Without dropping the schema, this will not work for multiple tests, right?

tolikzinovyev (commented):
Good addition! The tests in my accounting branch take really long time now.

idb/postgres/postgres_integration_common_test.go
winder: Yes exactly. This comment is my only contribution to Brian's work here.

brianolson (approved):

idb/postgres/postgres_integration_common_test.go
tolikzinovyev: I'm not sure what you are referring to. In any case, we need to drop the schema. Otherwise, the tests might not do what they are supposed to do!

brianolson: Setup is an external problem. I'd presume that whatever is handing us a link to a database just ran `createdb` or otherwise ensured the db was empty and ready to use.

tolikzinovyev: That's a reasonable assumption / requirement, but not sufficient if you are executing multiple tests.

PR #595, opened 2021-07-27
Allow overriding OS_TYPE and ARCH. This will allow signing locally on e.g. mac

onetechnical:
## Summary

To execute `make sign` you need to be able to issue the signing command. There is a script for this, but it presumes that the architecture you're signing for is the same as what you're running.

This is problematic if you are on mac, as OS_TYPE will register as darwin. To override this, we allow you to override OS_TYPE, e.g. set it to linux. We will also allow overriding ARCH, since this may be set to arm64 on newer macs.

## Test Plan

Verify signing works with these new changes, e.g.:

`ARCH=amd64 OS_TYPE=linux make sign`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 SRCPATH		:= $(shell pwd)
 VERSION		:= $(shell $(SRCPATH)/mule/scripts/compute_build_number.sh)
 OS_TYPE		?= $(shell $(SRCPATH)/mule/scripts/ostype.sh)
-ARCH			:= $(shell $(SRCPATH)/mule/scripts/archtype.sh)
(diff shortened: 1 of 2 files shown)


egieseke (approved):
Makes sense.

algobarb:
Since this is for signing, would this change only affect our team? Also, would we get a help message or some doc so we know we can pass in these new variables as described in the test plan `ARCH=amd64 OS_TYPE=linux make sign` ?

onetechnical:
> Since this is for signing, would this change only affect our team? Also, would we get a help message or some doc so we know we can pass in these new variables as described in the test plan `ARCH=amd64 OS_TYPE=linux make sign` ?

Correct, this is internal only. I am updating the release doc right now.

algobarb (approved):

PR #600, opened 2021-07-30
Fix OpenAPI tooling.

winder:
## Summary

Due to incompatible dependencies the OpenAPI codegen makefile wasn't
working properly.
## Test Plan

Manually tested the script.


tolikzinovyev:
It still doesn't work for me: either from outside any go repo or from a temporary repo.

tolikzinovyev (approved):

PR #601, opened 2021-08-02
Import go-algorand

tolikzinovyev:
## Summary

Changes in `Makefile` to import go-algorand: first tell go to fetch the repository, then run `make` in it. I'm not planning to merge it right away, but would like early feedback.

Closes https://github.com/algorand/go-algorand-internal/issues/1332.

## Test Plan

Tested manually.


Makefile
```diff
@@ -15,8 +18,19 @@ GOLDFLAGS += -X github.com/algorand/indexer/version.ReleaseVersion=$(shell cat .
...
+go-algorand-dl:
+	@if test -d ${GO_ALGORAND_CACHE_PATH}; then \
+		echo "go-algorand is already downloaded"; \
+	else \
+		echo "downloading go-algorand"; \
+		go get -d ${GO_ALGORAND_REPO}; \
+	fi
+
+go-algorand: go-algorand-dl
+	cd ${GO_ALGORAND_CACHE_PATH} && chmod +x `find -name *.sh` && chmod -R +w . && make
```
winder, line 33: Is this because `go get` is setting the wrong permissions?

Maybe we can avoid this by using a nested repository, or download with `git clone`?

winder (commented):
Looks pretty good

Makefile
tolikzinovyev: Yes, `go get` unsets those flags, probably for safety reasons. I will see what a nested repo will look like.

tsachiherman, line 33: I don't believe you really need to `make`. It would be enough for you to do what's being done in here :
`/Users/tsachi/go/src/github.com/algorand/go-algorand/scripts/travis/before_build.sh`, which might save you from having the need to do any attribute changes.

tolikzinovyev: It works, but I'm concerned that because it's a script for travis, it might do something arbitrary at some point.

PR #617, opened 2021-08-17
Rename helpers in `account_testutil.go`

tolikzinovyev:
Makes the naming more correct/consistent.

### DIFF
--- a/util/test/account_testutil.go
+++ b/util/test/account_testutil.go
@@ -51,10 +51,10 @@ var (
 )
 
 func init() {
-	OpenMainStxn, OpenMain = MakePayTxnRowOrPanic(Round, 1000, 10234, 0, 111, 1111, 0, AccountC,
+	OpenMainStxn, OpenMain = MakePaymentTxn(Round, 1000, 10234, 0, 111, 1111, 0, AccountC,
 		AccountA, basics.Address{}, basics.Address{})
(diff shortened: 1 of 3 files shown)


util/test/account_testutil.go
```diff
@@ -71,8 +71,8 @@ func DecodeAddressOrPanic(addr string) basics.Address {
 	return result
 }
 
-// MakeAssetConfigOrPanic is a helper to ensure test asset config are initialized.
-func MakeAssetConfigOrPanic(round, configid, assetid, total, decimals uint64, defaultFrozen bool, unitName, assetName, url string, addr basics.Address) (*transactions.SignedTxnWithAD, *idb.TxnRow) {
+// MakeCreateAssetTxn is a helper to ensure test asset config are initialized.
+func MakeCreateAssetTxn(round, configid, assetid, total, decimals uint64, defaultFrozen bool, unitName, assetName, url string, addr basics.Address) (*transactions.SignedTxnWithAD, *idb.TxnRow) {
```
winder, line 75: This is AssetConfig not create. Create is a special case of asset config when the asset ID is 0.

winder (changes requested):
Just one comment, otherwise the new names look good

util/test/account_testutil.go
tolikzinovyev: thanks

winder (approved):

PR #623, opened 2021-08-24
Generator Fixes

winder:
## Summary

The go-algorand Eval function does more error checking than Indexer was doing. This fixes some of the problems and adds options for writing Indexer output to a log file to make it easier to debug.

I haven't fixed all of the problems, for example, the block-generator attempts to use assets in the same round where they are created.

## Test Plan

Existing unit tests.

### DIFF
--- a/cmd/block-generator/generator/generate.go
+++ b/cmd/block-generator/generator/generate.go
@@ -113,6 +113,7 @@ func MakeGenerator(config GenerationConfig) (Generator, error) {
 
 	gen.feeSink[31] = 1
 	gen.rewardsPool[31] = 2
+	gen.genesisHash[31] = 3
 
 	gen.initializeAccounting()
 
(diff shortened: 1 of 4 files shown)


cmd/block-generator/generator/generate.go
```diff
@@ -390,7 +391,6 @@ func convert(txn transactions.Transaction, ad transactions.ApplyData) transactio
 	stxnib := transactions.SignedTxnInBlock{
 		SignedTxnWithAD: withAd,
```
tolikzinovyev, line 392: we should probably use `BlockHeader.EncodeSignedTxn()`

tolikzinovyev (commented):
Thanks for fixing it! Only one comment.

cmd/block-generator/generator/generate.go
winder: Nice, thanks for the tip! Still a little work to do for Assets but at least payment txns are working.

tolikzinovyev (approved):

PR #626, opened 2021-08-31
Move `ledger_for_evaluator` and `writer` tests to their packages

tolikzinovyev:
Closes https://github.com/algorand/go-algorand-internal/issues/1540.


idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
```diff
@@ -1,7 +1,8 @@
-package postgres
+package ledgerforevaluator_test
```
brianolson, line 1: `package ledger_for_evaluator` to match the directory name?

idb/postgres/internal/test_util/test_util.go
```diff
@@ -0,0 +1,56 @@
+package testutil
```
brianolson, line 1: package name should match directory

idb/postgres/internal/util/util.go
```diff
@@ -0,0 +1,69 @@
...
+		// If not serialization error.
+		var pqerr *pq.Error
+		if !errors.As(err, &pqerr) || (pqerr.Code != "40001") {
+			if (count > 0) && (log != nil) {
+				log.Printf("transaction was retried %d times", count)
+			}
+			return err
+		}
+
+		count++
```
brianolson, line 40: need to have a limit on retry count

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
tolikzinovyev: Go (linter) complains about underscores in package names. `_test` ensures that testing is black-box: that we can't e.g. access private struct fields from the tested code.

idb/postgres/internal/test_util/test_util.go
tolikzinovyev: same

idb/postgres/internal/util/util.go
tolikzinovyev: Maybe. But this is out of scope.

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
brianolson: Then maybe it can't live in a directory with underscores in the directory name.

brianolson (changes requested):

idb/postgres/internal/util/util.go
tolikzinovyev: I would like to avoid making assumptions about timing and this is out of scope.

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
tolikzinovyev: Just because go has a strange convention for package names doesn't mean we should propagate it to directory names. Removing underscores from the package name is a good compromise to keep using the linter.

idb/postgres/internal/util/util.go
tsachiherman: I agree with @brianolson that we definitely want to have this. however, this should not block merging this code in, since it's not any worse than prior to this change. @tolikzinovyev - could you please open a ticket for adding this technical debt ?

tsachiherman (approved):
I have added few small requests of my own.. but it looks reasonable to merge this one in.

idb/postgres/internal/util/util.go
tolikzinovyev: https://github.com/algorand/indexer/issues/635

idb/postgres/internal/util/util.go
```diff
@@ -0,0 +1,69 @@
...
+		tx, err := db.BeginTx(ctx, &opts)
+		if err != nil {
+			return err
+		}
+
+		err = f(ctx, tx)
+
+		// If not serialization error.
+		var pqerr *pq.Error
+		if !errors.As(err, &pqerr) || (pqerr.Code != "40001") {
```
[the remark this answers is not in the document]
tolikzinovyev: I rebased the PR. The new postgres driver has a constant for `40001`.

idb/postgres/internal/test_util/test_util.go
tolikzinovyev: done

tsachiherman (approved):

PR #627, opened 2021-09-02
Shutdown daemon gracefully

tolikzinovyev:
## Summary

Shutdown the daemon gracefully to allow the post-run hook in `main.go` to write the CPU profile to disk. Also, send a more polite signal to indexer from block generator.

## Test Plan

Tested manually.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -69,6 +69,9 @@ func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError
 	generated.RegisterHandlers(e, &api, middleware...)
 	common.RegisterHandlers(e, &api)
 
+	if ctx == nil {
+		ctx = context.Background()
+	}
 	getctx := func(l net.Listener) context.Context {
 		return ctx
 	}
--- a/api/server.go
+++ b/api/server.go
@@ -69,9 +69,6 @@ func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError
 	generated.RegisterHandlers(e, &api, middleware...)
 	common.RegisterHandlers(e, &api)
 
-	if ctx == nil {
-		ctx = context.Background()
-	}
 	getctx := func(l net.Listener) context.Context {
 		return ctx
 	}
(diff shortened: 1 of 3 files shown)


api/server.go
```diff
@@ -69,9 +69,6 @@ func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError
 	generated.RegisterHandlers(e, &api, middleware...)
 	common.RegisterHandlers(e, &api)
 
-	if ctx == nil {
-		ctx = context.Background()
-	}
 	getctx := func(l net.Listener) context.Context {
 		return ctx
 	}
```
winder, line 74: The nil check is required to avoid violating BaseContext constraints (and a panic below), from the docs:
```
	// If BaseContext is nil, the default is context.Background().
	// If non-nil, it must return a non-nil context.
```

cmd/algorand-indexer/daemon.go
```diff
@@ -51,6 +53,16 @@ var daemonCmd = &cobra.Command{
...
+		{
+			cancelCh := make(chan os.Signal, 1)
+			signal.Notify(cancelCh, syscall.SIGTERM, syscall.SIGINT)
+			go func() {
+				<-cancelCh
+				logger.Println("Stopping Indexer.")
+				cf()
+			}()
+		}
+
```
winder, line 65: I don't think the scope is needed here
```suggestion
		cancelCh := make(chan os.Signal, 1)
		signal.Notify(cancelCh, syscall.SIGTERM, syscall.SIGINT)
		go func() {
			<-cancelCh
			logger.Println("Stopping Indexer.")
			cf()
		}()

```

winder (changes requested):
Looks good, just a couple nits

api/server.go
tolikzinovyev: We can just assume that `Serve()` accepts non-nil context? It's a common assumption, I've never seen `nil` passed as a context parameter.

cmd/algorand-indexer/daemon.go
tolikzinovyev: I wanted to hide `cancelCh`.

api/server.go
winder: No, I don't think we should panic when it's entirely avoidable. Especially when it's already handled correctly.

tolikzinovyev: Ok, I added it back.

winder (approved):

PR #628, opened 2021-09-02
Add a cpu profile flag in block generator

tolikzinovyev:


cmd/block-generator/runner/run.go
```diff
@@ -335,7 +336,8 @@ func startIndexer(logfile string, loglevel string, indexerBinary string, algodNe
 		"--postgres", postgresConnectionString,
 		"--server", indexerNet,
 		"--logfile", logfile,
-		"--loglevel", loglevel)
+		"--loglevel", loglevel,
+		"--cpuprofile", cpuprofile)
```
winder, line 340: Is it disabled when this is an empty string?

tolikzinovyev: Yes. I think what happens is, an empty string is treated as one of the command line arguments.

winder (approved):

PR #630, opened 2021-09-02
Generator asset fix

winder:
## Summary

Add newly created assets to a pending state so that they aren't used until the next round.

## Test Plan

run_generator.sh no longer fails


cmd/block-generator/generator/generate.go
```diff
@@ -323,39 +337,43 @@ func (g *generator) WriteBlock(output io.Writer, round uint64) {
...
+		}
+		transactions = append(transactions, stib)
 	}
 
-	g.txnCounter += g.config.TxnPerBlock
-
-	block := bookkeeping.Block{
-		BlockHeader: header,
-		Payset:      transactions,
+	if numTxnForBlock != uint64(len(transactions)) {
```
tolikzinovyev, line 354: I think it's evident from the code that this should never happen.

tolikzinovyev (commented):
Thanks for fixing it! One small comment and please fix CI.

winder:
This is working now and I've tested it with all scenarios.

tolikzinovyev (approved):

PR #636, opened 2021-09-07
Use context properly with the new postgres driver

tolikzinovyev:
## Summary

With the old postgres driver, when we interacted with a database transaction created with a context that is now expired, the transaction rolled back automatically, and the user received an error right away. With the new driver, the context is only used when creating a transaction and is immediately forgotten by the driver. This PR adds the missing context expiration checks previously performed by the postgres driver.

## Test Plan

We currently don't have any tests that check behavior when the context is expired.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -706,6 +706,11 @@ func (db *IndexerDb) txnsWithNext(ctx context.Context, tx *sql.Tx, tf idb.Transa
 		return
 	}
 	tf.Limit -= uint64(count)
+	select {
+	case <-ctx.Done():
+		return
+	default:
+	}
 	tf.Round = origRound
 	if tf.Address != nil {
 		// (round,intra) descending into the past
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -109,13 +109,13 @@ type IndexerDb struct {
 }
 
 // txWithRetry is a helper function that retries the function `f` in case the database
-// transaction in it fails due to a serialization error. `f` is provided context `ctx`
-// and a transaction created using this context and `opts`. `f` takes ownership of the
+// transaction in it fails due to a serialization error. `f` is provided
+// a transaction created using `opts`. `f` takes ownership of the
 // transaction and must either call sql.Tx.Rollback() or sql.Tx.Commit(). In the second
 // case, `f` must return an error which contains the error returned by sql.Tx.Commit().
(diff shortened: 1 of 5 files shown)


tolikzinovyev:
Will, would be great if you could also go over `postgres.go` to check if there are any missing context expiration checks.

winder:
Is there any documentation that explains this difference in behavior?

tolikzinovyev:
I just looked into the source code more carefully, couple things:
1. The new postgres driver actually uses context when committing or rolling a transaction back, but only if used with the `database/sql` interface.
2. `database/sql` controls how the context is used when querying the database; turns out when using `Query()` or `Exec()` without a context, the context passed into `BeginTx()` is not used, so we have this wrong even now.
3. The new driver uses the context passed into `BeginTx()` only for sending the transaction command to the database. https://pkg.go.dev/github.com/jackc/pgx/v4#Conn.BeginTx

So, actually, this PR was necessary even for the old driver.

idb/postgres/postgres.go
```diff
@@ -706,11 +706,6 @@ func (db *IndexerDb) txnsWithNext(ctx context.Context, tx *sql.Tx, tf idb.Transa
 		return
 	}
 	tf.Limit -= uint64(count)
-	select {
```
brianolson, line 709: no reason to drop this? We _could_ rely on the QueryContext() to fail out if ctx is Done, but why not check between two biggish chunks of work?

brianolson (commented):
good catch on context usage. I guess I'd thought that BeginTx with context was enough to establish it for the rest of the life of the tx. :-/

idb/postgres/postgres.go
tolikzinovyev: Right before this we pass `ctx` to `yieldTxnsThreadSimple()`, and I thought it returned an error if the context expires. Btw, is there a reason why we do not notify the user if the context expires? Seems like, he will just get incomplete data and be unaware of it.

tolikzinovyev:
Yes, I thought the same thing. The documentation was not helpful...

idb/postgres/postgres.go
tolikzinovyev: I returned this code, but I think we later need to improve context handling.

brianolson: I expect the most likely cause of context expiration is that HTTP client closes their connection; so the database can stop doing work when there's no client to return data to.

winder (approved):
LGTM, thanks for the explanation

PR #639, opened 2021-09-08
Add eval time metric to prometheus

winder:
## Summary

Add a new metric to measure how much time is spent in the Eval function. This should allow us to start breaking down the performance so that we can tell how much time is being spent at different stages of the block import process.

## Test Plan

New script to call the block-generator runner tool.


tolikzinovyev (approved):
Looks good! I think it would also be useful to measure the Writer as well.

PR #641, opened 2021-09-09
block-generator runner with validation

winder:
## Summary

Add an option to the import test runner to run validation after the test.

## Test Plan

Run the script with different scenarios.


cmd/block-generator/generator/generate.go
```diff
@@ -637,3 +650,82 @@ func (g *generator) generateAssetTxn(round uint64, intra uint64) (transactions.S
...
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal address: %w", err)
+	}
+
+	idx := accountToIndex(addr)
+
+	// Asset Holdings
+	assets := make([]generated.AssetHolding, 0)
+	createdAssets := make([]generated.Asset, 0)
+	for _, a := range g.assets {
```
winder, line 665: The change to make asset holdings a pointer, and adding a reference instead of a bool was done to make this loop faster. Previously everything was optimized for creating / deleting random assets/holdings with no concern to lookups.

cmd/block-generator/generator/server.go
```diff
@@ -31,7 +31,20 @@ func initializeConfigFile(configFile string) (config GenerationConfig, err error
...
-func MakeServer(configFile string, addr string) *http.Server {
+func MakeServer(configFile string, addr string) (*http.Server, Generator) {
+	noOp := func(next http.Handler) http.Handler {
+		return next
+	}
+	return MakeServerWithMiddleware(configFile, addr, noOp)
+}
+
+// BlocksMiddleware is a middleware for the blocks endpoint.
+type BlocksMiddleware func(next http.Handler) http.Handler
```
winder, line 42: I needed a way to stop the `/v2/blocks` endpoint while leaving the `/v2/accounts` endpoint active. Maybe there is a cleaner way, but it's nice that this middleware approach means the test runner doesn't need to know anything about the server.

cmd/block-generator/runner.go
```diff
@@ -32,6 +32,8 @@ func init() {
 	runnerCmd.Flags().StringVarP(&runnerArgs.ReportDirectory, "report-directory", "r", "", "Location to place test reports.")
 	runnerCmd.Flags().StringVarP(&runnerArgs.LogLevel, "log-level", "l", "error", "LogLevel to use when starting Indexer. [error, warn, info, debug, trace]")
 	runnerCmd.Flags().StringVarP(&runnerArgs.CPUProfilePath, "cpuprofile", "", "", "Path where Indexer writes its CPU profile.")
+	runnerCmd.Flags().BoolVarP(&runnerArgs.ResetReportDir, "reset", "", false, "If set any existing report directory will be deleted before running tests.")
```
winder, line 35: deleting this directory between tests was driving me crazy!

cmd/validator/core/struct_processor.go
```diff
@@ -100,7 +100,7 @@ func equals(indexer, algod generated.Account) (differences []string) {
 	indexerAssets := assetLookupMap(indexer.Assets)
 	indexerAssetCount := len(indexerAssets)
 	// There should be the same number of undeleted indexer assets as algod assets
-	if algod.Assets == nil && indexerAssetCount > 0 || indexerAssetCount != len(*algod.Assets) {
```
winder, line 103: this needs parens

cmd/block-generator/generator/server.go
```diff
@@ -75,12 +96,27 @@ func getBlockHandler(gen Generator) func(w http.ResponseWriter, r *http.Request)
...
+		writeResponse(w, gen.WriteBlock(w, round))
+	}
+}
+
+func getAccountHandler(gen Generator) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// The generator doesn't actually care about the block...
+		account, err := parseAccount(r.URL.Path)
+		if err != nil {
+			fmt.Fprintf(w, err.Error())
```
tolikzinovyev, line 108: What does this do? Should we return an error using `http.Error()` like below?

cmd/block-generator/runner/run.go
```diff
@@ -35,14 +37,21 @@ type Args struct {
...
 
 // Run is a public helper to run the tests.
 // The test will run against the generator configuration file specified by 'args.Path'.
 // If 'args.Path' is a directory it should contain generator configuration files, a test will run using each file.
 func Run(args Args) error {
 	if _, err := os.Stat(args.ReportDirectory); !os.IsNotExist(err) {
-		return fmt.Errorf("report directory '%s' already exists", args.ReportDirectory)
+		if args.ResetReportDir {
+			fmt.Printf("Resetting existing report directory '%s'\n", args.ReportDirectory)
+			os.RemoveAll(args.ReportDirectory)
```
tolikzinovyev, line 51: should we check the error?

tolikzinovyev (commented):
Looks good overall! A couple comments.

cmd/block-generator/generator/generate.go
```diff
@@ -637,3 +650,82 @@ func (g *generator) generateAssetTxn(round uint64, intra uint64) (transactions.S
 
 	return signTxn(txn), transactions.ApplyData{}, nil
 }
+
+func (g *generator) WriteAccount(output io.Writer, accountString string) error {
+	addr, err := basics.UnmarshalChecksumAddress(accountString)
```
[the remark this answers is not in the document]
winder: I started with that, since it makes sense to validate the data early. Ended up with this because I'm using the encoded and decoded version so I avoid re-encoding.

cmd/block-generator/generator/server.go
winder: Yes, good idea.

cmd/block-generator/runner/run.go
winder: yes, thanks

cmd/block-generator/runner/run.go
```diff
@@ -125,8 +162,9 @@ func recordDataToFile(start time.Time, entry Entry, prefix string, out *os.File)
 	record("_cumulative", metrics.PostgresEvalName, floatTotal)
 	record("", metrics.ImportedRoundGaugeName, intTotal)
 
-	if writeErrors.Len() > 0 {
-		return fmt.Errorf("error writing metrics (%s): %w", writeErrors.String(), writeErr)
+	// Allow missing eval time metric
```
tolikzinovyev, line 165: why can it be missing?

winder: I added the exception here because this metric isn't in the build I was using

tolikzinovyev: Can we make an exception for that metric specifically. Otherwise, if there is another missing metric instead, we will not return an error.

winder: I'll remove it, this exception doesn't need to be checked in

tolikzinovyev:
LGTM. Can you please fix CI?

tolikzinovyev (approved):

PR #643, opened 2021-09-10
Use the native interface of the pgx postgres driver

tolikzinovyev:


brianolson:
Is there a measurable speedup?

winder (approved):
Looks pretty straightforward. Could you make a final pass over the imports before merging?

tolikzinovyev:
Brian, yes, about 15%. Will, sorry, what do you mean?

go.sum
```diff
@@ -416,6 +416,7 @@ github.com/jackc/pgx/v4 v4.13.0 h1:JCjhT5vmhMAf/YwBHLvrBn4OGdIQBiFG6ym8Zmdx570=
 github.com/jackc/pgx/v4 v4.13.0/go.mod h1:9P4X524sErlaxj0XSGZk7s+LD0eOyu1ZDUrrpznYDF0=
 github.com/jackc/puddle v0.0.0-20190413234325-e4ced69a3a2b/go.mod h1:m4B5Dj62Y0fbyuIc15OsIqK0+JU8nkqQjsgx7dvjSWk=
 github.com/jackc/puddle v0.0.0-20190608224051-11cab39313c9/go.mod h1:m4B5Dj62Y0fbyuIc15OsIqK0+JU8nkqQjsgx7dvjSWk=
+github.com/jackc/puddle v1.1.3 h1:JnPg/5Q9xVJGfjsO5CPUOjnJps1JaRUm8I9FXVCFK94=
```
tsachiherman, line 419: unless you plan to go back to puddle v0.0.0, you should be good cleaning up. `go mod tidy` would do that for you.

tolikzinovyev: I ran `go mod tidy`, but puddle v0.0.0 is required by github.com/jackc/pgconn@v1.10.0.

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
```diff
@@ -2,13 +2,14 @@ package ledgerforevaluator_test
...
 	"context"
-	"database/sql"
 	"testing"
 
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
```
winder, line 12: this should be grouped with the other non-algorand 3rd party packages

idb/postgres/internal/writer/writer.go
```diff
@@ -11,196 +11,123 @@ import (
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger/ledgercore"
 	"github.com/algorand/go-algorand/protocol"
+	"github.com/jackc/pgx/v4"
```
winder, line 14: different groups for algorand / third party packages

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
tolikzinovyev: I removed the separation between the two, since we already have lots of code where those are together.

idb/postgres/internal/writer/writer.go
tolikzinovyev: We already have code where those are together. I think we should leave it for a separate discussion. My take is that if we had a linter that warned about the imports, it would not be so smart to allow these two separate.

winder: Right, I pointed out the missing groups in one of your earlier PRs but you didn't update them so things are becoming more inconsistent...

tolikzinovyev: Or rather more consistent :)

winder: +1 for a linter - https://github.com/algorand/go-algorand-internal/issues/1431

PR #652, opened 2021-09-15
Implement the new `indexerLedgerForEval` interface in go-algorand

tolikzinovyev:
## Summary

This significantly simplifies the code in Indexer, and paves a way to querying asset/app creators in a batch.

## Test Plan

New tests for batch querying of asset/app creators.

### DIFF
--- a/idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
+++ b/idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
@@ -2,11 +2,8 @@ package ledgerforevaluator
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
-	"github.com/algorand/go-algorand/config"
-	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/data/transactions"
@@ -44,25 +41,21 @@ var statements = map[string]string{
 		"WHERE addr = $1 AND NOT deleted",
 }
 
-// LedgerForEvaluator implements the ledgerForEvaluator interface from
+// LedgerForEvaluator implements the indexerLedgerForEval interface from
 // go-algorand ledger/eval.go and is used for accounting.
 type LedgerForEvaluator struct {
 	tx          pgx.Tx
-	genesisHash crypto.Digest
 	// Indexer currently does not store the balances of special account, but
 	// go-algorand's eval checks that they satisfy the minimum balance. We thus return
 	// a fake amount.
 	// TODO: remove.
 	specialAddresses transactions.SpecialAddresses
-	// Value is nil if account was looked up but not found.
-	preloadedAccountData map[basics.Address]*basics.AccountData
 }
 
 // MakeLedgerForEvaluator creates a LedgerForEvaluator object.
-func MakeLedgerForEvaluator(tx pgx.Tx, genesisHash crypto.Digest, specialAddresses transactions.SpecialAddresses) (LedgerForEvaluator, error) {
+func MakeLedgerForEvaluator(tx pgx.Tx, specialAddresses transactions.SpecialAddresses) (LedgerForEvaluator, error) {
 	l := LedgerForEvaluator{
 		tx:               tx,
-		genesisHash:      genesisHash,
 		specialAddresses: specialAddresses,
 	}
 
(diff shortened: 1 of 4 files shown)


tolikzinovyev:
I think we can now update go-algorand.

tsachiherman:
> I think we can now update go-algorand.

are you referring to the changes in https://github.com/algorand/go-algorand/pull/2846 ?

tolikzinovyev:
No, we were pointing go-algorand to its release branch. Now that we have released a beta Indexer, we can point go-algorand to master again.

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
```diff
@@ -388,96 +376,101 @@ func (l *LedgerForEvaluator) loadAccounts(addresses map[basics.Address]struct{})
...
+			res[address] = accountData
+		}
+	}
+
 	return res, nil
 }
 
-// PreloadAccounts loads the account data for the given addresses and stores them
-// in the internal cache.
-func (l *LedgerForEvaluator) PreloadAccounts(addresses map[basics.Address]struct{}) error {
```
winder, line 396: Cool. So this is no longer needed because algod uses `LookupWithoutRewards`, which is now has a batch-able interface?

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
```diff
@@ -388,96 +376,101 @@ func (l *LedgerForEvaluator) loadAccounts(addresses map[basics.Address]struct{})
...
-	// The balance of a special address must pass the minimum balance check in
-	// go-algorand's evaluator, so return a sufficiently large balance.
-	if l.isSpecialAddress(address) {
-		var balance uint64 = 1000 * 1000 * 1000 * 1000 * 1000
-		accountData := basics.AccountData{
-			MicroAlgos: basics.MicroAlgos{Raw: balance},
-		}
-		return accountData, round, nil
+// GetAssetCreator is part of go-algorand's indexerLedgerForEval interface.
+func (l LedgerForEvaluator) GetAssetCreator(indices map[basics.AssetIndex]struct{}) (map[basics.AssetIndex]ledger.FoundAddress, error) {
```
winder, line 413: Was there any thought about organizing this so that it could be batched along with the account lookup?

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
```diff
@@ -388,96 +376,101 @@ func (l *LedgerForEvaluator) loadAccounts(addresses map[basics.Address]struct{})
...
 
-	switch ctype {
-	case basics.AssetCreatable:
-		row = l.tx.QueryRow(context.Background(), assetCreatorStmtName, uint64(cindex))
-	case basics.AppCreatable:
-		row = l.tx.QueryRow(context.Background(), appCreatorStmtName, uint64(cindex))
-	default:
-		panic("unknown creatable type")
+// GetAppCreator is part of go-algorand's indexerLedgerForEval interface.
+func (l LedgerForEvaluator) GetAppCreator(indices map[basics.AppIndex]struct{}) (map[basics.AppIndex]ledger.FoundAddress, error) {
```
winder, line 442: Same question with this one

idb/postgres/postgres.go
```diff
@@ -223,7 +218,7 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			proto.EnableAssetCloseAmount = true
```
winder, line 218: Do you think things like this could move into algod also?

winder (approved):
Looks great! Just had a couple questions.

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
winder: What about the query duration? Sounds like the RTT isn't really a factor, but running each of these in parallel could have some value.

idb/postgres/postgres.go
winder: The main reason would be to keep consensus param "magic" as close to the source as possible. If we have an interface in algod designed for Indexer, shouldn't we use it?

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
tolikzinovyev: I don't know if we can influence parallelism and gain performance. One thing that comes to mind is read requests going to different read replicas of aurora can help, but we don't have control over it, and I'm not sure the load balancer is smart enough. In addition, other databases (regular postgres) might have worse performance because of parallelism (one thing that comes to mind is HDD where parallel reading causes too many head jumps so the overall performance is worse, not sure how it works for SSD). Also, if we want to parallelize, we should start with parallelizing account query from the `account` table, as that is longest part of reading data.

idb/postgres/postgres.go
tolikzinovyev: I don't have a strong opinion. @tsachiherman maybe does.

PR #656, opened 2021-09-16
Inner transactions and logs support

algorandskiy:
## Summary

REST API changes for inner txn and logs
- `Transaction` model split into `TransactionBase` and inlined object with `id` and `sigs` required fields.
- optional `Logs` and `InnerTxns` fields added into `TransactionBase` model
- new `/applications/{app-id}/logs` endpoint

Indexing:
`ApplyData.InnerTxns` now contribute to transaction participants list

## Test Plan

Added unit tests for REST API
Adeed indexing test


api/converter_utils.go
```diff
@@ -401,6 +422,35 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
...
+		}
+		logs = &l
+	}
+
+	var inners *[]generated.Transaction
+	if len(stxn.ApplyData.EvalDelta.InnerTxns) > 0 {
+		itxns := make([]generated.Transaction, 0, len(stxn.ApplyData.EvalDelta.InnerTxns))
+		for _, t := range stxn.ApplyData.EvalDelta.InnerTxns {
+			extra2 := extra
+			// TODO: fix txn counter/number calculation and use ApplyData.ConfigAsset/ApplicationID
```
algorandskiy, line 439: the latest go-algorand has ApplyData.ConfigAsset/ApplicationID so newer transactions can use it for asset/app ids instead of calculating (both regular and inner)

idb/postgres/internal/writer/writer.go
```diff
@@ -220,13 +208,60 @@ func getTransactionParticipants(txn transactions.Transaction) []basics.Address {
...
+	}
+}
+
+// getTransactionParticipants returns referenced addresses from the txn and all inner txns
+func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) []basics.Address {
+	var res []basics.Address
+	var add func(address basics.Address)
+	const acctsPerTxn = 7
+
+	if stxnad.Txn.Type != protocol.ApplicationCallTx {
```
brianolson, line 222: or we could test on presence of ApplyData InnerTxns?

algorandskiy: done

tolikzinovyev (commented):
I didn't look carefully at the tests, but overall looks good!

api/handlers.go
```diff
@@ -269,6 +269,57 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+
+	// Fetch the transactions
+	txns, next, round, err := si.fetchTransactions(ctx.Request().Context(), filter)
+	if err != nil {
+		return indexerError(ctx, fmt.Sprintf("%s: %v", errTransactionSearch, err))
+	}
+
+	var logData []generated.ApplicationLogData
+	for _, txn := range txns {
+		if txn.Logs != nil && len(*txn.Logs) > 0 {
```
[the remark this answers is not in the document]
tolikzinovyev: never mind

algorandskiy:
In order to exclude ids from inner txns `Transaction` model was split into `TransactionBase` and inlined object with `id` and `sigs` required fields.

api/handlers_test.go
```diff
@@ -253,174 +259,207 @@ func TestFetchTransactions(t *testing.T) {
...
+		// },
+		// {
+		// 	name: "Application With Foreign Assets",
+		// 	txnBytes: [][]byte{
+		// 		loadResourceFileOrPanic("test_resources/app_foreign_assets.txn"),
+		// 	},
+		// 	response: []generated.Transaction{
+		// 		loadTransactionFromFile("test_resources/app_foreign_assets.response"),
+		// 	},
+		// },
```
winder, line 427: Don't forget to re-enable the old tests before merging

algorandskiy: My bad, disabled when was fixing ids. Restored now.

api/indexer.oas2.json
```diff
@@ -945,6 +991,28 @@
...
+    "ApplicationLogData": {
+      "description": "Stores the global information associated with an application.",
+      "type": "object",
+      "required": [
+        "txid",
+        "logs"
+      ],
+      "properties": {
+        "txid": {
+          "description": "Transaction ID",
```
winder, line 1003: Do you think we should specify that this is the "parent" txid?

Thinking out loud about the next bit of work... If we were to add inner transactions to the txn table, we might run into issues with logs being double reported. For example, suppose an inner transaction generates logs and we search for logs on that round. We need to make sure that when we match the parent and the child we only print the logs once.

api/indexer.oas2.json
```diff
@@ -1427,15 +1495,37 @@
     },
     "Transaction": {
       "description": "Contains all fields common to all transactions and serves as an envelope to all transactions type.\n\nDefinition:\ndata/transactions/signedtxn.go : SignedTxn\ndata/transactions/transaction.go : Transaction\n",
+      "allOf": [
```
winder, line 1498: I haven't seen this before. So this is taking the `TransactionBase` type and adding in the id/signature fields?

Do you need the "required" field to mark id/signature as required?

winder: You've got them, nice :+1:

winder: There is one complication when we enable searching for inner transactions.

What happens when we search for transactions and just have an inner transaction in the results?

1) if we lookup the entire tree, this works nicely because we would use this object as designed.
2) if we want to return the TransactionBase it doesn't fit so well anymore

api/indexer.oas2.json
```diff
@@ -2003,6 +2101,13 @@
       "in": "path",
       "required": true
     },
+    "sender-address": {
```
winder, line 2104: we have `address-role=sender` for this.

algorandskiy: this is one for logs only that are in app call txns with Sender only

api/indexer.oas2.json
algorandskiy: I agree it might look a bit confusing. I have not found better way of preserving existing `Transaction`, excluding id/sig and reusing the definition. If we decide empty id/sig is OK for inners then `TransactionBase` can go. Another option is enforce required fields outside of the spec but I do not like this approach

api/indexer.oas2.json
algorandskiy: Current implementation does not look up recursively - it is more about the implementation and not API.
About the `txid`, as discussed, it is an implicit parent id since only regular txns have ids, and specifying "parent" word looks a bit confusing since logs can exist even without inner txns

winder (approved):

idb/postgres/internal/writer/writer.go
```diff
@@ -220,13 +208,59 @@ func getTransactionParticipants(txn transactions.Transaction) []basics.Address {
...
 	add(txn.FreezeAccount)
 
-	return res
+	for _, inner := range stxnad.ApplyData.EvalDelta.InnerTxns {
+		getTransactionParticipantsImpl(inner, add)
+	}
+}
+
+// getTransactionParticipants returns referenced addresses from the txn and all inner txns
+func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) (res []basics.Address) {
```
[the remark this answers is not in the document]
tolikzinovyev: Ok, but could you not use defer? This logic is a bit convoluted. Just call `getTransactionParticipantsImpl()` in both cases.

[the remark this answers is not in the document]
algorandskiy: Will do in subseq PR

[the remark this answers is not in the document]
algorandskiy: done https://github.com/algorand/indexer/pull/661/commits/550d0a30796d3ecdd9c143d0a1a431786eab836b

PR #659, opened 2021-09-17
Encode the address in code instead of in the DB.

winder:
## Summary

Avoid unnecessary 'decode' when searching for an auth-addr.

## Test Plan

Manually tested


tolikzinovyev:
Why is it better?

winder:
It makes the equality check in the DB more efficient. Previously the DB needed to decode this field on every account, now we encode the account once and compare the encoded values to each other directly.

tolikzinovyev (approved):
Makes sense.

PR #661, opened 2021-09-17
Use ApplyData to fill asset/app idx on creation

algorandskiy:
## Summary

* Use `ApplyData.ConfigAsset` and `ApplyData.ApplicationID` while determining asset/app id that is being created in a current txn.

## Test Plan

Added REST API unit test for inner transactions.


tsachiherman:
@winder does this have impact on the block generator ?

api/converter_utils.go
```diff
@@ -401,17 +432,49 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
...
+		itxns := make([]generated.TransactionBase, 0, len(stxn.ApplyData.EvalDelta.InnerTxns))
+		for _, t := range stxn.ApplyData.EvalDelta.InnerTxns {
+			extra2 := extra
+			if t.Txn.Type == protocol.ApplicationCallTx {
+				extra2.AssetID = uint64(t.ApplyData.ApplicationID)
+			} else if t.Txn.Type == protocol.AssetConfigTx {
+				extra2.AssetID = uint64(t.ApplyData.ConfigAsset)
+			} else {
+				extra2.AssetID = 0
+			}
```
winder, line 455: Is it possible to put the intra ID in here too? It isn't strictly required for the REST API, but we'll need it soon when indexing the inner transactions.

algorandskiy: It is already set to parent's one

winder: Yes, but don't we allocate an intra-round offset for each of the inner transactions?

brianolson (commented):
LGTM, but I think I should wait for a second opinion.

api/converter_utils.go
tolikzinovyev: I have the same question.

tolikzinovyev: Also, can you add a comment that `extra2.AssetCloseAmount` should be 0 here?

idb/postgres/internal/writer/writer.go
```diff
@@ -141,13 +141,20 @@ func setSpecialAccounts(addresses transactions.SpecialAddresses, batch *pgx.Batc
 // (0 if not an asset or app transaction).
 func transactionAssetID(block *bookkeeping.Block, intra uint64, typeenum idb.TxnTypeEnum) uint64 {
 	assetid := uint64(0)
+	stxnad := block.Payset[intra]
 	txn := block.Payset[intra].Txn
```
tolikzinovyev, line 145: could you delete this since you are adding `stxnad`?

idb/postgres/internal/writer/writer.go
```diff
@@ -141,13 +141,20 @@ func setSpecialAccounts(addresses transactions.SpecialAddresses, batch *pgx.Batc
...
 
 	switch typeenum {
 	case idb.TypeEnumAssetConfig:
 		assetid = uint64(txn.ConfigAsset)
 		if assetid == 0 {
-			assetid = block.TxnCounter - uint64(len(block.Payset)) + intra + 1
+			if stxnad.ApplyData.ConfigAsset != 0 {
```
tolikzinovyev, line 151: Is `ConfigAsset` set for outer transactions too?

idb/postgres/internal/writer/writer.go
```diff
@@ -232,28 +245,30 @@ func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) (res []basi
...
-		// so the resultant slice is created after collecting all the data from nested transactions.
-		// this is probably a bit slower than the default case due to two mem allocs and additional iterations
-		size := acctsPerTxn * (1 + len(stxnad.ApplyData.EvalDelta.InnerTxns)) // approx
-		participants := make(map[basics.Address]struct{}, size)
-		add = func(address basics.Address) {
-			if address.IsZero() {
-				return
-			}
-			participants[address] = struct{}{}
+
```
tolikzinovyev, line 248: Thanks. Could you also delete `var add func(address basics.Address)` above and not use the named return value? There is no reason.

api/converter_utils.go
winder: > > Yes, but don't we allocate an intra-round offset for each of the inner transactions?
> 
> `intra-round offset` is a position in a payset within a block. What value are you expecting for inners?

Sorry, I was mixing up terms here. The "position in a payset within a block" comment is no longer accurate, it was assuming the TxnCounter increments once per transaction in the payset.

The TxnCounter was changed to increment once per Transaction OR inner transaction so that we can create more than one app/asset inside the inner transactions.

The `intra-round` offset should be incremented in the same way.

algorandskiy: Does it mean all the places that use `TxnRow.Intra` field need to be updated as well?
My thinking was to preserve its meaning in order to find parent within a block if needed. Or are you suggesting tto change `intra-round-offset` only in REST response?

idb/postgres/internal/writer/writer.go
algorandskiy: txn usage here looks ok to me

api/converter_utils.go
tolikzinovyev: We should change how we populate tables `txn` and `txn_participation`, right?

algorandskiy: then probably this should be bundled with upcoming inner txns indexing

algorandskiy: No I don't.

winder (approved):
LGTM

PR #664, opened 2021-09-18
updating go-algorand

barnjamin:
## Summary

Just upping the go-algorand version in the third_party dir so it runs

## Test Plan

in the sandbox just ran `make`, copied the bin to /tmp, ran `/tmp/start.sh` and it started catching up


tolikzinovyev:
Thanks for contribution! The plan is to depend on the release version of go-algorand. Could you point go-algorand to `upstream/rel/beta-3.0.1`?

barnjamin:
updated and appears to build and run with sandbox 👍

tolikzinovyev (approved):

PR #667, opened 2021-09-20
Use COPY to insert into `txn` and `txn_participation` tables

tolikzinovyev:
## Summary

Postgres' `COPY` statement is much faster than `INSERT`. This PR improves import TPS of payment transactions by ~25%. A new migration is added to delete rows from `txn` and `txn_participation` tables for unaccounted rounds, since `COPY` doesn't allow overriding data.

## Test Plan

Using existing tests for testing importing. There is a new test for the migration as well.


idb/postgres/internal/writer/writer_test.go
```diff
@@ -1293,57 +1293,6 @@ func TestWriterAccountAppTableCreateDeleteSameRound(t *testing.T) {
 	assert.Equal(t, block.Round(), basics.Round(closedAt))
 }
 
-// Check that adding same block twice does not result in an error.
```
winder, line 1296: Seems like we should still want to avoid an error in this case. Why remove the test?

tolikzinovyev: COPY doesn't allow overriding rows. This test was needed because with the old indexer imported blocks could be ahead of accounting.

winder: I think you'd still have that case if there were multiple writers deployed

tolikzinovyev: 1. No, because everything is done in one transaction.
2. We don't support multiple writers.

winder: Do we have a test like this:
```diff
- // Check that adding same block twice does not result in an error.
+ // Check that adding same block twice results in an error.
```

tolikzinovyev: There is `TestMultipleWriters()` in `postgres_integration_test.go`.

PR #668, opened 2021-09-20
Fetch block concurrently with import

tolikzinovyev:
## Summary

This improves TPS ~5% for large blocks and ~10% for empty blocks. Update: for empty blocks the improvement is 4.5%, for large blocks it's less. Apparently, fetching a block interferes with accounting, that's why it's less than expected. This is still a good change when algod is remote.

## Test Plan

I ran this code successfully. Any block reorderings are caught in the postgres implementation.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -100,12 +100,11 @@ func (bot *fetcherImpl) setError(err error) {
 
 func (bot *fetcherImpl) processQueue() {
 	for {
-		select {
-		case <-bot.ctx.Done():
+		block, ok := <-bot.blockQueue
+		if !ok {
 			return
-		case block := <-bot.blockQueue:
-			bot.handleBlock(block)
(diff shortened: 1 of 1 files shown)


fetcher/fetcher.go
```diff
@@ -94,6 +98,28 @@ func (bot *fetcherImpl) setError(err error) {
 	bot.errmu.Unlock()
 }
 
+func (bot *fetcherImpl) processQueue() {
+	for {
+		select {
+		case <-bot.ctx.Done():
```
winder, line 104: Should you `close(bot.blockQueue)` here?

tolikzinovyev: It's a good point. I changed the code so that `processQueue()` reads from the channel until it's closed, and we close it when `Run()` exits.

winder (approved):

PR #671, opened 2021-09-22
Runtime configurable server read/write timeouts

gmalouf:
## Summary

Made configurations for read/write timeouts configurable at runtime, with sensible defaults and a basic generation formula for handler timeout.

## Test Plan

Started up server locally, exercised both using default values and passing in command line overrides for each.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -44,9 +44,9 @@ func (e ExtraOptions) handlerTimeout() time.Duration {
 	// If less, subtract 10% as a safety valve.
 	if e.WriteTimeout >= 2*time.Second {
 		return e.WriteTimeout - time.Second
-	} else {
-		return e.WriteTimeout - time.Duration(0.1*float64(e.WriteTimeout))
 	}
+
+	return e.WriteTimeout - time.Duration(0.1*float64(e.WriteTimeout))
 }
 
 // Serve starts an http server for the indexer API. This call blocks.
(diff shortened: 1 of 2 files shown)


winder (approved):

bricerisingalgorand:
Do we provide these timeouts in seconds or milliseconds?

gmalouf:
// ParseDuration parses a duration string.
// A duration string is a possibly signed sequence of
// decimal numbers, each with optional fraction and a unit suffix,
// such as "300ms", "-1.5h" or "2h45m".
// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
func ParseDuration(s string) (Duration, error) {

So you can do "5s" or "5000ms" to get the same parsed duration value.

api/server.go
```diff
@@ -39,6 +30,23 @@ type ExtraOptions struct {
...
+	//off of this.
+	WriteTimeout time.Duration
+
+	// ReadTimeout is the maximum duration for reading the entire request, including the body.
+	ReadTimeout time.Duration
+}
+
+func (e ExtraOptions) handlerTimeout() time.Duration {
+	// Basically, if write timeout is 2 seconds or greater, subtract a second.
+	// If less, subtract 10% as a safety valve.
```
tsachiherman, line 44: valve -> value

gmalouf: Funny, I was definitely intending safety valve as a phrase here.

PR #678, opened 2021-09-24
Add test for accounts with different values.

winder:
## Summary

Add some test cases to see what happens with different microalgo amounts.

## Test Plan

NA


tolikzinovyev:
LGTM. Please fix CI.

tolikzinovyev (approved):

PR #680, opened 2021-09-24
Do not compute pending rewards when rewinding

tolikzinovyev:
## Summary

There are three bugs that make pending rewards calculation incorrect:
- not computing pending rewards when `txcount = 0`
- computing pending rewards even when the account is non-participating
- the latest transaction before the requested round, chosen for the rewards base, did not necessarily flush the pending rewards, so this rewards base is not always correct

Fixing these issues is unrealistic. For now, set pending rewards to 0. In the long term we should either deprecate rewinding or completely redesign it.

## Test Plan

Tested manually.


winder (approved):
LGTM

PR #683, opened 2021-09-27
README.md - load balance configs

### LINKED ISSUE #620: Rewinding with a distributed database
### Subject of the issue

When using a distributed database such as Amazon Aurora, rewinding can fail with an error due to multiple non-transactional requests to the database.

yaovi-a:
## Summary

Added a section in the README for load balance configurations. This should close #620

## Test Plan

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,16 +208,17 @@ sudo systemctl start algorand-indexer
 ```
 
 If you wish to run multiple indexers on one server under systemd, see the comments in `/lib/systemd/system/algorand-indexer@.service` or [misc/systemd/algorand-indexer@.service](misc/systemd/algorand-indexer@.service)
-<!-- USAGE_END_MARKER_LINE -->
-
-# Migrating from Indexer v1
-
-Indexer v1 was built into the algod v1 REST API. It has been removed with the algod v2 REST API, all of the old functionality is now part of this project. The API endpoints, parameters, and response objects have all been modified and extended. Any projects depending on the old endpoints will need to be updated accordingly.
 
 # Unique Database Configurations 
 
 Load balancing: 
 If indexer is deployed with a clustered database using multiple readers behind a load balancer, query discrepancies are possible due to database replication lag. Users should check the `current-round` response field and be prepared to retry queries when stale data is detected.
+
+<!-- USAGE_END_MARKER_LINE -->
+
+# Migrating from Indexer v1
+
+Indexer v1 was built into the algod v1 REST API. It has been removed with the algod v2 REST API, all of the old functionality is now part of this project. The API endpoints, parameters, and response objects have all been modified and extended. Any projects depending on the old endpoints will need to be updated accordingly.
 
 # Building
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -214,6 +214,11 @@ If you wish to run multiple indexers on one server under systemd, see the commen
 
 Indexer v1 was built into the algod v1 REST API. It has been removed with the algod v2 REST API, all of the old functionality is now part of this project. The API endpoints, parameters, and response objects have all been modified and extended. Any projects depending on the old endpoints will need to be updated accordingly.
 
+# Unique Database Configurations 
+
+Load balancing: 
+If indexer is deployed in an environment such that the database is seating behind a load balancer and multiple transaction queries are initiated, it is unclear whether these queries will be handled by a master instance or a non-master instance. Due to the ambiguity behind the handler, users will need to retry their query to ensure that they are not retrieving stale data.
```
tsachiherman, line 220: Seating -> sitting

README.md
```diff
@@ -214,6 +214,11 @@ If you wish to run multiple indexers on one server under systemd, see the commen
 
 Indexer v1 was built into the algod v1 REST API. It has been removed with the algod v2 REST API, all of the old functionality is now part of this project. The API endpoints, parameters, and response objects have all been modified and extended. Any projects depending on the old endpoints will need to be updated accordingly.
 
+# Unique Database Configurations 
+
+Load balancing: 
+If indexer is deployed in an environment such that the database is sitting behind a load balancer and multiple transaction queries are initiated, it is unclear whether these queries will be handled by a master instance or a non-master instance. Due to the ambiguity behind the handler, users will need to retry their query to ensure that they are not retrieving stale data.
```
winder, line 220:
```suggestion
If indexer is deployed with a clustered database using multiple readers behind a load balancer, query discrepancies are possible due to database replication lag. Users should check the `current-round` response field and be prepared to retry queries when stale data is detected.
```

tolikzinovyev: Multiple reader instances are not necessary. As long as there is one reader (in addition to writer), we can have a problem. I would say in the introduction that all write queries go to the writer instance, which then replicates the changes to the reader instances, with a certain lag. But when making a read query, it could go to a reader instance, and the user might get stale data.

+1 to Will's suggestion to check the `current-round` field.

winder: With only 1 reader the user shouldn't see any discrepancy, so I'm not sure we need to explain that part

tolikzinovyev: Why? The 1 reader will still be behind the writer?

tolikzinovyev: Can't a read query also go to the writer instance?

winder: Not with Aurora, they give you 2 different URLs

tolikzinovyev: Well, Indexer takes only 1 URL, and write queries go to the writer instance, and read queries can go to the read instances. Why can't the load balancer forward a read query to the master?

winder (changes requested):
This section should go just above the line with `<!-- USAGE_END_MARKER_LINE -->` so that it's included with the release documentation.

winder (approved):
LGTM

PR #684, opened 2021-09-27
Inner Transaction Search Pt 1

winder:
## Summary

This is the first part of the inner-transaction search. It adds an inner transaction stub to the transaction table.

TODO in follow-up PRs:
1) migration to add old inner transactions to the transaction table.
2) update queries to join in the parent transaction bytes.

## Test Plan

New test cases.


idb/postgres/internal/writer/writer.go
```diff
@@ -228,16 +258,18 @@ func getTransactionParticipantsImpl(stxnad transactions.SignedTxnWithAD, add fun
...
 	}
 }
 
 // getTransactionParticipants returns referenced addresses from the txn and all inner txns
-func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) []basics.Address {
+func getTransactionParticipants(stxnad transactions.SignedTxnWithAD, includeInner bool) []basics.Address {
 	const acctsPerTxn = 7
 
-	if len(stxnad.ApplyData.EvalDelta.InnerTxns) == 0 {
+	if includeInner && len(stxnad.ApplyData.EvalDelta.InnerTxns) == 0 {
```
tolikzinovyev, line 272: `!includeInner ||`

idb/postgres/internal/writer/writer.go
```diff
@@ -252,14 +284,17 @@ func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) []basics.Ad
...
+		getTransactionParticipantsImpl(stxnad, includeInner, add)
 		return res
 	}
 
 	// inner transactions might have inner transactions might have inner...
 	// so the resultant slice is created after collecting all the data from nested transactions.
 	// this is probably a bit slower than the default case due to two mem allocs and additional iterations
-	size := acctsPerTxn * (1 + len(stxnad.ApplyData.EvalDelta.InnerTxns)) // approx
+	size := acctsPerTxn
+	if includeInner {
```
tolikzinovyev, line 295: given the correction above, `includeInner` must be true here

idb/postgres/internal/writer/writer.go
```diff
@@ -278,17 +313,33 @@ func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) []basics.Ad
 	return res
 }
 
-func addTransactionParticipation(block *bookkeeping.Block, batch *pgx.Batch) error {
-	for i, stxnad := range block.Payset {
-		// TODO: replace with a function from go-algorand.
-		participants := getTransactionParticipants(stxnad.SignedTxnWithAD)
+func addInnerTransactionParticipation(stxnad transactions.SignedTxnWithAD, round, intra uint64, batch *pgx.Batch) uint64 {
+	finalIntra := intra
```
tolikzinovyev, line 317: I think there is no reason for a copy.

idb/postgres/internal/writer/writer.go
```diff
@@ -278,17 +313,33 @@ func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) []basics.Ad
...
+		// Only search inner transactions by direct participation.
+		// TODO: Should inner app calls be surfaced by their participants?
+		participants := getTransactionParticipants(itxn, false)
 
 		for j := range participants {
-			batch.Queue(addTxnParticipantStmtName, participants[j][:], uint64(block.Round()), i)
+			batch.Queue(addTxnParticipantStmtName, participants[j][:], round, finalIntra)
 		}
+
+		finalIntra = addInnerTransactionParticipation(itxn, round, finalIntra+1, batch)
```
tolikzinovyev, line 327: Add 1? It would probably be better to have `addInnerTransactionParticipation()` return the index following the last.

tolikzinovyev (changes requested):

idb/idb.go
```diff
@@ -61,6 +61,7 @@ func DecodeTxnRowNext(s string) (round uint64, intra uint32, err error) {
 // TxnExtra is some additional metadata needed for a transaction.
 type TxnExtra struct {
 	AssetCloseAmount   uint64      `codec:"aca,omitempty"`
+	RootTxid           string      `codec:"root-txid,omitempty"`
```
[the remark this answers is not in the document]
winder: Yes, bytes would be better. I'll probably change this to 'intra' in the next PR, so I don't think this needs to change right now.

idb/postgres/internal/writer/writer.go
```diff
@@ -211,13 +241,13 @@ func addTransactions(block *bookkeeping.Block, modifiedTxns []transactions.Signe
 			encoding.EncodeSignedTxnWithAD(stxnad),
 			encoding.EncodeJSON(extra))
 
-		intra += countTxns(modifiedTxns[idx].SignedTxnWithAD)
+		intra, err = addInnerTransactions(modifiedTxns[idx].SignedTxnWithAD, block, intra+1, id, batch)
```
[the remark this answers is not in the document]
winder: That made sense for the previous `countTxns` function. Counting the root transaction made sense.

The logic for adding the root transaction and the inner transaction is a little different, so the function is no longer processing (or counting) the stxn parameter. With that in mind, I changed the way intra is incremented to be part of the parameter. I think this is a pretty common convention for recursive functions.

idb/postgres/internal/writer/writer.go
winder: you're technically right, I have trust issues when it comes to arguments so I avoid modifying them.

idb/postgres/internal/writer/writer.go
winder: I prefer this way, I think it's a better pattern for recursive functions

idb/postgres/internal/encoding/encode.go
```diff
@@ -153,11 +153,21 @@ func convertSignedTxnWithAD(stxn transactions.SignedTxnWithAD) signedTxnWithAD {
 	}
 }
 
+func convertInnerSignedTxnWithAD(stxn transactions.SignedTxnWithAD) signedTxnWithAD {
+	stxn.ApplyData.EvalDelta.InnerTxns = nil
```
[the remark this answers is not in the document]
winder: This one doesn't have a corresponding Decode function, so it clearly isn't reversible. I'll make a comment to call that out.

idb/postgres/internal/writer/writer.go
```diff
@@ -172,12 +177,36 @@ func transactionAssetID(txn transactions.SignedTxnWithAD, intra uint64, block *b
 	return assetid
 }
 
-func countTxns(stxnad transactions.SignedTxnWithAD) uint64 {
-	num := uint64(1)
+func addInnerTransactions(stxnad transactions.SignedTxnWithAD, block *bookkeeping.Block, intra uint64, rootTxid string, batch *pgx.Batch) (uint64, error) {
```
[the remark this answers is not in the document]
tolikzinovyev: This function largely seems to be a duplicate of `addTransactions()`. Have you considered unifying these with something like this? `addTransaction(stxnad, inner bool, ...)`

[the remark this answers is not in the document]
tolikzinovyev: Also, I think we should pass a pointer for `stxnad`. This is because the programmer might think that `stxnad` is owned and we can do anything we want with it, when in fact the maps inside are borrowed. We are in a weird state where some data is copied and owned, and some is borrowed. So I think if we pass an object by value, all maps and slices should be copied; otherwise, we should pass a pointer. I realize some code here already passes transactions by value, but going forward I think we need to improve it.

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
tolikzinovyev: Please use `block.Payset` instead of `modifiedTxns` for consistency and easier testing.

idb/postgres/internal/writer/writer.go
tolikzinovyev: `finalIntra` should be called `next` or similar

idb/postgres/internal/writer/writer.go
```diff
@@ -278,17 +316,37 @@ func getTransactionParticipants(stxnad transactions.SignedTxnWithAD) []basics.Ad
...
 
-func addTransactionParticipation(block *bookkeeping.Block, batch *pgx.Batch) error {
-	for i, stxnad := range block.Payset {
-		// TODO: replace with a function from go-algorand.
-		participants := getTransactionParticipants(stxnad.SignedTxnWithAD)
+// addInnerTransactionParticipation traverses the inner transaction tree and
+// adds txn participation records for each. It performs a preorder traversal
+// to correctly compute the intra round offset, the offset for the next
+// transaction is returned.
+func addInnerTransactionParticipation(stxnad transactions.SignedTxnWithAD, round, intra uint64, batch *pgx.Batch) uint64 {
```
tolikzinovyev, line 323: I think this function can be modified to also add participation for the outer txn and reduce code duplication with `addTransactionParticipation()`.

tolikzinovyev (changes requested):

idb/idb.go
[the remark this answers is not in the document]
winder: No need, I've already started working on that in the next PR

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
winder: OK. I was still considering the idea that we would update the consensus params, but we can change this back later if needed.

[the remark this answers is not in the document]
winder: > Sorry, I meant add 1 to the result of `addInnerTransactions()`. I thought this function returned the last txn added.

`addInnerTransactions` returns the `next` (I have renamed the parameter). So the intra is incremented after each call to `batch.Queue`. It is a little tricky, so I have a test to make sure it's done correctly.

idb/postgres/internal/writer/writer.go
winder: We still need the `Payset` loop and the `InnerTxns` loop. Maybe there's an elegant solution, but I think this one is OK.

idb/postgres/internal/writer/writer.go
winder: `addInnerTransactionParticipation` does return the index following the last, sorry for the confusion.

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
winder: I don't think this is bad, so decided to stick with something that worked. I tried to unify them, but it is tricky. Specifically:
1) the logic has slight differences.
2) we still need to iterate `Payset` and `InnerTxns`, so it is tricky.

tolikzinovyev:
The biggest issue I see is dropping inner txns in `internal/encoding`. Please see my last comment in case you didn't: https://github.com/algorand/indexer/pull/684#discussion_r718713630.

winder:
> The biggest issue I see is dropping inner txns in `internal/encoding`. Please see my last comment in case you didn't: [#684 (comment)](https://github.com/algorand/indexer/pull/684#discussion_r718713630).

Thanks. I moved this out of encode.go.

idb/postgres/internal/writer/writer_test.go
```diff
@@ -64,13 +65,41 @@ func txnQuery(db *pgxpool.Pool, query string) ([]txnRow, error) {
...
 			return nil, err
 		}
 		result.txid = string(txid)
+		result.json = string(json)
+		results = append(results, result)
+	}
+	return results, rows.Err()
+}
+
+type txnParticipation struct {
```
tolikzinovyev, line 82: should we name it `txnParticipationRow()`?

idb/postgres/internal/writer/writer_test.go
```diff
@@ -354,38 +383,24 @@ func TestWriterTxnParticipationTableBasic(t *testing.T) {
...
-	require.NoError(t, err)
-	defer rows.Close()
+	results, err := txnParticipationQuery(db, `SELECT * FROM txn_participation ORDER BY round, intra, addr`)
+	assert.NoError(t, err)
+	assert.Len(t, results, 3)
 
-	var addr []byte
-	var round uint64
-	var intra uint64
+	// check address
```
tolikzinovyev, line 390: I think it would be better to do `assert.Equal(t, txnParticipationRow{...}, results[0])`.

idb/postgres/internal/writer/writer_test.go
```diff
@@ -64,13 +65,41 @@ func txnQuery(db *pgxpool.Pool, query string) ([]txnRow, error) {
...
+		var json []byte
 		err = rows.Scan(&result.round, &result.intra, &result.typeenum,
-			&result.asset, &txid, &result.txnbytes, &result.txn,
+			&result.asset, &txid, &result.txnbytes, &json,
 			&result.extra)
 		if err != nil {
 			return nil, err
 		}
 		result.txid = string(txid)
+		result.json = string(json)
```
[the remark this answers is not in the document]
winder: I believe I tried that and it didn't work.

tolikzinovyev (approved):
Thanks!

PR #687, opened 2021-09-28
Fix "conn busy" issue.

tolikzinovyev:
## Summary

Querying any REST endpoint during catchup sometimes returned an error like this:
```
curl "http://localhost:10001/v2/accounts/FT6OTSGGH5AHWUB537CGH2YWUS6DK7TZ26K2D42JVC3CPH3PLLRF6PWPQY?round=10000000"
{"message":"Indexer health error: getNextRoundToAccount() err: unable to get import state err: getMetastate() err: conn busy"}
```

The problem was that we closed prepared statements after committing the database transaction. This PR fixes the problem.

## Test Plan

Ran indexer and used `watch -n 0.2 curl -s "http://localhost:10000/health"` to watch for any errors.

### DIFF
--- a/idb/postgres/internal/util/util.go
+++ b/idb/postgres/internal/util/util.go
@@ -16,10 +16,10 @@ import (
 
 // TxWithRetry is a helper function that retries the function `f` in case the database
 // transaction in it fails due to a serialization error. `f` is provided
-// a transaction created using `opts`. `f` must either return an error or
-// call sql.Tx.Commit(). In the second case, `f` must return an error which contains the
+// a transaction created using `opts`. `f` should either return an error or
+// call sql.Tx.Commit(). In the second case, `f` should return an error that contains the
 // error returned by sql.Tx.Commit(). The easiest way is to just return the result of
-// sql.Tx.Commit().
+// sql.Tx.Commit(). If sql.Tx.Commit() is not called, results are rolled back.
 func TxWithRetry(db *pgxpool.Pool, opts pgx.TxOptions, f func(pgx.Tx) error, log *log.Logger) error {
 	count := 0
 	for {
--- a/idb/postgres/internal/util/util.go
+++ b/idb/postgres/internal/util/util.go
@@ -16,10 +16,10 @@ import (
 
 // TxWithRetry is a helper function that retries the function `f` in case the database
 // transaction in it fails due to a serialization error. `f` is provided
-// a transaction created using `opts`. `f` takes ownership of the
-// transaction and must either call sql.Tx.Rollback() or sql.Tx.Commit(). In the second
-// case, `f` must return an error which contains the error returned by sql.Tx.Commit().
-// The easiest way is to just return the result of sql.Tx.Commit().
+// a transaction created using `opts`. `f` must either return an error or
+// call sql.Tx.Commit(). In the second case, `f` must return an error which contains the
+// error returned by sql.Tx.Commit(). The easiest way is to just return the result of
+// sql.Tx.Commit().
 func TxWithRetry(db *pgxpool.Pool, opts pgx.TxOptions, f func(pgx.Tx) error, log *log.Logger) error {
 	count := 0
 	for {
@@ -29,6 +29,7 @@ func TxWithRetry(db *pgxpool.Pool, opts pgx.TxOptions, f func(pgx.Tx) error, log
 		}
 
 		err = f(tx)
+		tx.Rollback(context.Background())
 
 		// If not serialization error.
 		var pgerr *pgconn.PgError
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -170,9 +170,8 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 	db.accountingLock.Lock()
 	defer db.accountingLock.Unlock()
 
+	// `f` borrows `tx`.
 	f := func(tx pgx.Tx) error {
-		defer tx.Rollback(context.Background())
-
 		// Check and increment next round counter.
 		importstate, err := db.getImportState(context.Background(), tx)
 		if err != nil {
@@ -242,6 +241,18 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			}
 		}
 
+		return nil
+	}
+	// `ff` takes ownership of `tx`. We split functionality in two functions,
+	// `f` and `ff`, so that we can easily free the database resources once and only once
+	// using defer before committing the database transaction.
+	ff := func(tx pgx.Tx) error {
+		err := f(tx)
+		if err != nil {
+			tx.Rollback(context.Background())
+			return err
+		}
+
 		err = tx.Commit(context.Background())
 		if err != nil {
 			return fmt.Errorf("AddBlock() tx commit err: %w", err)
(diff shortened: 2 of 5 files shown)


idb/postgres/postgres.go
```diff
@@ -242,60 +241,87 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
 
+		return nil
+	}
+	// `ff` takes ownership of `tx`. We split functionality in two functions,
+	// `f` and `ff`, so that we can easily free the database resources once and only once
+	// using defer before committing the database transaction.
+	ff := func(tx pgx.Tx) error {
+		err := f(tx)
+		if err != nil {
+			tx.Rollback(context.Background())
```
winder, line 252: Can this go in `TxWithRetry`? It would be easy to forget this in the future.

idb/postgres/postgres.go
```diff
@@ -242,60 +241,87 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
-	}
-	err = db.setImportState(tx, importstate)
-	if err != nil {
-		return
+		return nil
 	}
 
-	err = tx.Commit(context.Background())
-	return err
+	return db.txWithRetry(serializable, ff)
```
winder, line 324: This was just missing `txWithRetry` previously?

winder (changes requested):
Nice, looks good. Just one nit that can hopefully make the helper easier to use.

idb/postgres/postgres.go
tolikzinovyev: yes

idb/postgres/postgres.go
tolikzinovyev: Good idea!

idb/postgres/internal/util/util.go
```diff
@@ -29,6 +29,7 @@ func TxWithRetry(db *pgxpool.Pool, opts pgx.TxOptions, f func(pgx.Tx) error, log
 		}
 
 		err = f(tx)
+		tx.Rollback(context.Background())
```
tsachiherman, line 32: if `tx.Commit` is called, then `tx.Rollback` wouldn't do anything; so logically it's ok. But it also seems to be wasteful to call `tx.Rollback` if it's not needed.

idb/postgres/internal/util/util.go
```diff
@@ -16,10 +16,10 @@ import (
...
 // TxWithRetry is a helper function that retries the function `f` in case the database
 // transaction in it fails due to a serialization error. `f` is provided
-// a transaction created using `opts`. `f` takes ownership of the
-// transaction and must either call sql.Tx.Rollback() or sql.Tx.Commit(). In the second
-// case, `f` must return an error which contains the error returned by sql.Tx.Commit().
-// The easiest way is to just return the result of sql.Tx.Commit().
+// a transaction created using `opts`. `f` must either return an error or
+// call sql.Tx.Commit(). In the second case, `f` must return an error which contains the
+// error returned by sql.Tx.Commit(). The easiest way is to just return the result of
+// sql.Tx.Commit().
```
winder, line 22: I think it's fine to handle either case and describe what happens.
```suggestion
// a transaction created using `opts`. `f` should either return an error or
// call sql.Tx.Commit(). In the second case, `f` should return an error that contains the
// error returned by sql.Tx.Commit(). The easiest way is to just return the result of
// sql.Tx.Commit(). If sql.Tx.Commit() is not called, results are rolled back.
```

idb/postgres/postgres.go
```diff
@@ -242,60 +240,85 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			}
 		}
 
+		return nil
+	}
+	// We split functionality in two functions, `f` and `ff`, so that we can easily free
+	// the database resources once and only once using defer before committing the database
+	// transaction.
+	ff := func(tx pgx.Tx) error {
```
winder, line 248: Replace this with `return tx.Commit` in f?

idb/postgres/postgres.go
```diff
@@ -242,60 +240,85 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
 		}
-		if len(alloc.State.AssetParams) > 0 || len(alloc.State.Assets) > 0 {
-			return fmt.Errorf("genesis account[%d] has unhandled asset", ai)
+
+		return nil
+	}
+	// We split functionality in two functions, `f` and `ff`, so that we can easily free
+	// the database resources once and only once using defer before committing the database
+	// transaction.
+	ff := func(tx pgx.Tx) error {
```
winder, line 306: same as above

idb/postgres/internal/util/util.go
tolikzinovyev: done

idb/postgres/internal/util/util.go
tolikzinovyev: agree with Will

idb/postgres/postgres.go
```diff
@@ -111,10 +111,10 @@ type IndexerDb struct {
...
 // txWithRetry is a helper function that retries the function `f` in case the database
 // transaction in it fails due to a serialization error. `f` is provided
-// a transaction created using `opts`. `f` takes ownership of the
-// transaction and must either call sql.Tx.Rollback() or sql.Tx.Commit(). In the second
-// case, `f` must return an error which contains the error returned by sql.Tx.Commit().
-// The easiest way is to just return the result of sql.Tx.Commit().
+// a transaction created using `opts`. `f` must either return an error or
+// call sql.Tx.Commit(). In the second case, `f` must return an error which contains the
+// error returned by sql.Tx.Commit(). The easiest way is to just return the result of
+// sql.Tx.Commit().
```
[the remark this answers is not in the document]
tolikzinovyev: done

idb/postgres/postgres.go
tolikzinovyev: If we do, we would have the same problem as before - closing prepared statements after committing the transaction.

winder: Oh, I thought that was what you moved into TxRetry. Is there a way to fix this in the helper function?

tolikzinovyev: Fix what?

idb/postgres/internal/util/util.go
tsachiherman: I think that the patten where `f()` is expected to call `Commit` is risky. I would rather seeing the `BeginTx`/`Commit`/`Rollback` all at the same place.

But that shouldn't be a blocker for this PR.

idb/postgres/postgres.go
winder: The situation where using a helper function incorrectly leads to unexpected orphaned connections.

tolikzinovyev: Since we now call Rollback in `TxWithRetry`, leaving opened transactions is not an issue. There can be other issues like using the transaction after committing it. We could commit in `TxWithRetry`

idb/postgres/internal/util/util.go
winder: I think the pattern we use in go-algorand is intuitive and safe -- `Atomic` manages all of this and a bit more. Maybe we can keep the code as-is and have a followup that does something like that.

idb/postgres/postgres.go
winder: Ok. In that case, I think this can stay as-is, maybe we can have a follow-up that commits in TxWithRetry.

winder (approved):

idb/postgres/internal/util/util.go
winder: Created an issue to capture this idea: https://github.com/algorand/indexer/issues/688

PR #675, opened 2021-09-23
Compute intra the same way TxnCounter increments.

winder:
## Summary

Update the way we assign the `intra` field on the transaction table to match TxnCounter.

For anyone running 2.7.0-beta, this change might cause the 2.7.1 upgrade to break.

## Test Plan

New unit test.

PR #699, opened 2021-10-02
Backport AVM 1.0 DB support to 2.6.2

winder:
## Summary

Update accounting to process inner transactions.
Update transaction participation table to find the root transaction by inner transaction participants.

Some of these changes are backported from these issues:
#656 
#661 
#675 

## Test Plan

Unit tests.

### DIFF
--- a/importer/importer.go
+++ b/importer/importer.go
@@ -45,15 +45,18 @@ func (imp *dbImporter) ImportBlock(blockbytes []byte) (txCount int, err error) {
 }
 
 func countInnerAndAddParticipation(stxns []types.SignedTxnWithAD, participants [][]byte) (int, [][]byte) {
+	copyAddr := func(addr types.Address) []byte {
+		return append([]byte(nil), addr[:]...)
+	}
 	num := 0
 	for _, stxn := range stxns {
-		participants = participate(participants, stxn.Txn.Sender[:])
-		participants = participate(participants, stxn.Txn.Receiver[:])
-		participants = participate(participants, stxn.Txn.CloseRemainderTo[:])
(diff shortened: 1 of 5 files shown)


brianolson (approved):
The one-inner-txn limit in importer is weird, but I believe that is correct and good enough for now.

gmalouf (approved):

importer/importer.go
```diff
@@ -44,6 +44,27 @@ func (imp *dbImporter) ImportBlock(blockbytes []byte) (txCount int, err error) {
 	return imp.ImportDecodedBlock(&blockContainer)
 }
 
+func countInnerAndAddParticipation(stxns []types.SignedTxnWithAD, participants [][]byte) (int, [][]byte) {
+	copyAddr := func(addr types.Address) []byte {
+		return append([]byte(nil), addr[:]...)
+	}
+	num := 0
+	for _, stxn := range stxns {
+		participants = participate(participants, copyAddr(stxn.Txn.Sender))
```
algorandskiy, line 53: why copying address?

algorandskiy (approved):

importer/importer.go
winder: It was being overwritten by the range

algorandskiy: Not sure I understand. `stxn` - yes, it is overwritten and that's why it is an error to take a loop var address.
In the same time referencing to values is fine - they either fully copied (primitives) or shallow-copied and point to different underlying memory in case of slices/maps, like in this example https://play.golang.org/p/6yT1jFVJc3s

PR #711, opened 2021-10-04
Add a migration that fails telling the user that upgrade is not supported

tolikzinovyev:


idb/postgres/postgres_migrations.go
```diff
@@ -199,8 +201,13 @@ func sqlMigration(db *IndexerDb, state *types.MigrationState, sqlLines []string)
 const unsupportedMigrationErrorMsg = "unsupported migration: please downgrade to %s to run this migration"
 
 // disabled creates a simple migration handler for unsupported migrations.
+//lint:ignore U1000 this function might be used in the future
 func disabled(version string) func(db *IndexerDb, migrationState *types.MigrationState) error {
```
winder, line 205: LOL, that was short-lived.

idb/postgres/postgres_migrations.go
```diff
@@ -199,8 +201,13 @@ func sqlMigration(db *IndexerDb, state *types.MigrationState, sqlLines []string)
...
 	return func(_ *IndexerDb, _ *types.MigrationState) error {
 		return fmt.Errorf(unsupportedMigrationErrorMsg, version)
 	}
 }
+
+func upgradeNotSupported(db *IndexerDb, migrationState *types.MigrationState) error {
+	return errors.New("upgrading to Indexer 2.7 is not supported; create a new database")
```
winder, line 212: I think this is slightly more accurate and would still be correct when we reach 2.8.
```suggestion
	return errors.New("upgrading from this version is not supported; create a new database")
```

winder (changes requested):
Great idea

idb/postgres/postgres_migrations.go
tolikzinovyev: good idea

winder (approved):

PR #712, opened 2021-10-04
Add e2e extra pages test

algorandskiy:
## Summary

Added idb to rest test for extra pages

## Test Plan

This is a new test


api/handers_e2e_test.go
```diff
@@ -0,0 +1,99 @@
...
+					GenesisHash: test.GenesisHash,
+				},
+				ApplicationCallTxnFields: transactions.ApplicationCallTxnFields{
+					ApprovalProgram:   []byte{0x02, 0x20, 0x01, 0x01, 0x22},
+					ClearStateProgram: []byte{0x02, 0x20, 0x01, 0x01, 0x22},
+					ExtraProgramPages: extraPages,
+				},
+			},
+			Sig: test.Signature,
+		},
```
winder, line 62: Technically, the ApplyData / ApplicationID should be set here. I think it's even used with the new accounting code.

winder (changes requested):
Nice pattern here! One small comment about the test, I think you can inject the app index to avoid knowing that it's going to be 1.

api/handers_e2e_test.go
algorandskiy: I tried and it did not work. I thought because the it uses re-evaluated txn and not txn from a block, but looks like it is [not the case](https://github.com/algorand/indexer/blob/develop/idb/postgres/internal/writer/writer.go#L202). I'll take another look.

winder (approved):

api/handers_e2e_test.go
```diff
@@ -0,0 +1,102 @@
+package api
```
tolikzinovyev, line 1: typo in file name

algorandskiy: thanks, fixed

PR #716, opened 2021-10-05
Do not encode addresses as base64

tolikzinovyev:
## Summary

Now that we don't need backward compatibility, we can remove special encoding handling for addresses. I needed to change some database queries though.

## Test Plan

TODO


idb/postgres/postgres.go
```diff
@@ -1639,7 +1639,7 @@ func (db *IndexerDb) Assets(ctx context.Context, filter idb.AssetsQuery) (<-chan
 	}
 	if filter.Creator != nil {
 		whereParts = append(whereParts, fmt.Sprintf("a.creator_addr = $%d", partNumber))
```
winder, line 1641: were you thinking about changing these to strings as well?

tolikzinovyev: Changing `creator_addr` column from bytes to human readable format? I wasn't.

winder: Yes, that's what I was thinking. A weird side effect of these changes is that we would no longer be able to join regular columns on jsonb fields.

I don't think that's a problem for Indexer but might limit people querying the DB directly.

@anya2go do you ever join the Indexer byte address columns with json columns?

tolikzinovyev:
The import performance degrades by 6% with this, so I'm closing this PR.

PR #717, opened 2021-10-05
Cleanup in import and migration state

tolikzinovyev:


winder (commented):
Wouldn't this prevent us from showing the upgrade error message if people try to run the new version with a version < 2.7.x?

PR #721, opened 2021-10-08
Inner Transaction Search

winder:
## Summary

Update queries to support returning the root transaction bytes when matching an inner transaction.

## Test Plan

New unit tests.

### DIFF
--- a/cmd/block-generator/generator/generate.go
+++ b/cmd/block-generator/generator/generate.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/algorand/go-algorand/agreement"
+	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
 	"github.com/algorand/go-algorand/data/basics"
@@ -259,6 +260,20 @@ func (g *generator) WriteGenesis(output io.Writer) error {
 			},
 		})
 	}
+	// Also add the rewards pool account with minimum balance. Without it, the evaluator
+	// crashes.
+	proto, ok := config.Consensus[g.protocol]
+	if !ok {
+		return fmt.Errorf("protocol version %s not found", g.protocol)
+	}
+	allocations = append(allocations, bookkeeping.GenesisAllocation{
+		Address: g.rewardsPool.String(),
+		Comment: "RewardsPool",
+		State: basics.AccountData{
+			MicroAlgos: basics.MicroAlgos{Raw: proto.MinBalance},
+			Status:     basics.NotParticipating,
+		},
+	})
 
 	gen := bookkeeping.Genesis{
 		SchemaID:    "v1",


idb/idb.go
```diff
@@ -28,6 +31,9 @@ type TxnRow struct {
 	// TxnBytes is the raw signed transaction with apply data object.
 	TxnBytes []byte
 
+	// RootTxnBytes the root transaction raw signed transaction with apply data object, only inner transactions have this.
+	RootTxnBytes []byte
```
winder, line 35: Coalesce could be used to return RootTxnBytes / TxnBytes in the same parameter, but that could make it easy to mix up the other TxnRow

api/handlers.go
```diff
@@ -495,16 +495,42 @@ func (si *ServerImplementation) SearchForTransactions(ctx echo.Context, params g
...
 
+	// Used for filtering duplicates after getting results.
+	rootTxnDedupeMap := make(map[string]struct{})
+	if filter.NextToken != "" {
+		_, _, txid, err := idb.DecodeTxnRowNext(filter.NextToken)
+		if err != nil {
+			badRequest(ctx, "bad next token")
+		}
+		rootTxnDedupeMap[txid] = struct{}{}
+	}
```
winder, line 506: Bootstrapping the dedupe map with NextToken solves the duplicate-root-txn-while-paging problem.

idb/idb.go
```diff
@@ -39,29 +45,51 @@ type TxnRow struct {
...
 	return
 }
 
 // TxnExtra is some additional metadata needed for a transaction.
 type TxnExtra struct {
 	AssetCloseAmount uint64 `codec:"aca,omitempty"`
-	RootTxid         string `codec:"root-txid,omitempty"`
+	// RootIntra is a string to allow distinguishing between 0 and empty.
+	RootIntra string `codec:"root-intra,omitempty"`
+	RootTxid  string `codec:"root-txid,omitempty"`
```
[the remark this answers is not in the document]
winder: No. Yes.

idb/postgres/postgres.go
```diff
@@ -722,11 +733,12 @@ func (db *IndexerDb) txnsWithNext(ctx context.Context, tx pgx.Tx, tf idb.Transac
...
 		// (round,intra) descending into the past
 		tf.OffsetLT = origOLT
-		if nextround == 0 {
+		tf.MaxRound = nextround - 1
+
+		if tf.MaxRound == 0 {
 			// NO second query
 			return
 		}
-		tf.MaxRound = nextround - 1
```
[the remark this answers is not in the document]
winder: The problem I saw was that when `nextround` is  1, we set `tf.MaxRound = 0`. For this parameter the zero value indicates the value isn't set, so we get a bunch of unexpected results.

I think the old code was written under the assumption that transactions on round 0 was a valid thing.

Looking at this again, I think I can change the early exit to `if nextround <= 1 { return }` to improve this check slightly.

idb/idb.go
```diff
@@ -39,29 +45,56 @@ type TxnRow struct {
...
+	}
+
+	return base64.URLEncoding.EncodeToString(b[:]), nil
 }
 
 // DecodeTxnRowNext unpacks opaque string returned from TxnRow.Next()
-func DecodeTxnRowNext(s string) (round uint64, intra uint32, err error) {
-	var b []byte
-	b, err = base64.URLEncoding.DecodeString(s)
+func DecodeTxnRowNext(s string) (round uint64, intra uint32, rootTxid string, err error) {
```
brianolson, line 69: If we wanted to be clever we could decode the previous version of next-token here based on the decoded length; but I don't care enough to press it and approximately zero API paging sequences will be disrupted.

winder: Ok, it never occurred to me to consider someone saving the next-token string between releases. I think I'm comfortable letting that generate an error.

api/handlers.go
```diff
@@ -495,16 +495,42 @@ func (si *ServerImplementation) SearchForTransactions(ctx echo.Context, params g
...
+
 	// Fetch the transactions
 	txns, next, round, err := si.fetchTransactions(ctx.Request().Context(), filter)
```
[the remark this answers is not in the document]
winder: done.

PR #730, opened 2021-10-11
Statically link glib libraries

egieseke:
## Summary

Use static linker flags to support static linking of go-algorand dependencies on non-glibc based operating sytems.

## Test Plan

Testing build and deployment within a docker container.


brianolson (approved):
LGTM. I wish it was possible to do nice indenting on the if-blob added to the Makefile, but oh well.

PR #739, opened 2021-10-15
Delete Null characters from Logs

### LINKED ISSUE #735: Bug - unsupported characters using log
### Subject of the issue

Indexer may crash if there are unsupported characters when using log.

```
INDEX:{"error":"error committing block, CommitBlock(): COPY txn Exec() pq: unsupported Unicode escape sequence","level":"error","msg":"ImportDecodedBlock 5","time":"2021-10-14T16:04:37Z"}
```

As reported by someone external, the following errors:
```
byte base64(AAAAAAAAAAA=)
log
```

That's 8 0s and it crashes the indexer, then

```
byte base64(AQEBAQEBAQE=)
log
```

That's 8 1s and it DOESN'T crash

Using a fresh network. May be related to: https://github.com/algorand/indexer/pull/656

### Your environment

Sandbox can be used to verify.

### Expected behaviour

Log result.

### Actual behaviour

Crashes on import.

algochoi:
## Summary

This commit strips null characters from logs, which should fix indexers from crashing.

Closes #735

### DIFF
--- a/misc/Dockerfile
+++ b/misc/Dockerfile
@@ -6,7 +6,7 @@ RUN echo "Go image: $GO_IMAGE"
 # Misc dependencies
 ENV HOME /opt    
 ENV DEBIAN_FRONTEND noninteractive    
-RUN apt-get update && apt-get install -y apt-utils curl git git-core bsdmainutils python3 python3-pip make
+RUN apt-get update && apt-get install -y apt-utils curl git git-core bsdmainutils python3 python3-pip make libffi-dev build-essential
 
 # Install algod nightly binaries to the path
 RUN mkdir -p /opt/algorand/{bin,data}


idb/postgres/internal/encoding/encode.go
```diff
@@ -81,7 +82,19 @@ func convertLocalDeltas(deltas map[uint64]types.StateDelta) map[uint64]types.Sta
 	return res
 }
 
+func stripNull(logs []string) []string {
+	if logs == nil {
+		return nil
+	}
+	res := make([]string, len(logs))
+	for i, log := range logs {
+		res[i] = strings.ReplaceAll(log, "\x00", "")
```
tsachiherman, line 91: This would remove all the nil characters, which would fail the json persistence; however, please note that there are other non-uff-8 characters that aren't `\x00` which would also fail in the same glorious way.

onetechnical: We think the null chars are the only ones causing problems with postgres, but we'll test.

tsachiherman: consider using the following to ensure that all the characters are truly "nice" utf characters:
```golang
// unicodePrintable scan the input string str, and find if it contains any
// non-printable unicode characters. If so, it returns false, along with the
// printable characters that do appear in the input string. Otherwise, it
// returns true, along with a copy of the original string. The returned string
// printableString is gurenteed to be printable in all cases.
func unicodePrintable(str string) (isPrintable bool, printableString string) {
	isPrintable = true
	encRuneBuf := make([]byte, 8)
	for _, r := range str {
		if !unicode.IsPrint(r) {
			isPrintable = false
		} else {
			n := utf8.EncodeRune(encRuneBuf, r)
			printableString += string(encRuneBuf[:n])
		}
	}
	return
}
```

tsachiherman: or, maybe a complete all-or-nothing approach:
```golang
// printableUTF8OrEmpty checks to see if the entire string is a UTF8 printable string.
// If this is the case, the string is returned as is. Otherwise, the empty string is returned.
func printableUTF8OrEmpty(in string) string {
	// iterate throughout all the characters in the string to see if they are all printable.
	// when range iterating on go strings, go decode each element as a utf8 rune.
	for _, c := range in {
		// is this a printable character, or invalid rune ?
		if c == utf8.RuneError || !unicode.IsPrint(c) {
			return ""
		}
	}
	return in
}
```

jasonpaulos: [utf8.ValidString](https://pkg.go.dev/unicode/utf8#ValidString) is also an option

tsachiherman (commented):
Looks correct to me.

idb/postgres/internal/encoding/encode.go
winder: This is binary data, not a string. Removing the null byte is more correct than verifying UTF8.

The alternative would be to b64 encode the data manually.

winder: https://www.postgresql.org/docs/11/datatype-json.html

winder: I see, I thought this was supposed to be bytes but the raw type is `[]string`:
https://github.com/algorand/indexer/blob/hotfix/2.6.3/types/types.go#L274

The REST API correctly declares it as `[][]byte`:
https://github.com/algorand/indexer/blob/develop/api/generated/v2/types.go#L581

PR #741, opened 2021-10-15
Apply hotfix/2.6.4 to the develop branch

algochoi:
## Summary

Applies #736 to the develop branch, which removed non UTF-8 characters from logs in Indexer.


tolikzinovyev:
Also for the future: we should probably use `[]byte` instead of `string` for strings that are not necessarily utf8 (in go-algorand). `string` in Go is supposed to be utf8.

PR #743, opened 2021-10-20
Add LogsOverride and integration test for non UTF-8 Logs

### LINKED ISSUE #740: Properly encode / decode log messages in postgres
### Subject of the issue

This is a followup to #735 to properly fix this issue. We issued a hotfix in v2.6.4 to just strip non-UTF-8 characters, but would like to handle this properly in the main line. We should ensure that when storing we properly store base64 encoded data in postgres, which will avoid this issue in the future.

Notes from Pavel:

Why base64 and not logsOverride [][]byte in internal/types.go?
REST already returns base64 encoded, it is just a storage issue. Cast to []byte and let postgres to store it as it wants
and remove conversion string -> []byte from rest since it would be already []byte

Included in this should be tests.

### Your environment

N/A

### Steps to reproduce

1. Instantiate a sandbox indexer off of the develop branch
2. Create a transaction using the log opcode with null chars
3. Observe indexer falling over

### Expected behaviour

Indexer able to properly store and return blocks.

### Actual behaviour

Indexer crashes on import.

algochoi:
## Summary

This PR adds a `LogsOverride` and `InnerTxnsOverride` field to `EvalDelta`, and appropriate encode/decode functions, so indexer will properly store logs as bytes and handle non UTF-8 strings. 

Closes #740 .

## Test Plan

Adds an integration test to check that the database can handle non UTF-8 characters in the logs. Also adds encoding tests for logs and inner txns.

### DIFF
--- a/idb/postgres/internal/encoding/encoding_test.go
+++ b/idb/postgres/internal/encoding/encoding_test.go
@@ -62,31 +62,8 @@ func TestEncodeSignedTxnWithADSynthetic(t *testing.T) {
 		Bytes:  string(nonutf8b),
 	}
 	stxn.EvalDelta.LocalDeltas[1] = ld
-	stxn.EvalDelta.Logs = []string{nonutf8}
-
-	// Test Inner transaction Logs encoding
-	stxn.ApplyData.EvalDelta.InnerTxns = []transactions.SignedTxnWithAD{{
-		ApplyData: transactions.ApplyData{
-			EvalDelta: transactions.EvalDelta{
-				GlobalDelta: map[string]basics.ValueDelta{
-					nonutf8: {
-						Action: basics.SetBytesAction,
-						Bytes:  string(nonutf8b),
-					},
-				},
-				LocalDeltas: map[uint64]basics.StateDelta{
-					1: {nonutf8: {
-						Action: basics.SetBytesAction,
-						Bytes:  string(nonutf8b),
-					}},
-				},
-				Logs: []string{nonutf8},
-			},
-		},
-	}}
-
 	js := EncodeSignedTxnWithAD(stxn)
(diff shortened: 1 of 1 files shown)


idb/postgres/internal/encoding/types.go
```diff
@@ -68,6 +68,7 @@ type evalDelta struct {
 	transactions.EvalDelta
 	GlobalDeltaOverride stateDelta            `codec:"gd"`
 	LocalDeltasOverride map[uint64]stateDelta `codec:"ld"`
+	LogsOverride        [][]byte              `codec:"lg"`
```
jasonpaulos, line 71: We also need to add the following (plus encoding and decoding functions) in order to apply the existing conversions to inner transactions as well.

```go
InnerTxnsOverride []signedTxnWithAD `codec:"itx"`
```

idb/postgres/internal/encoding/encode.go
```diff
@@ -138,11 +138,24 @@ func convertLocalDeltas(deltas map[uint64]basics.StateDelta) map[uint64]stateDel
 	return res
 }
 
+func convertLogs(logs []string) [][]byte {
```
jasonpaulos, line 141: As discussed, this needs an equivalent `unconvertLogs` function in `idb/postgres/internal/encoding/decode.go` as well.

idb/postgres/internal/encoding/encoding_test.go
```diff
@@ -62,8 +62,31 @@ func TestEncodeSignedTxnWithADSynthetic(t *testing.T) {
 		Bytes:  string(nonutf8b),
 	}
 	stxn.EvalDelta.LocalDeltas[1] = ld
+	stxn.EvalDelta.Logs = []string{nonutf8}
```
tolikzinovyev, line 65: Could you please move this to `TestSignedTxnWithADEncoding()`? This test is old and doesn't test decoding.

PR #744, opened 2021-10-20
Update go-algorand

tolikzinovyev:
## Summary

Bump go-algorand to the latest master commit. To enable special accounts, we need to make a change in go-algorand so that account totals are updated in eval for Indexer. Then, we need to make quite a few changes in Indexer to enable special accounts. To avoid large PRs, this one resolves recent regressions in go-algorand related to Indexer.


winder (approved):
Assuming that we are OK waiting to release Indexer until the next algod release, LGTM!

PR #748, opened 2021-10-21
Travis to circleci migration

algojack:
## Summary
We were having issues with Travis, so decided to port this to circleci sooner than planned

## Test Plan
Tested in circleci


algojack:
Still need to find a way to pass integration tests without reverting #730

PR #749, opened 2021-10-22
Preload asset/app creators before evaluation

tolikzinovyev:
## Summary

Preloading asset/app creators in a batch improves performance by ~10%.

Depends on #746.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -203,7 +203,10 @@ func prepareEvalResources(l *ledger_for_evaluator.LedgerForEvaluator, block *boo
 		}
 	}
 
-	var res ledger.EvalForIndexerResources
+	res := ledger.EvalForIndexerResources{
+		Accounts: nil,
+		Creators: make(map[ledger.Creatable]ledger.FoundAddress),
+	}
 
 	assetCreators, err := l.GetAssetCreator(assets)
 	if err != nil {
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -19,6 +19,7 @@ import (
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger"
 	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/protocol"
 
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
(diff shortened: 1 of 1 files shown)


winder (commented):
This looks correct, but should it be part of the EvalForIndexer interface? I thought that interface's main purpose was to make it easier to preload everything.

tolikzinovyev:
> This looks correct, but should it be part of the EvalForIndexer interface? I thought that interface's main purpose was to make it easier to preload everything.

I don't think I understand.

winder:
> > This looks correct, but should it be part of the EvalForIndexer interface? I thought that interface's main purpose was to make it easier to preload everything.
> 
> I don't think I understand.

I'm asking about the design here, not the implementation.

You're using a combination of these new methods:
```
	LookupWithoutRewards(map[basics.Address]struct{}) (map[basics.Address]*basics.AccountData, error)
	GetAssetCreator(map[basics.AssetIndex]struct{}) (map[basics.AssetIndex]FoundAddress, error)
	GetAppCreator(map[basics.AppIndex]struct{}) (map[basics.AppIndex]FoundAddress, error)
```

I thought one reason for those methods was to avoid needing this sort of preloading, so I wonder if the interface is wrong.

Should it be a single method instead of 3? Or perhaps we add a 4th method to the interface specifically designed to help with caching.

tolikzinovyev:
If your suggestion is to have the code that decides what to preload in go-algorand, then @tsachiherman was against it. We still need the interface functions in case we forget to preload something and the evaluator needs it. Let me know if this doesn't quite answer your questions.

tsachiherman:
@winder - there are two "faces" for the evaluator:
We try to preload all the required resources before the evaluation. However, if during evaluation we find that we need additional resources, we need to provide the functionality to retrieve them.

The reason the evaluator need to be "dynamic" is that it's also being used for the transaction pool, where we add transactions on the fly. I wouldn't oppose to open a discussion on whether the evaluator need to be resource-predictable or not ( outside the scope of this PR ).

winder:
@tolikzinovyev @tsachiherman thanks for the explanations. I guess the difference between algod and indexer is that algod's caching system exists outside of these functions and Indexer's is using them to initialize the cache.

It might be good to rename this function to something like `preloadEvalResources` to be more specific about what sort of preparation is happening.

Do we need to modify `GetAssetCreator` and `GetAppCreator` to check the cache? Wouldn't the evaluator call these functions a second time after preloading everything?

PR #760, opened 2021-10-27
Improve DB indexes for better worst case accounting time

tolikzinovyev:
## Summary

Closes https://github.com/algorand/go-algorand-internal/issues/1718.

## Test Plan

Ran block generator. No noticeable difference in performance.


gmalouf (changes requested):
Question: Do we need to manage migration for existing deployments, or are we just assuming no backwards compatibility with upcoming release / simply deleting references to pre-existing indexes? cc @winder 

I.E. Some of these indexes being updated it seems we would want runners to drop before applying the new ones, yet we give no on/off ramp to do this at first glance.

tolikzinovyev:
The coming "refactoring" release will no backward compatibility with the previous ones. The users will have to create a new database and do catch up from round 0. This is simply because 2.6.* has accounting problems, and 2.7 will in all likelihood crash if we try using a 2.6 database. Hence, we should use the opportunity to make any changes that would require a migration if we were to support backward compatibility.

gmalouf:
I'm approving, but have we run tests with these indexes at least locally to show the performance improvement?

gmalouf (approved):

gmalouf (commented):
We need to keep in mind that not everyone uses the same indexes, although these in particular -should- be universal. We'll need to work with release eng to communicate the changes in existing ones for people who pick and choose which ones they apply today..

winder (approved):
Seems fine to me.

tolikzinovyev:
> I'm approving, but have we run tests with these indexes at least locally to show the performance improvement?

The performance is asymptotically better in the worst case (when the number of deleted items is much larger than the number of non-deleted items). I'm not sure testing performance is even necessary.

PR #763, opened 2021-10-29
Use integer for root intra in `TxnExtra`

tolikzinovyev:
## Summary

Found a way use integer for root intra. This doesn't change the format of json.

## Test Plan

Added usual encoding tests.

### DIFF
--- a/idb/idb.go
+++ b/idb/idb.go
@@ -44,8 +44,8 @@ type TxnRow struct {
 	Error error
 }
 
-func countInner(stxn *transactions.SignedTxnWithAD) int {
-	num := 0
+func countInner(stxn *transactions.SignedTxnWithAD) uint {
+	num := uint(0)
 	for _, itxn := range stxn.ApplyData.EvalDelta.InnerTxns {
 		num++
 		num += countInner(&itxn)
@@ -59,10 +59,10 @@ func (tr TxnRow) Next(ascending bool) (string, error) {
 	var b [12]byte
 	binary.LittleEndian.PutUint64(b[:8], tr.Round)
 
-	intra := tr.Intra
-	if tr.Extra.RootIntra != "" {
+	intra := uint(tr.Intra)
+	if tr.Extra.RootIntra.Present {
 		// initialize for descending order, the root intra.
-		intra, err = strconv.Atoi(tr.Extra.RootIntra)
+		intra = tr.Extra.RootIntra.Value
 		if err != nil {
 			return "", fmt.Errorf("Next() could not parse root intra: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


tolikzinovyev:
Our current codec without the `RecursiveEmptyCheck` option and a pointer also achieves it, but I thought the reason why with this option `nil` and `&0` are both treated as empty, is because we are not supposed to use a pointer for an optional value.

The only extra code here is the marshal/unmarshal functions plus tests. I think this is worth having to isolate encoding and logic.

idb/idb.go
```diff
@@ -106,13 +106,45 @@ func DecodeTxnRowNext(s string) (round uint64, intra uint32, err error) {
...
+// MarshalText implements TextMarshaler interface.
+func (ou OptionalUint) MarshalText() ([]byte, error) {
+	if !ou.Present {
+		return nil, nil
+	}
+	return []byte(fmt.Sprintf("%d", ou.Value)), nil
+}
+
+// UnmarshalText implements TextUnmarshaler interface.
+func (ou *OptionalUint) UnmarshalText(text []byte) error {
```
gmalouf, line 124: This feels more like a utility function that should live independently of idb and have its own testing.

idb/idb.go
```diff
@@ -106,13 +106,45 @@ func DecodeTxnRowNext(s string) (round uint64, intra uint32, err error) {
...
+
+	return nil
+}
+
 // TxnExtra is some additional metadata needed for a transaction.
 type TxnExtra struct {
 	AssetCloseAmount uint64 `codec:"aca,omitempty"`
 	// RootIntra is set on inner transactions. Combined with the confirmation
 	// round it can be used to lookup the root transaction.
 	// The type is string to allow distinguishing between 0 and empty.
```
gmalouf, line 146: Update the comment to match code changes.

gmalouf (commented):
Biggest ask - what is the problem this is solving? Is it a driver of our mismatches/or being done to more closely align with go-algorand?

idb/idb.go
tolikzinovyev: done

idb/idb.go
tolikzinovyev: This is a method on type `OptionalUint`, so we can't separate them. Do you think I should move the type elsewhere?

tolikzinovyev:
> Biggest ask - what is the problem this is solving? Is it a driver of our mismatches/or being done to more closely align with go-algorand?

This change is too separate encoding from the business logic. As you can see, before we were doing integer <-> string conversions in the business logic. Functionally, this PR doesn't change anything.

tolikzinovyev:
More concretely, the json codec will invoke `MarshalText()` / `UnmarshalText()` methods when encoding / decoding json.

idb/idb.go
gmalouf: Sorry, I was referring to the overall OptionalUint type as being general purpose - though we should have at least 1 other use-case before moving it (I'm used to optionals etc from functional languages).

gmalouf (approved):

idb/idb.go
```diff
@@ -106,13 +106,44 @@ func DecodeTxnRowNext(s string) (round uint64, intra uint32, err error) {
...
+}
+
 // TxnExtra is some additional metadata needed for a transaction.
 type TxnExtra struct {
 	AssetCloseAmount uint64 `codec:"aca,omitempty"`
-	// RootIntra is set on inner transactions. Combined with the confirmation
+	// RootIntra is set only on inner transactions. Combined with the confirmation
 	// round it can be used to lookup the root transaction.
-	// The type is string to allow distinguishing between 0 and empty.
-	RootIntra string `codec:"root-intra,omitempty"`
```
winder, line 115: It just occurred to me that we can use `RootTxid` to determine if this field exists or not. We can avoid special encoding/decode code entirely and use a uint64 type here and use `len(RootTxid) > 0` as the "Present" condition.

tolikzinovyev: The query would need to change though to handle the special case when `t.extra->>'root-intra'` is empty string. https://github.com/algorand/indexer/pull/721/files#diff-8fbec00be6e6b2f77a294cbe9a3ee8e3fbdcb5eb13be3046e5da8277302264c2R637
If you would like to change it, I will not object.

By the way, I'm afraid this query has quadratic complexity. I'm not sure postgres is smart enough to look up the right `root` txn with one query. Possibly, for each `txn` row there will be a scan of all rows for this round.

PR #764, opened 2021-11-01
Search for serialization error in function's return error

tolikzinovyev:
## Summary

We used to assume that false positive serialization errors with postgres are only returned from `Commit()`. Turns out, any database operation such as reading rows can return this error.

## Test Plan

Haven't been able to reproduce this behavior.

### DIFF
--- a/idb/postgres/internal/util/util.go
+++ b/idb/postgres/internal/util/util.go
@@ -14,36 +14,50 @@ import (
 	"github.com/algorand/indexer/idb"
 )
 
+func isSerializationError(err error) bool {
+	var pgerr *pgconn.PgError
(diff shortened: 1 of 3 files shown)


gmalouf (commented):
Thoughts on adding a test to prove the serialization testing works as expected?

winder (approved):
Seems fine to me. I guess this would be pretty easy to test by putting some sort of delay into the callback.

Now that it's defined/used in several places, it might be worth adding `type TxnHandler func(pgx.Tx) error`. Then you can move some of the comments from TxWithRetry directly to that type detailing its requirements.

tolikzinovyev:
> Thoughts on adding a test to prove the serialization testing works as expected?

Added a test to check that the return value of `f` is searched for a serialization error. However, mocking a database transaction to check that the return value of `Commit()` is checked is more difficult.

PR #611, opened 2021-08-09
Use go-algorand types

tolikzinovyev:
## Summary

Replace indexer's own types and sdk types with types from go-algorand. This is the first big change towards using the accounting code from go-algorand.

A big portion of this PR has to do with making sure the types from go-algorand are encoded into json the same way as before. It is needed for backward compatibility, because we store these json values in the indexer database. An example of when special handling is needed is when a go `string` has to be encoded as base64 in json because the string might contain non-utf8 characters. To accomplish that, we embed the original `struct` that contains the string in a new struct and add a new field of type `[]byte` with the same codec tag. The json codec will then encode/decode from this new override field (`[]byte` is encoded as base64).

## Test Plan

There are unit tests for the new encoding/decoding functions in `/idb/postgres/internal/encoding`.

PR #765, opened 2021-11-01
Use our own json handle for decoding

tolikzinovyev:
## Summary

We already use our own json handle for encoding. For consistency and to guard against changes in go-algorand, use the same json handle for decoding.


gmalouf (commented):
Why move the jsonCodeHandle into a new `encoding.go` file in the same package? Readability, or is there something obvious I am missing?

winder (changes requested):
I agree with the comments Brian made some time ago. I'd rather see encoding/decoding merged into a single file with all functions side-by-side. For me, separate encode/decode files make this code more difficult to read and review, further separation would make things even more difficult.

https://github.com/algorand/indexer/pull/611#discussion_r686858713

tolikzinovyev:
> I agree with the comments Brian made some time ago. I'd rather see encoding/decoding merged into a single file with all functions side-by-side. For me, separate encode/decode files make this code more difficult to read and review, further separation would make things even more difficult.
> 
> [#611 (comment)](https://github.com/algorand/indexer/pull/611#discussion_r686858713)

I really think it's not a problem, as you can have two windows side-by-side, but I put encoding, decoding functions in the same file.

winder (approved):
LGTM

gmalouf (approved):

PR #770, opened 2021-11-03
Write app call addresses in `txn_participation` table

### LINKED ISSUE #768: Add application call accounts to `txn_participation` table
### Subject of the issue

We currently add the following to the `txn_participation` table: sender, receiver, close address, asset sender, asset receiver, asset close address, freeze account. However, we are missing `Transaction.ApplicationCallTxnFields.Accounts` which are the accounts referenced in an application call. In the future, we might want to construct queries that need this information (e.g., get app call transactions that reference a particular address).

tolikzinovyev:
## Summary

Closes #768.

TODO: don't filter out the zero address. It is a real address and people are curious about how much algos it receives.

## Test Plan

Added a unit test that checks that app call address are written in the `txn_participation` table.

### DIFF
--- a/idb/postgres/internal/writer/writer_test.go
+++ b/idb/postgres/internal/writer/writer_test.go
@@ -410,6 +410,70 @@ func TestWriterTxnParticipationTableBasic(t *testing.T) {
 	}
 }
 
+func TestWriterTxnParticipationTableAppCallAddresses(t *testing.T) {
+	db, shutdownFunc := setupPostgres(t)
+	defer shutdownFunc()
+
+	block := bookkeeping.Block{
+		BlockHeader: bookkeeping.BlockHeader{
+			Round:       basics.Round(2),
+			GenesisID:   test.MakeGenesis().ID(),
+			GenesisHash: test.GenesisHash,
+			UpgradeState: bookkeeping.UpgradeState{
+				CurrentProtocol: test.Proto,
+			},
+		},
+		Payset: make(transactions.Payset, 1),
+	}
+
+	stxnad := test.MakeCreateAppTxn(test.AccountA)
+	stxnad.Txn.ApplicationCallTxnFields.Accounts =
+		[]basics.Address{test.AccountB, test.AccountC}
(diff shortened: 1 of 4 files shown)


idb/postgres/internal/writer/writer_test.go
```diff
@@ -410,6 +410,70 @@ func TestWriterTxnParticipationTableBasic(t *testing.T) {
 	}
 }
 
+func TestWriterTxnParticipationTableAppCallAddresses(t *testing.T) {
+	db, shutdownFunc := setupPostgres(t)
+	defer shutdownFunc()
+
+	block := bookkeeping.Block{
```
winder, line 417: I think you can use `block, err := test.MakeBlockForTxns(test.MakeGenesisBlock().BlockHeader, &stxnad)` instead of  constructing this object yourself.

idb/postgres/internal/writer/writer_test.go
```diff
@@ -410,6 +410,70 @@ func TestWriterTxnParticipationTableBasic(t *testing.T) {
 	}
 }
 
+func TestWriterTxnParticipationTableAppCallAddresses(t *testing.T) {
```
winder, line 413: Could you merge this with `TestWriterTxnParticipationTableBasic`? These tests appear to be identical except for the transactions and expected results, so they could be data-driven with one of those test case structs.

winder (changes requested):
One request otherwise looks fine.

idb/postgres/internal/writer/writer_test.go
tolikzinovyev: done

idb/postgres/internal/writer/writer_test.go
tolikzinovyev: I wanted to use a round number that is not 0 or 1 for the basic test, and now that the tests are merged, I think we should keep the current construct.

winder (approved):
Thanks for the change, looks good.

PR #771, opened 2021-11-04
Support search by zero address.

tolikzinovyev:
## Summary

Zero address is a real address and we shouldn't ignore it. People are curious about transactions in which the zero address participates.


winder (approved):
This looks good, 2 requests for before you merge:

1. Rename the commit with the feature being added so that the release notes are easier to write: `Support search by zero address.`
2. Add comments for each switch case about why zero is only allowed sometimes. I'm pretty sure you got them all correct, but these are subtle details so things comments like "AssetSender is optional, it supposed to be zero unless used in a clawback transaction" may help the next person reading this code.

tolikzinovyev:
Thanks, done. I wonder if we need to check for zero address in `buildTransactionQuery()`. For example, it doesn't make sense to set address to zero and address role to `AddressRoleCloseRemainderTo` in `TransactionFilter`. Should we trust the user to construct the right request?

winder:
> Thanks, done. I wonder if we need to check for zero address in `buildTransactionQuery()`. For example, it doesn't make sense to set address to zero and address role to `AddressRoleCloseRemainderTo` in `TransactionFilter`. Should we trust the user to construct the right request?

In that case, we'd just return no results. Technically correct, right? I don't think we need to bother with custom error messages.

winder (approved):

tolikzinovyev:
> > Thanks, done. I wonder if we need to check for zero address in `buildTransactionQuery()`. For example, it doesn't make sense to set address to zero and address role to `AddressRoleCloseRemainderTo` in `TransactionFilter`. Should we trust the user to construct the right request?
> 
> In that case, we'd just return no results. Technically correct, right? I don't think we need to bother with custom error messages.

No. Suppose the query is Address = zero, Role = CloseAddress. It will find a transaction where receiver is zero and close address is zero. But it will not find a transaction where receiver is non-zero and close address is zero. We could suppose the user should not make such a query. Any thoughts?

winder:
> > > Thanks, done. I wonder if we need to check for zero address in `buildTransactionQuery()`. For example, it doesn't make sense to set address to zero and address role to `AddressRoleCloseRemainderTo` in `TransactionFilter`. Should we trust the user to construct the right request?
> > 
> > 
> > In that case, we'd just return no results. Technically correct, right? I don't think we need to bother with custom error messages.
> 
> No. Suppose the query is Address = zero, Role = CloseAddress. It will find a transaction where receiver is zero and close address is zero. But it will not find a transaction where receiver is non-zero and close address is zero. We could suppose the user should not make such a query. Any thoughts?

I'm pretty this is OK. We use the txn_participation table to identify the initial set of transactions, but then have a second filter on the jsonb to match the specific field we're looking for. I guess it might timeout if there are too many transactions li ke this for the slow jsonb filter.

winder:
@tolikzinovyev I thought about this over the weekend and think the simplest solution is to prevent using AddressRole and a zero address. In `converter_utils.go : transactionParamsToTransactionFilter` this should do it:
```
	// Don't allow address role + zero address search, there are several invalid configurations.
	if params.AddressRole != nil && params.Address != nil {
		var zeroAddr basics.Address
		if zeroAddr.String() == *params.Address {
			errorArr = append(errorArr, errInvalidZeroAddressSearch)
		}
	}
```

winder (approved):
~LGTM~

api/handlers.go
```diff
@@ -487,6 +487,15 @@ func (si *ServerImplementation) SearchForTransactions(ctx echo.Context, params g
...
 
+	if filter.AddressRole != 0 {
+		var address basics.Address
+		copy(address[:], filter.Address)
+		if address.IsZero() {
+			return badRequest(
+				ctx, "searching transactions by zero address with address role is not supported")
+		}
+	}
+
```
winder, line 498: This should go into `transactionParamsToTransactionFilter` or `si.fetchTransactions` to make sure the constraint is checked for all account searches.

winder (changes requested):
I think the parameter checking should go in a different spot

api/handlers.go
tolikzinovyev: Only `LookupApplicationLogsByID()` would violate, which I think is fine. If we do put it in `fetchTransactions()`, `LookupApplicationLogsByID()` would return a weird error.

tolikzinovyev: `transactionParamsToTransactionFilter()` is a conversion function, I don't want to put business logic there.

I think the current code is ok, especially given that no one in their right mind will use `LookupApplicationLogsByID()` with zero address.

winder: It also validates the input, there's a test in `api/handlers_test.go` for it also so you can add a test:
```
    {
      name: "Invalid zero address parameters",
      params: generated.SearchForTransactionsParams{
        Address:     strPtr(basics.Address{}.String()),
        AddressRole: 9,
      },
      filter:        idb.TransactionFilter{},
      errorContains: []string{errInvalidZeroAddressSearch},
    }
```

Could you also add the error message to `error_messages.go`?

tolikzinovyev: Moved the error message to `error_messages.go`. `transactionParamsToTransactionFilter()` does a couple of sanity checks, but no more. Definitely no artificial limitations should be in there.

tolikzinovyev: I think it is incorrect that https://github.com/algorand/indexer/blob/830e1ba83500b2c78251070fa67031efdd418ed7/api/handlers.go#L277 becomes `idb.AddressRoleSender | idb.AddressRoleAssetSender`, so I set `filter.AddressRole` directly and run validation that used to be in `transactionParamsToTransactionFilter()` after.

api/handlers.go
```diff
@@ -274,13 +315,18 @@ func (si *ServerImplementation) LookupApplicationLogsByID(ctx echo.Context, appl
...
 		MaxRound:      params.MaxRound,
 		Address:       params.SenderAddress,
-		AddressRole:   strPtr(addrRoleSender),
 	}
 
 	filter, err := transactionParamsToTransactionFilter(searchParams)
 	if err != nil {
 		return badRequest(ctx, err.Error())
 	}
+	filter.AddressRole = idb.AddressRoleSender
```
winder, line 324: Nice :+1:

winder (approved):
LGTM

PR #772, opened 2021-11-05
Delete `txnbytes` column

### LINKED ISSUE #767: Delete `txnbytes` column
### Subject of the issue

The `txn` table has two columns in addition to others: `txnbytes` and `txn`. The first encodes a transaction in msgpack and the second one in json. We need json encoding to be able to construct sql queries that access individual transaction fields. Presumably, msgpack encoding was added to reduce the risk of encoding / decoding issues for accounting, since import used to be done in two phases: 1) write transactions to the database, 2) read transactions and do accounting.

We now do import in one phase, so this concern is not applicable. To improve import performance, reduce the size of the database, and simplify the code, we should delete the msgpack encoding.

tolikzinovyev:
## Summary

Closes #767.

## Test Plan

Relying on existing tests.


idb/postgres/postgres.go
```diff
@@ -790,25 +790,41 @@ func (db *IndexerDb) yieldTxnsThreadSimple(ctx context.Context, rows pgx.Rows, r
 		var round uint64
 		var asset uint64
 		var intra int
-		var txnbytes []byte
-		var roottxnbytes []byte
+		var txn []byte
+		var roottxn []byte
 		var extraJSON []byte
```
winder, line 795: Since these are all the same jsonb column type, could you rename `extraJSON` to `extra`?

idb/postgres/postgres.go
```diff
@@ -790,25 +790,41 @@ func (db *IndexerDb) yieldTxnsThreadSimple(ctx context.Context, rows pgx.Rows, r
...
+				}
+			} else {
+				// Root transaction.
+				row.Txn = new(transactions.SignedTxnWithAD)
+				*row.Txn, err = encoding.DecodeSignedTxnWithAD(txn)
+				if err != nil {
+					err = fmt.Errorf("error decoding txn, err: %w", err)
+					row.Error = err
+				}
+			}
```
winder, line 820: could you add a test for these deserialization errors?

idb/postgres/postgres.go
```diff
@@ -790,25 +790,41 @@ func (db *IndexerDb) yieldTxnsThreadSimple(ctx context.Context, rows pgx.Rows, r
...
-			row.RootTxnBytes = roottxnbytes
+			if roottxn != nil {
+				// Inner transaction.
+				row.RootTxn = new(transactions.SignedTxnWithAD)
+				*row.RootTxn, err = encoding.DecodeSignedTxnWithAD(roottxn)
+				if err != nil {
+					err = fmt.Errorf("error decoding roottxn, err: %w", err)
+					row.Error = err
+				}
+			} else {
```
winder, line 812: nit: this would make it easier to support returning inner transactions in the future.
```suggestion
			}
			if {
```

winder (commented):
This is much more straightforward than I was expecting, but changing the encoding right before the release makes me nervous. I'm thinking we should make another integration test tool to query all of the transactions and make sure we're able to decode them.

[From the documentation](https://www.postgresql.org/docs/13/sql-altertable.html):
> The DROP COLUMN form does not physically remove the column, but simply makes it invisible to SQL operations. Subsequent insert and update operations in the table will store a null value for the column. Thus, dropping a column is quick but it will not immediately reduce the on-disk size of your table, as the space occupied by the dropped column is not reclaimed. The space will be reclaimed over time as existing rows are updated.
>
>To force immediate reclamation of space occupied by a dropped column, you can execute one of the forms of ALTER TABLE that performs a rewrite of the whole table. This results in reconstructing each row with the dropped column replaced by a null value.

According to this, I think it would be OK to drop this column later and tell people that a full rebuild will be more space-efficient. But it's weird that your DB jumped to 100% utilization... so maybe there is something else going on.

gmalouf:
I'd like us to put this on the sideline and add in a follow-up after
extensive testing.

tolikzinovyev:
> This is much more straightforward than I was expecting, but changing the encoding right before the release makes me nervous. I'm thinking we should make another integration test tool to query all of the transactions and make sure we're able to decode them.
> 
> [From the documentation](https://www.postgresql.org/docs/13/sql-altertable.html):
> 
> > The DROP COLUMN form does not physically remove the column, but simply makes it invisible to SQL operations. Subsequent insert and update operations in the table will store a null value for the column. Thus, dropping a column is quick but it will not immediately reduce the on-disk size of your table, as the space occupied by the dropped column is not reclaimed. The space will be reclaimed over time as existing rows are updated.
> > To force immediate reclamation of space occupied by a dropped column, you can execute one of the forms of ALTER TABLE that performs a rewrite of the whole table. This results in reconstructing each row with the dropped column replaced by a null value.
> 
> According to this, I think it would be OK to drop this column later and tell people that a full rebuild will be more space-efficient. But it's weird that your DB jumped to 100% utilization... so maybe there is something else going on.

I think, in that case we can indeed delay this.

idb/postgres/postgres.go
tolikzinovyev: done

idb/postgres/postgres.go
tolikzinovyev: Just tried. Postgres doesn't let me import bad json.

winder (commented):
Changes look great.

Like Tolik mentioned we have special handling to record the type of signature being used, we need to make sure these changes don't cause a regression in that logic. How are these keys used?

Currently, the logic is simple. We override the accounts sig type in the DB with whatever is used to sign the transaction. Do we need to skip that step when something is signed with the state proof?

tolikzinovyev:
> Changes look great.
> 
> Like Tolik mentioned we have special handling to record the type of signature being used, we need to make sure these changes don't cause a regression in that logic. How are these keys used?
> 
> Currently, the logic is simple. We override the accounts sig type in the DB with whatever is used to sign the transaction. Do we need to skip that step when something is signed with the state proof?

How is sigtype related to this PR?

winder (approved):
LGTM

PR #775, opened 2021-11-05
Import validation tool implementation

tolikzinovyev:
## Summary

This is import validation tool I used to catch the recent accounting bug. I think it's a good idea to run this continuously to learn fast about any other accounting issues.

## Test Plan

It indeed found a discrepancy for 2.7.1.


winder (approved):
Very cool :+1:

PR #778, opened 2021-11-08
Support StateProof keys.

shiqizng:
## Summary

Add new field, StateProofKey, to key reg txn fields and account participation. Accounts data written to postgres is also updated accordingly to include the additional field. 

## Test Plan
Add new unit and e2e tests for api handler to validate that StateProofKey is returned correctly in response.


jasonpaulos (commented):
Generally looks good to me, though I can't give great feedback on some indexer-specific testing changes, like if it's ok to modify `createTxn`.

Also note that the go-algorand submodule was updated to use the branch from https://github.com/algorand/go-algorand/pull/2990, so probably that needs to be merged and the submodule changed back to master (or stable?) before this PR should be merged

api/handlers_e2e_test.go
```diff
@@ -331,3 +425,5 @@ func TestVersion(t *testing.T) {
 	// This is weird looking because the version is set with -ldflags
 	require.Equal(t, response.Version, "(unknown version)")
 }
+
```
tolikzinovyev, line 428: please delete empty lines

api/handlers_e2e_test.go
```diff
@@ -300,6 +302,98 @@ func TestPagingRootTxnDeduplication(t *testing.T) {
...
+	// When // We query the txn
+	//////////
+
+	e := echo.New()
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	rec := httptest.NewRecorder()
+	c := e.NewContext(req, rec)
+	c.SetPath("/v2/transactions/:txid")
+	api := &ServerImplementation{db: db}
+	//params := generated.LookupAccountByIDParams{}
```
tolikzinovyev, line 359: please delete

api/handlers_e2e_test.go
```diff
@@ -300,6 +302,98 @@ func TestPagingRootTxnDeduplication(t *testing.T) {
...
+	//////////
+	// Then // The key reg txn response has state proof key
+	//////////
+
+	var response generated.TransactionResponse
+	data := rec.Body.Bytes()
+	err = json.Decode(data, &response)
+	require.NoError(t, err)
+	require.NotNil(t, response.Transaction.KeyregTransaction.StateProofKey)
+	require.Equal(t, json.Encode(sprfkey), json.Encode(*response.Transaction.KeyregTransaction.StateProofKey))
```
tolikzinovyev, line 373: This is interesting. Is there a reason to compare json encodings instead of byte arrays?

api/handlers_e2e_test.go
```diff
@@ -300,6 +302,98 @@ func TestPagingRootTxnDeduplication(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = db.AddBlock(&block)
+	require.NoError(t, err, "failed to commit")
+
+	//////////
+	// When // We query the txn
+	//////////
+
+	e := echo.New()
```
tolikzinovyev, line 353: I would put this and next section in its own scope so that the section after doesn't accidentally use any of the old variables.
```
{
  req := httptest.NewRequest(http.MethodGet, "/", nil)
  ...
  require.Equal(t, json.Encode(sprfkey), json.Encode(*response.Transaction.KeyregTransaction.StateProofKey))
}
```

api/handlers_e2e_test.go
```diff
@@ -300,6 +302,98 @@ func TestPagingRootTxnDeduplication(t *testing.T) {
...
+			Txn: transactions.Transaction{
+				Type: "keyreg",
+				Header: transactions.Header{
+					Sender:      test.AccountA,
+					GenesisHash: test.GenesisHash,
+				},
+				KeyregTxnFields: transactions.KeyregTxnFields{
+					VotePK: votePK,
+					SelectionPK:selectionPK,
+					StateProofPK: sprfkey,
```
tolikzinovyev, line 332: It would be nice to rename `sprfkey` to `stateProofPK` for consistency with `votePK` and `selectionPK`.

tolikzinovyev (changes requested):
Thanks for the PR. The code looks good overall, but the compact cert transaction type needs to be added in `idb/txn_type_enum.go` as discussed.

shiqizng:
> Thanks for the PR. The code looks good overall, but the compact cert transaction type needs to be added in `idb/txn_type_enum.go` as discussed.

compact cert will be part of a different release. I was told to include keyreg changes only.

api/handlers_e2e_test.go
shiqizng: updated.

api/handlers_e2e_test.go
shiqizng: sprfKey type merklekeystore.Verifier and response.Transaction.KeyregTransaction.StateProofKey is deserialized to *byte[].  This check fails if I don't convert them to the same type.

jasonpaulos: Casting both to `[]byte` would be a simpler option

shiqizng: it seems that I'm only allowed to cast sprfKey to [64]byte because merklekeystore.Verifier is a type alias for [64]byte.

jasonpaulos: Oh good point. Will `sprfkey[:]` to convert to a slice?

shiqizng: updated.

tolikzinovyev (commented):
LGTM. We will need to wait for state proof changes to arrive in go-algorand before merging.

winder (approved):
LGTM. But please don't merge it in yet, we're still finalizing a release.

api/handlers_test.go
```diff
@@ -571,47 +582,34 @@ func TestFetchAccountsRewindRoundTooLarge(t *testing.T) {
 func createTxn(t *testing.T, target string) []byte {
```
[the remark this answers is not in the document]
algoidan: Adding this thing makes the CI fail.
do you have a different suggestion for that?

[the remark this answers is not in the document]
winder: Why is this function being called? It looks like a helper for adding a new test.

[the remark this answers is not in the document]
algoidan: yes
the idea is to keep the code that generates the test_resource.
I couldn't find more functions like this. Do you think it is worth keeping? (not as a Test function?)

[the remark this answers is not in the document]
winder: Yes, it is worth keeping.

`addNewTest` above is set to false, so this function is never called. It should never be called in a finished test which is why I suggested putting `assert.Fail` in here.

tolikzinovyev (approved):

PR #783, opened 2021-11-12
Update sigtype along with the rest of account data

tolikzinovyev:
## Summary

Before we were updating the `account.keytype` column separately from updating accounts which resulted in one extra account lookup and one extra account rewrite by the database. This PR changes the way sigtype is updated; now it's done with the rest of account changes. The TPS improvement is 1.6X in my tests.

## Test Plan

We already have tests for the keytype, but I added another one that tests that even when an account is deleted after a round, the keytype is updated anyway.

One concern I had is that the evaluator might skip an account delta when the account data doesn't change even when it is a transaction sender (which triggers keytype change), but we already have a test (`TestKeytypeBasic`) that uses zero fee transactions that don't modify the account state.

### DIFF
--- a/idb/postgres/internal/writer/writer.go
+++ b/idb/postgres/internal/writer/writer.go
@@ -20,20 +20,19 @@ import (
 )
 
 const (
-	addBlockHeaderStmtName       = "add_block_header"
-	setSpecialAccountsStmtName   = "set_special_accounts"
-	upsertAssetStmtName          = "upsert_asset"
-	upsertAccountAssetStmtName   = "upsert_account_asset"
-	upsertAppStmtName            = "upsert_app"
-	upsertAccountAppStmtName     = "upsert_account_app"
-	deleteAccountStmtName        = "delete_account"
-	upsertAccountStmtName        = "upsert_account"
-	deleteAssetStmtName          = "delete_asset"
-	deleteAccountAssetStmtName   = "delete_account_asset"
-	deleteAppStmtName            = "delete_app"
-	deleteAccountAppStmtName     = "delete_account_app"
-	updateAccountKeyTypeStmtName = "update_account_key_type"
-	updateAccountTotalsStmtName  = "update_account_totals"
+	addBlockHeaderStmtName      = "add_block_header"
+	setSpecialAccountsStmtName  = "set_special_accounts"
+	upsertAssetStmtName         = "upsert_asset"
+	upsertAccountAssetStmtName  = "upsert_account_asset"
+	upsertAppStmtName           = "upsert_app"
+	upsertAccountAppStmtName    = "upsert_account_app"
+	deleteAccountStmtName       = "delete_account"
+	upsertAccountStmtName       = "upsert_account"
+	deleteAssetStmtName         = "delete_asset"
+	deleteAccountAssetStmtName  = "delete_account_asset"
+	deleteAppStmtName           = "delete_app"
+	deleteAccountAppStmtName    = "delete_account_app"
+	updateAccountTotalsStmtName = "update_account_totals"
 )
 
 var statements = map[string]string{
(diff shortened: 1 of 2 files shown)


idb/postgres/internal/writer/writer.go
```diff
@@ -364,7 +364,30 @@ func (w *Writer) addTransactionParticipation(block *bookkeeping.Block) error {
 	return nil
 }
 
-func writeAccountData(round basics.Round, address basics.Address, accountData basics.AccountData, batch *pgx.Batch) {
+type optionalSigType struct {
```
winder, line 367: This type isn't needed, a `map[Address]SigType` can be used. On a rekey you could `delete(res, payset[i].Txn.Sender)`, the present check would change to compare sigtype to the zero value (or check if it is in the map).

idb/postgres/internal/writer/writer.go
```diff
@@ -364,7 +364,30 @@ func (w *Writer) addTransactionParticipation(block *bookkeeping.Block) error {
...
+			res[payset[i].Txn.Sender] = optionalSigType{present: true, value: sigtype}
+		} else {
+			res[payset[i].Txn.Sender] = optionalSigType{}
+		}
+	}
+
+	return res, nil
+}
+
+func writeAccount(round basics.Round, address basics.Address, accountData basics.AccountData, sigTypeDelta optionalSigType, batch *pgx.Batch) {
```
winder, line 390: Nice :+1:

winder (approved):
Looks good. Just a small nit about how we could avoid adding the new optional type

idb/postgres/internal/writer/writer.go
```diff
@@ -61,16 +60,18 @@ var statements = map[string]string{
...
 	deleteAccountStmtName: `INSERT INTO account
 		(addr, microalgos, rewardsbase, rewards_total, deleted, created_at, closed_at,
-			account_data)
-		VALUES($1, 0, 0, 0, TRUE, $2, $2, 'null'::jsonb) ON CONFLICT (addr) DO UPDATE SET
+		 keytype, account_data)
+		VALUES($1, 0, 0, 0, TRUE, $2, $2, $3, 'null'::jsonb) ON CONFLICT (addr) DO UPDATE SET
 		microalgos = EXCLUDED.microalgos, rewardsbase = EXCLUDED.rewardsbase,
 		rewards_total = EXCLUDED.rewards_total, deleted = TRUE,
-		closed_at = EXCLUDED.closed_at, account_data = EXCLUDED.account_data`,
+		closed_at = EXCLUDED.closed_at, keytype = EXCLUDED.keytype,
```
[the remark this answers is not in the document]
tolikzinovyev: Right, this sql query will not clear it.

idb/postgres/internal/writer/writer.go
tolikzinovyev: I think that wouldn't work. We need to distinguish between three operations: leave sigtype as is, clear it, or set to something non-null. If an address is not in `map[basics.Address]optionalSigType`, we leave sigtype as is. If it is in the map and the value is `optionalSigType` with `present = false`, we clear it. Otherwise, we set sigtype to `optionalSigType.value`.

winder: You aren't using the 3rd case, this code sets something for each sender in the payset, even if it doesn't change.

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
winder: There are some weird edge cases. For example if you tried to rekey an account that was being deleted, you would skip clearing the sig type, unless it already had an auth-addr in which case it would still need to be cleared.

I don't see the code that checks for auth-addr and conditionally clears it.

idb/postgres/internal/writer/writer.go
tolikzinovyev: you mean the first case?

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
tolikzinovyev: I think there is no problem. Here is where clearing happens. https://github.com/algorand/indexer/pull/783/files#diff-940e0e8ce1ffe2b07a3c71d6bb32f7b74b918d0c9d4661d9b3fb339dc5b11b18R383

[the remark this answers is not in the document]
winder: There are 3 cases:
1) `RekeyTo` is set, we must clear the sigtype field.
2) `RekeyTo` is not set, sigtype is not set, we must set the sigtype field.
3) `RekeyTo` is not set, sigtype is set, setting the sigtype field is not necessary.

The 3 states can be used to handle each of these cases individually, but currently the 3rd is not triggering any different behavior.

idb/postgres/internal/writer/writer.go
```diff
@@ -364,7 +379,35 @@ func (w *Writer) addTransactionParticipation(block *bookkeeping.Block) error {
...
+			res[payset[i].Txn.Sender] = sigTypeDelta{present: true, value: sigtype}
+		} else {
+			res[payset[i].Txn.Sender] = sigTypeDelta{}
+		}
+	}
+
+	return res, nil
+}
+
+type optionalSigTypeDelta struct {
```
winder, line 405: This nested optional implementation is really confusing. Even knowing the problem it solves I still had a lot of trouble following how it works.

I think the code would be easier to understand with a single optional type (the one named `sigTypeDelta`), and a check for `value.IsZero()`. That way you either set the value to zero, or skip setting it.



 is more complicated than the `writeAccounts` function you replaced with it. I think it would be better to have a single optional and check `value.IsZero()` for

tolikzinovyev: The reason why I did it this way is because `idb.SigType` doesn't have a special zero value. I think it shouldn't. I will add a comment for `sigTypeDelta`.

idb/postgres/internal/writer/writer.go
```diff
@@ -379,6 +379,8 @@ func (w *Writer) addTransactionParticipation(block *bookkeeping.Block) error {
 	return nil
 }
 
+// Describes a change to the `account.keytype` column. If `present` is true,
+// `value` is the new value. Otherwise, NULL will be the new value.
```
winder, line 383: Thanks. This is still pretty confusing, but the comment helps. We probably should have added a `UKNOWN` SigType, but this works.

winder (approved):

PR #784, opened 2021-11-15
Return `TotalAppSchema` and `TotalExtraAppPages` from `AccountData`

tolikzinovyev:
## Summary

To make the output of the accounts query consistent with that of algod, return `TotalAppSchema` and `TotalExtraAppPages` directly from `AccountData`.

## Test Plan

Tested manually.


winder (approved):
lgtm

PR #793, opened 2021-11-23
Updating README.MD with CircleCI status badges

algojack:
we moved our builds to circleci, so updating our badges too


winder (approved):
Thanks!

PR #797, opened 2021-11-24
Client generators produce int instead of long without OpenAPI integer…

FrankSzendzielarz:
… format hint.

## Summary

## Test Plan


winder:
Closing this PR, after further discussions with Frank the issue is a bit larger than this small patch and we'd like to find a way to collaborate and address the problem head-on.

PR #801, opened 2021-11-29
Add `IndexerDb.Close()`

tolikzinovyev:
## Summary

When running test with the `--test-pg` flag, we open many database connections without closing them. This leads to the database refusing new connections.

This PR adds `IndexerDb.Close()` function that closes all active connections to the database.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -83,8 +84,13 @@ var daemonCmd = &cobra.Command{
 			opts.ReadOnly = true
 		}
 		db, availableCh := indexerDbFromFlags(opts)
+		defer db.Close()
+		var wg sync.WaitGroup
 		if bot != nil {
+			wg.Add(1)
 			go func() {
+				defer wg.Done()
+
 				// Wait until the database is available.
 				<-availableCh
 
@@ -103,7 +109,6 @@ var daemonCmd = &cobra.Command{
 
 				logger.Info("Starting block importer.")
 				bot.Run()
-				cf()
 			}()
 		} else {
 			logger.Info("No block importer configured.")
@@ -112,6 +117,7 @@ var daemonCmd = &cobra.Command{
 		fmt.Printf("serving on %s\n", daemonServerAddr)
 		logger.Infof("serving on %s", daemonServerAddr)
 		api.Serve(ctx, daemonServerAddr, db, bot, logger, makeOptions())
+		wg.Wait()
 	},
 }
 
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -103,12 +103,21 @@ var daemonCmd = &cobra.Command{
 				maybeFail(err, "failed to get next round, %v", err)
 				bot.SetNextRound(nextRound)
 
-				bih := blockImporterHandler{imp: importer.NewImporter(db)}
-				bot.AddBlockHandler(&bih)
-				bot.SetContext(ctx)
+				imp := importer.NewImporter(db)
+				handler := func(ctx context.Context, block *rpcs.EncodedBlockCert) error {
+					return handleBlock(block, &imp)
+				}
+				bot.SetBlockHandler(handler)
 
 				logger.Info("Starting block importer.")
-				bot.Run()
+				err = bot.Run(ctx)
+				if err != nil {
+					// If context is not expired.
+					if ctx.Err() == nil {
+						logger.WithError(err).Errorf("fetcher exited with error")
+						os.Exit(1)
+					}
+				}
 			}()
 		} else {
 			logger.Info("No block importer configured.")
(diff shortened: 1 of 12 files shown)


cmd/algorand-indexer/daemon.go
```diff
@@ -103,7 +109,6 @@ var daemonCmd = &cobra.Command{
 
 				logger.Info("Starting block importer.")
 				bot.Run()
-				cf()
```
winder, line 106: There's another one on line 64 that can be removed.

winder (approved):

cmd/algorand-indexer/daemon.go
tolikzinovyev: We need that one to cancel the context when Indexer is politely killed.

winder: Gotcha, calling cf there causes `bot.Run` and `api.Serve` to terminate on an exit signal.

fetcher/fetcher.go
```diff
@@ -222,36 +198,44 @@ func (bot *fetcherImpl) Run() {
 	}
 }
 
-// SetContext is part of the Fetcher interface
-func (bot *fetcherImpl) SetContext(ctx context.Context) {
-	bot.ctx = ctx
+// Run is part of the Fetcher interface
+func (bot *fetcherImpl) Run(ctx context.Context) error {
+	// In theory a buffer of size one should be enough, but let's make it bigger.
+	bot.blockQueue = make(chan *rpcs.EncodedBlockCert, 5)
```
winder, line 204:
```suggestion
	bot.blockQueue = make(chan *rpcs.EncodedBlockCert, 1)
```

fetcher/fetcher.go
```diff
@@ -222,36 +198,44 @@ func (bot *fetcherImpl) Run() {
...
+
+	ch1 := make(chan error)
+	go func() {
+		ch1 <- bot.mainLoop(ctx)
+	}()
+
+	select {
+	case err := <-ch0:
+		cancelFunc()
+		<-ch1
```
winder, line 222: Why do we need to wait for the other goroutine to finish before exiting? I think you can just return the errors.

fetcher/fetcher.go
```diff
@@ -222,36 +198,44 @@ func (bot *fetcherImpl) Run() {
...
+		return fmt.Errorf("Run() err: %w", err)
+	case err := <-ch1:
+		cancelFunc()
+		<-ch0
+		return fmt.Errorf("Run() err: %w", err)
```
winder, line 227: nit: Disambiguate the errors, more descriptive channel names would also be helpful.
```suggestion
		return fmt.Errorf("Run() processQueue err: %w", err)
	case err := <-ch1:
		cancelFunc()
		<-ch0
		return fmt.Errorf("Run() mainLoop err: %w", err)
```

winder (changes requested):
Looks good, just a couple small things to address.

fetcher/fetcher.go
tolikzinovyev: I guess here we don't. In the other case we do because we want to finish database operations before closing the database.

fetcher/fetcher.go
tolikzinovyev: the `err` objects will tell where they come from

cmd/algorand-indexer/daemon.go
```diff
@@ -97,13 +103,21 @@ var daemonCmd = &cobra.Command{
...
 				bot.SetNextRound(nextRound)
 
-				bih := blockImporterHandler{imp: importer.NewImporter(db)}
-				bot.AddBlockHandler(&bih)
-				bot.SetContext(ctx)
+				imp := importer.NewImporter(db)
+				handler := func(ctx context.Context, block *rpcs.EncodedBlockCert) error {
+					return handleBlock(block, &imp)
+				}
+				bot.SetBlockHandler(handler)
```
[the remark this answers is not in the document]
tolikzinovyev: Why? Interface objects are harder to construct at the user side.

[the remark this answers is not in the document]
winder: The terseness mainly, you could create/set it in one line if you wanted. Here you must create the importer object separately.

fetcher/fetcher.go
winder: Sure, we could still change it.

fetcher/fetcher.go
winder: > the err objects ~will~ **should** tell where they come from

tolikzinovyev: But they do. You can check `mainLoop()` and `processQueue()`.

fetcher/fetcher.go
tolikzinovyev: sure

cmd/algorand-indexer/daemon.go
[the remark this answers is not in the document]
tolikzinovyev: I'm not sure about terseness. With a closure, looking at the function signature, you immediately see what the closure is. Plus, an interface is a set of functions. That's redundant.

winder (approved):
Looks good

tolikzinovyev:
Thanks!

PR #805, opened 2021-12-06
Write into `txn` and `txn_participation` tables in parallel with other import procedures

### LINKED ISSUE #795: Import into `txn` and `txn_participation` concurrently with other db operations
Prototype implemented in 23b184ed63af0130a667b023727622b09887607a. Gives about 18% boost in TPS.

tolikzinovyev:
## Summary

`AddTransactions()` and `AddTransactionParticipation()` are taken out of the writer struct and made free functions, so that they can be called with a different database transaction in a different goroutine. This code performs at 9200 TPS in my tests. The current develop branch performs at 7000 TPS.

`AddTransactions()` and `AddTransactionParticipation()` are called before the main db transaction commits, so `txn` and `txn_participation` tables can only be ahead of all other state. Extra care was taken to ensure that this desynchronization is not a problem. First, if `idb.AddBlock()` sees postgres unique violation error when writing to `txn` and `txn_participation` tables, it ignores this error since it means that these tables are ahead. Second, we need to make sure that Indexer query code works fine. At the moment only `idb.Transactions()` uses those two tables. Since they are JOINed with the `block_header` table, rows in `txn` and `txn_participation` that are ahead will be filtered out. In the future, if we need to add more queries that use these two tables, we could use database views such as `SELECT * FROM txn WHERE round < $1`.

Closes #795.

## Test Plan

Added a test that checks that `idb.AddBlock()` runs successfully when `txn` and `txn_participation` are ahead.
Added a test checking that `idb.AddBlock()` does call `AddTransactionParticipation()`.
Added a test checking that when `txn` and `txn_participation` tables are ahead, `idb.Transactions()` behaves as if they were not.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -7,6 +7,7 @@ package postgres
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -21,6 +22,8 @@ import (
 	"github.com/algorand/go-algorand/ledger/ledgercore"
 	"github.com/algorand/go-algorand/protocol"
 
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
 	log "github.com/sirupsen/logrus"
@@ -275,20 +278,35 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			return fmt.Errorf("AddBlock() err: %w", err)
 		}
 
-		writer, err := writer.MakeWriter(tx)
+		w, err := writer.MakeWriter(tx)
 		if err != nil {
 			return fmt.Errorf("AddBlock() err: %w", err)
 		}
-		defer writer.Close()
+		defer w.Close()
 
 		if block.Round() == basics.Round(0) {
 			// Block 0 is special, we cannot run the evaluator on it.
-			err := writer.AddBlock0(block)
+			err := w.AddBlock0(block)
 			if err != nil {
 				return fmt.Errorf("AddBlock() err: %w", err)
 			}
 		} else {
-			ledgerForEval, err := ledger_for_evaluator.MakeLedgerForEvaluator(tx, block.Round()-1)
+			var wg sync.WaitGroup
+			defer wg.Wait()
+
+			var err0 error
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				f := func(tx pgx.Tx) error {
+					return writer.AddTransactionParticipation(block, tx)
+				}
+				err0 = db.txWithRetry(serializable, f)
+			}()
+
+			ledgerForEval, err :=
+				ledger_for_evaluator.MakeLedgerForEvaluator(tx, block.Round()-1)
 			if err != nil {
 				return fmt.Errorf("AddBlock() err: %w", err)
 			}
@@ -314,9 +332,32 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			}
 			metrics.PostgresEvalTimeSeconds.Observe(time.Since(start).Seconds())
 
-			err = writer.AddBlock(block, modifiedTxns, delta)
+			var err1 error
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				f := func(tx pgx.Tx) error {
+					return writer.AddTransactions(block, modifiedTxns, tx)
+				}
+				err1 = db.txWithRetry(serializable, f)
+			}()
+
+			err = w.AddBlock(block, modifiedTxns, delta)
 			if err != nil {
 				return fmt.Errorf("AddBlock() err: %w", err)
+			}
+
+			wg.Wait()
+			isUniqueViolationFunc := func(err error) bool {
+				var pgerr *pgconn.PgError
+				return errors.As(err, &pgerr) && (pgerr.Code == pgerrcode.UniqueViolation)
+			}
+			if (err0 != nil) && !isUniqueViolationFunc(err0) {
+				return fmt.Errorf("AddBlock() err0: %w", err0)
+			}
+			if (err1 != nil) && !isUniqueViolationFunc(err1) {
+				return fmt.Errorf("AddBlock() err1: %w", err1)
 			}
 		}
 
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -294,6 +294,7 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			var wg sync.WaitGroup
 			defer wg.Wait()
 
+			// Write transaction participation in a parallel db transaction.
 			var err0 error
 			wg.Add(1)
 			go func() {
@@ -332,6 +333,7 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 			}
 			metrics.PostgresEvalTimeSeconds.Observe(time.Since(start).Seconds())
 
+			// Write transactions in a parallel db transaction.
 			var err1 error
 			wg.Add(1)
 			go func() {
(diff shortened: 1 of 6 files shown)


winder:
@tolikzinovyev Could you move things back to `writer.go` so that it's easier to see what changed? Next time if you do this sort of change, using 2 commits would work. That way I can look at one commit for the refactoring, and the other for the code changes.

tolikzinovyev:
> @tolikzinovyev Could you move things back to `writer.go` so that it's easier to see what changed? Next time if you do this sort of change, using 2 commits would work. That way I can look at one commit for the refactoring, and the other for the code changes.

I didn't change any code there, only moved it.

idb/postgres/postgres.go
```diff
@@ -275,20 +278,35 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
+
+			var err0 error
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				f := func(tx pgx.Tx) error {
+					return writer.AddTransactionParticipation(block, tx)
+				}
+				err0 = db.txWithRetry(serializable, f)
```
winder, line 305: The nested tx with retry is a little strange. Could you call `writer.AddTransactionParticipation` directly?

idb/postgres/postgres.go
```diff
@@ -314,10 +332,33 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
+			isUniqueViolationFunc := func(err error) bool {
+				var pgerr *pgconn.PgError
+				return errors.As(err, &pgerr) && (pgerr.Code == pgerrcode.UniqueViolation)
+			}
+			if (err0 != nil) && !isUniqueViolationFunc(err0) {
+				return fmt.Errorf("AddBlock() err0: %w", err0)
+			}
+			if (err1 != nil) && !isUniqueViolationFunc(err1) {
+				return fmt.Errorf("AddBlock() err1: %w", err1)
+			}
```
winder, line 361: More descriptive error messages. Also above `s/err/block header`
```suggestion
			if (err0 != nil) && !isUniqueViolationFunc(err0) {
				return fmt.Errorf("AddBlock() participation: %w", err0)
			}
			if (err1 != nil) && !isUniqueViolationFunc(err1) {
				return fmt.Errorf("AddBlock() transactions: %w", err1)
			}
```

idb/postgres/postgres.go
winder: Why is this in a go routine, doesn't the evaluator need the txn participation in order to prefetch accounts?

tolikzinovyev: It would be more strange to retry the main transaction if this one fails.

Prefetching happens separately 15 lines below.

idb/postgres/postgres.go
tolikzinovyev: `err0` and `err1` will tell where they come from. The pattern I've been using is `variablename: variablevalue`.

tolikzinovyev:
I agree that the interface is not pretty anymore. Suggestions are welcome.

I run `config.payment.jumbo.yml` for one hour.

idb/postgres/postgres.go
```diff
@@ -314,10 +332,33 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
+
+			err = w.AddBlock(block, modifiedTxns, delta)
 			if err != nil {
 				return fmt.Errorf("AddBlock() err: %w", err)
 			}
+
+			wg.Wait()
+			isUniqueViolationFunc := func(err error) bool {
+				var pgerr *pgconn.PgError
+				return errors.As(err, &pgerr) && (pgerr.Code == pgerrcode.UniqueViolation)
```
winder, line 354: Why is a `UniqueViolation` OK?

idb/postgres/postgres_integration_test.go
```diff
@@ -1540,13 +1540,23 @@ func TestSearchForInnerTransactionReturnsRootTransaction(t *testing.T) {
...
+		}
+
+		w, err := writer.MakeWriter(tx)
+		if err != nil {
+			return err
+		}
+		defer w.Close()
 
-		return nil
+		return w.AddBlock(&block, block.Payset, ledgercore.StateDelta{})
```
winder, line 1559: Can this can use db.AddBlock instead of w.AddBlock to simplify the test?

idb/postgres/postgres_integration_test.go
```diff
@@ -1783,3 +1793,93 @@ func TestKeytypeDoNotResetReceiver(t *testing.T) {
 	assertKeytype(t, db, test.AccountA, &keytype)
 	assertKeytype(t, db, test.AccountB, &keytype)
 }
+
+// Test that if information in `txn` and `txn_participation` tables is ahead of
+// the current round, AddBlock() still runs successfully.
+func TestAddBlockTxnTxnParticipationAhead(t *testing.T) {
```
winder, line 1799: How about the "Behind" case? Previously txn and txn_participation were always added first, so they were kind of always ahead.

idb/postgres/postgres_integration_test.go
```diff
@@ -1783,3 +1793,93 @@ func TestKeytypeDoNotResetReceiver(t *testing.T) {
...
+			VALUES ($1, 1, 0)`
+		_, err := db.db.Exec(context.Background(), query, test.AccountA[:])
+		require.NoError(t, err)
+	}
+
+	txn := test.MakePaymentTxn(
+		0, 0, 0, 0, 0, 0, test.AccountA, test.AccountA, basics.Address{}, basics.Address{})
+	block, err := test.MakeBlockForTxns(block.BlockHeader, &txn)
+	require.NoError(t, err)
+	err = db.AddBlock(&block)
```
winder, line 1821: Internally, this would be hitting/ignoring the UniqueViolation case?

idb/postgres/postgres_integration_test.go
```diff
@@ -1783,3 +1793,93 @@ func TestKeytypeDoNotResetReceiver(t *testing.T) {
...
+	row, ok := <-rowsCh
+	require.True(t, ok)
+	require.NoError(t, row.Error)
+	require.NotNil(t, row.Txn)
+	assert.Equal(t, txn, *row.Txn)
+}
+
+// Test that if information in the `txn` table is ahead of the current round,
+// Transactions() doesn't return the rows ahead of the state.
+func TestTransactionsTxnAhead(t *testing.T) {
```
winder, line 1852: How do these cases actually happening? Are the txn/txn_participation being added in a separate transaction? I thought it was in a nested transaction, but the test suggests otherwise.

idb/postgres/postgres.go
tolikzinovyev: it indicates that `txn` or `txn_participation` table is ahead

idb/postgres/postgres_integration_test.go
tolikzinovyev: it is your test, why did you use writer? :)

idb/postgres/postgres_integration_test.go
tolikzinovyev: Behind is not possible. We commit the main transaction only when the two auxiliary transactions are committed.

idb/postgres/postgres_integration_test.go
tolikzinovyev: yes

idb/postgres/postgres_integration_test.go
tolikzinovyev: No, there are no nested transactions. We need to have multiple database sessions if we want to have parallelism.

idb/postgres/postgres_integration_test.go
winder: I was probably trying to reduce the scope of the unit test, but now that it requires more implementation details from db.AddBlock that makes less sense.

Could you switch it to db.AddBlock?

idb/postgres/postgres_integration_test.go
winder: Thanks, I think this is what I was missing earlier. Your decisions here make a lot more sense to me now.

idb/postgres/postgres_integration_test.go
tolikzinovyev: done

winder (approved):

mxmauro:
Hi, what happens if connection to postgres is lost (or app crashes) while writing to `txn` and `txn_participation` and one of them completes but not the other? Does all of this run inside an outer transaction or is there a recovery process for this scenario?

tolikzinovyev:
> Hi, what happens if connection to postgres is lost (or app crashes) while writing to `txn` and `txn_participation` and one of them completes but not the other? Does all of this run inside an outer transaction or is there a recovery process for this scenario?

The transactions writing to `txn` and `txn_participation` tables must finish before we commit the "main" transaction. So, the only inconsistency that can happen is that information in `txn` and/or `txn_participation` tables is ahead of all other information. There is no recovery process.

You can study the changes in `idb/postgres/postgres.go`.

mxmauro:
Hi @tolikzinovyev excellent. Thanks for the feedback.

PR #807, opened 2021-12-10
Write transactions earlier if possible

tolikzinovyev:
## Summary

If `proto.EnableAssetCloseAmount` is already true, we can start writing transactions before running the evaluator.

9200 -> 9460 TPS

## Test Plan

`TestAddBlockAssetCloseAmountInTxnExtra` already tests that we write transactions when `proto.EnableAssetCloseAmount` is false. There are multiple test checking that transactions are written when `proto.EnableAssetCloseAmount` is true.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -296,16 +296,31 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 				return fmt.Errorf("AddBlock() err: %w", err)
 			}
 		} else {
+			proto, ok := config.Consensus[block.BlockHeader.CurrentProtocol]
+			if !ok {
+				return fmt.Errorf(
+					"AddBlock() cannot find proto version %s", block.BlockHeader.CurrentProtocol)
+			}
+			protoChanged := !proto.EnableAssetCloseAmount
+			proto.EnableAssetCloseAmount = true
+
 			var wg sync.WaitGroup
 			defer wg.Wait()
 
-			// Write transaction participation in a parallel db transaction.
+			// Write transaction participation and possibly transactions in a parallel db
+			// transaction.
 			var err0 error
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
 
 				f := func(tx pgx.Tx) error {
+					if !protoChanged {
+						err := writer.AddTransactions(block, block.Payset, tx)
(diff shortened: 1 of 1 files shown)


tolikzinovyev:
@winder let me know if you like the change. If so, I will add a small test.

idb/postgres/postgres.go
```diff
@@ -296,16 +296,31 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
+			// Write transaction participation and possibly transactions in a parallel db
+			// transaction.
 			var err0 error
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
 
 				f := func(tx pgx.Tx) error {
+					if !protoChanged {
+						err := writer.AddTransactions(block, block.Payset, tx)
```
winder, line 319: I think you want want this in a separate `txWithRetry` so that it is written in parallel to `AddTransactionParticipation`?

I like this change but am not sure we need it right now. At the very least I think it's good to add a comment below near `AddTransactions`, something along the lines of:
> if proto.EnableAssetCloseAmount was already true, the transactions can be loaded with block.Payset instead of modifiedTransactions while the evaluator is running

tolikzinovyev: No, writing transactions and transaction participation together takes less time then eval + writing other state, so we don't need more parallelism; it would probably hurt.

I will add a comment.

tolikzinovyev:
Actually, we already have a test that I wanted: `TestAddBlockAssetCloseAmountInTxnExtra`.

idb/postgres/postgres.go
```diff
@@ -338,17 +347,19 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
-			wg.Add(1)
-			go func() {
-				defer wg.Done()
-
-				f := func(tx pgx.Tx) error {
-					return writer.AddTransactions(block, modifiedTxns, tx)
-				}
-				err1 = db.txWithRetry(serializable, f)
-			}()
+			if protoChanged {
```
winder, line 351:
```suggestion
			// Skip if transaction writing has already started.
			if protoChanged {
```

winder (approved):
LGTM, please just add one more comment suggestion

PR #810, opened 2021-12-17
Updating readme for postgres 13 and AZ note

algojack:
We decided to add a note to the public about availability zones in AWS greatly affecting latency on indexer writer and RDS in our tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ There are two primary modes of operation:
 ### Database updater
 In this mode, the database will be populated with data fetched from an [Algorand archival node](https://developer.algorand.org/docs/run-a-node/setup/types/#archival-mode). Because every block must be fetched to bootstrap the database, the initial import for a ledger with a long history will take a while. If the daemon is terminated, it will resume processing wherever it left off.
 
+Keeping indexer daemon as close as possible to database helps minimize latency. For example if using AWS EC2 and RDS, we suggest putting EC2 in the same VPC, Region, and even Availability Zone.
+
 You should use a process manager, like systemd, to ensure the daemon is always running. Indexer will continue to update the database as new blocks are created.
 
 To start indexer as a daemon in update mode, provide the required fields:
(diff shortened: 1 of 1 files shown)


algobarb (approved):

egieseke (approved):
Looks good.

README.md
```diff
@@ -90,6 +90,8 @@ Alternatively if indexer is running on the same host as the archival node, a sim
 ~$ algorand-indexer daemon --algod-net yournode.com:1234 -d /path/to/algod/data/dir --postgres "user=readonly password=YourPasswordHere {other connection string options for your database}"
 ```
 
+**Note**: Keeping indexer daemon as close as possible to database helps minimize latency. For example if using AWS EC2 and RDS, we suggest putting EC2 in the same VPC, Region, and even Availability Zone.
```
winder, line 93: Could you move this to above the command examples? I think it would fit nicely between the `In this mode, ...` and `You s should use a process manager...` sections. No need for `**Note**: `, everything here is important to read not just this recommendation.

algojack: done

README.md
```diff
@@ -78,6 +78,8 @@ There are two primary modes of operation:
 ### Database updater
 In this mode, the database will be populated with data fetched from an [Algorand archival node](https://developer.algorand.org/docs/run-a-node/setup/types/#archival-mode). Because every block must be fetched to bootstrap the database, the initial import for a ledger with a long history will take a while. If the daemon is terminated, it will resume processing wherever it left off.
 
+Keeping indexer daemon as close as possible to database helps minimize latency. For example if using AWS EC2 and RDS, we suggest putting EC2 in the same VPC, Region, and even Availability Zone.
```
winder, line 81:
```suggestion
Keeping the indexer daemon as close as possible to the database helps minimize latency. For example, if using AWS EC2 and RDS, we suggest putting EC2 in the same VPC, Region, and even Availability Zone.
```

winder (approved):

PR #813, opened 2021-12-28
adding dedupe map to track transaction ids we've already seen

### LINKED ISSUE #812: Duplicate Transactions in /v2/block/N endpoint
### Subject of the issue

When requesting a block that contains app calls with inner transactions, the root app call transactions are duplicated for each inner transaction. 

### Steps to reproduce

https://algoindexer.algoexplorerapi.io/v2/blocks/17733599
TxID: DCXHCTQUA4PIZX4IO2JNZYON7JZHDTVPDX6QVIPCZX64XOCGMYTQ is returned multiple times

barnjamin:
Closes #812

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -794,12 +794,21 @@ func (si *ServerImplementation) fetchBlock(ctx context.Context, round uint64) (g
 			UpgradeVote:       &upgradeVote,
 		}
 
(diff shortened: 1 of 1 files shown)


winder:
Thanks Ben. I added a test and am resolving this with #840

PR #814, opened 2021-12-28
Hotfix/2.6.6

winder:
## Summary

Allow unexpected sig type changes in cache. The transaction has already been validated by algod, so any discrepancy here would be correct.

This particular discrepancy appears to be related to rekeyed LSIG accounts and has also been fixed with the 2.8.0 refactoring.

Most of the changes in this PR relate to getting a logger into the cache object.

## Test Plan

New unit test.


algorandskiy (approved):

jasonpaulos (approved):
Looks good.

Unfortunately the 2.6.x branches have been consistently failing `make e2e`, so I'm not too worried about the CI failure. The go unit tests have still passed

PR #817, opened 2022-01-03
Better context usage

tolikzinovyev:
## Summary

Add missing context expiration checks and remove unnecessary context expiration checks.


idb/postgres/postgres.go
```diff
@@ -954,18 +954,14 @@ func (db *IndexerDb) yieldTxnsThreadSimple(ctx context.Context, rows pgx.Rows, r
 				}
 			}
 		}
-		select {
-		case <-ctx.Done():
```
winder, line 958: Removing the context from these functions because `pgx.Rows` is the blocking object and it's responsible for terminating early?

tolikzinovyev: Yes, `Rows.Next()` is blocking but will terminate early. `Rows.Err()` will return any error occurred while reading rows, including context expiration.

winder (approved):

PR #820, opened 2022-01-10
Return Inner and Root transactions for nested logs support in C2C calls

algochoi:
## Summary

This PR includes some changes to support proper nested logs in inner transactions and adds/changes some tests to reflect this.

`feature/contract-to-contract` will now allow inner transactions to have `appl` calls. This means inner txns may also be recursive and many "levels" deep, as inner `appl` calls may have inner txns themselves. 

Since logs are associated with each `EvalDelta`, logs returned by inner transactions would have to be recursively parsed or else the current API will return only the root txn's logs, even if the query matches an inner txn. This PR changes the transaction query to fetch for an inner txn if a `returnInnerTxn` flag in the filter is set to true so nested logs can be returned properly. 

## Test Plan

- Changed some existing tests that check for inner txns to include nested `appl` calls. 
- Added an additional check to return the nested logs properly and check that non UTF-8s can be parsed correctly in inner asset/appl txns. 
- Added a test to check if inner logs are being returned properly.

**Note**: This PR also depends on the go-algorand `feature/contract-to-contract` branch due to nested inner txn logic.


util/test/account_testutil.go
```diff
@@ -320,33 +321,142 @@ func MakeAppCallWithInnerTxn(appSender, paymentSender, paymentReceiver, assetSen
 		{
 			SignedTxn: transactions.SignedTxn{
 				Txn: transactions.Transaction{
-					Type: protocol.PaymentTx,
+					Type: protocol.ApplicationCallTx,
```
algochoi, line 324: Pedantic change - just to reflect that only `appl` calls currently support itxns

idb/postgres/postgres.go
```diff
@@ -935,20 +935,20 @@ func (db *IndexerDb) yieldTxnsThreadSimple(ctx context.Context, rows pgx.Rows, r
...
-			} else {
-				// Root transaction.
-				row.Txn = new(transactions.SignedTxnWithAD)
-				*row.Txn, err = encoding.DecodeSignedTxnWithAD(txn)
-				if err != nil {
-					err = fmt.Errorf("error decoding txn, err: %w", err)
-					row.Error = err
-				}
 			}
+			// Root transaction.
```
algochoi, line 939: Return the itxn and root txn when reading from the db if possible.

winder: Is there ever a scenario where we want the root txn and the inner txn?

I think we should add an option to the transaction parameters to select which we want. It would be a faster query and require less data.

accounting/rewind.go
```diff
@@ -99,7 +99,7 @@ func AccountAtRound(account models.Account, round uint64, db idb.IndexerDb) (acc
 		MinRound: round + 1,
 		MaxRound: account.Round,
 	}
-	txns, r := db.Transactions(context.Background(), tf)
```
winder, line 102: This flag should be in the `idb.TransactionFilter` object. We're using a "parameter object" pattern pretty heavily in indexer for these things since there are so many options.

algochoi: That sounds like a much better option - I'll fix some of the changes and move the flag in there

idb/postgres/postgres.go
```diff
@@ -744,13 +744,22 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
...
 		whereParts = append(whereParts, "(t.txn -> 'txn' -> 'rekey') IS NOT NULL")
 	}
-	query = "SELECT t.round, t.intra, t.txn, root.txn, t.extra, t.asset, h.realtime FROM txn t JOIN block_header h ON t.round = h.round"
+
+	// If returnInnerTxnOnly flag is false, then return the root transaction
+	if !returnInnerTxnOnly {
+		query = "SELECT t.round, t.intra, t.txn, root.txn, t.extra, t.asset, h.realtime FROM txn t JOIN block_header h ON t.round = h.round"
+	} else {
+		query = "SELECT t.round, t.intra, t.txn, NULL, t.extra, t.asset, h.realtime FROM txn t JOIN block_header h ON t.round = h.round"
+	}
```
winder, line 753: Nice use of NULL to avoid needing different Scan parameters later on.

This works until we have another flag (i.e. the account queries). Since there's only one flag now, this is good.

winder (changes requested):
Thanks for the updates, your changes look good. The only thing I see missing is a test that uses the new flag directly. Call `db.Transactions` with and without `ReturnInnerTxnOnly` and verify that the RootTxn is not returned when `ReturnInnerTxnOnly` is set.

idb/postgres/postgres_integration_test.go
```diff
@@ -1502,35 +1519,66 @@ func TestAddBlockAppOptInOutSameRound(t *testing.T) {
...
 	var appAddr basics.Address
 	appAddr[1] = 99
 
 	tests := []struct {
-		name    string
-		matches int
-		filter  idb.TransactionFilter
+		name        string
+		matches     int
+		returnInner bool
```
algochoi, line 1532: I just added a flag to the existing tests here and checked if the inner txn was being returned.

winder: nice :+1:

winder (approved):

PR #823, opened 2022-01-14
Add retry loop to block handler.

winder:
## Summary

If the Postgres database handles are exhausted due to many pending queries, block import will fail. With the existing code that would cause Indexer to terminate.

This PR adds a retry loop to ensure the process eventually recover after this sort of transient failure.

## Test Plan

New unit test.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -104,7 +104,7 @@ var daemonCmd = &cobra.Command{
 				bot.SetNextRound(nextRound)
 
 				imp := importer.NewImporter(db)
-				handler := blockHandler(&imp)
+				handler := blockHandler(imp, 1 * time.Second)
 				bot.SetBlockHandler(handler)
 
 				logger.Info("Starting block importer.")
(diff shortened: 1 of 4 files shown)


tolikzinovyev:
How about creating 2 dedicated connections for block import? (`AddBlock()` needs 2).

winder:
> How about creating 2 dedicated connections for block import? (`AddBlock()` needs 2).

I think this is still useful as a basic safety net, but your idea is even better.

Another possible issue is that Aurora replication (presumably) needs its own connection handle(s). One suspicion is this is why our mobile wallet deployments fall behind from time to time. To resolve that one we might need to make the pgx pool size configurable. Seems like [ConnectConfig](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#ConnectConfig) has plenty of options for doing that.

tolikzinovyev:
I'm just not sure ignoring an error is a good solution.

Not following about aurora replication.

winder:
> I'm just not sure ignoring an error is a good solution.
It doesn't ignore the error. An error is logged, it's up to the service runner to decide whether or not to ignore errors.

> Not following about aurora replication.
I can follow up on this separately.

winder:
@tolikzinovyev I created some followup tickets to reduce the chance that this will happen:
#826 
#827

AlgoStephenAkiki (approved):

shiqizng (approved):
looks fine.

tolikzinovyev (approved):
Looks good!

PR #824, opened 2022-01-14
Add CLI option for providing configfile.

shiqizng:
## Summary

Add a new CLI flag, `--configfile`, to take a file path to the configs file. Values set on the CLI would overwrite the values set in this file.

### DIFF
--- a/cmd/algorand-indexer/main.go
+++ b/cmd/algorand-indexer/main.go
@@ -61,14 +61,12 @@ var rootCmd = &cobra.Command{
 		if configFile != "" {
 			configs, err := os.Open(configFile)
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "%v", err)
-				os.Exit(1)
+				maybeFail(err, "%v", err)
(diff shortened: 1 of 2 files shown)


cmd/algorand-indexer/main.go
```diff
@@ -58,6 +58,20 @@ var rootCmd = &cobra.Command{
 			err = pprof.StartCPUProfile(profFile)
 			maybeFail(err, "%s: start pprof, %v", cpuProfile, err)
 		}
+		if configFile != "" {
+			configs, err := os.Open(configFile)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "%v", err)
+				os.Exit(1)
+			}
```
winder, line 66: You can use the `maybeFail` helper for print + exit messages like this and the one below.

winder (changes requested):
Looks great. One request, could you update the "Configuration file" section of the README?

winder (approved):

Issue #676, opened 2021-09-24
Account Min Balance


barnjamin:
## Problem

I'd like a simple way to know the min-balance on a given account

Also filed an issue in go-algorand https://github.com/algorand/go-algorand/issues/2898 but seems like it should be here as well

## Solution

Compute the min-balance from the opted in assets/apps from the go-algorand Account method

winder:
Closing in favor of #808

PR #825, opened 2022-01-14
account min-balance

### LINKED ISSUE #808: Expose new account info field `min-balance`
## Problem

https://github.com/algorand/go-algorand/pull/3287 has introduced `min-balance` to the response of `algod`'s `/v2/accounts/{{ACCOUNT}}` as well as to the `goal account info`.  However, this is not yet available in the indexer

## Dependencies

https://github.com/algorand/go-algorand/pull/3287 

## Urgency

Medium

tzaffi:
### Fixes
#808 #676 

### Summary of changes
* `accounting/calculated.go` - `EnrichMinBalance`: main workhorse of thie PR
* `accounting/calculated_test.go` - new unit tests
* `api/error_messages.go` - new error message
* `api/handlers.go` - mostly in `fetchAccounts()`
* `api/indexer.oas2.json` - add `min-balance` and fix a couple of typos (discovered by future circle test in #839 )
* `idb/postgres/postgres.go` - modify `GetAccounts` adding a third output of type `*bookkeeping.BlockHeader`

#### Minor Changes in `GetAccounts()` Just for Tests to Pass
* `api/handlers_test.go`
* `cmd/e2equeries/main.go`
* `cmd/idbtest/idbtest.go`
* `idb/dummy/dummy.go`
* `idb/idb.go`
* `idb/mocks/IndexerDb.go`
* `idb/postgres/postgres_integration_test.go`
* `util/test/testutil.go`

#### Auto Generated
* `api/generated/common/routes.go`
* `api/generated/common/types.go`
* `api/generated/v2/routes.go`
* `api/generated/v2/types.go`
* `api/indexer.oas3.yml`

## Test Plan
Unit Tests. Working on an integration tests via the sandbox.

#### Related Follow-up PR
#839 - Add a test that alerts automatically when a new discrepency arises between `algod` and `indexer`

### `algod` compatriate
https://github.com/algorand/go-algorand/pull/3287


Makefile
```diff
@@ -18,10 +18,16 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 cmd/algorand-indexer/algorand-indexer: idb/postgres/internal/schema/setup_postgres_sql.go go-algorand
 	cd cmd/algorand-indexer && go build -ldflags="${GOLDFLAGS}"
 
-go-algorand:
-	git submodule update --init && cd third_party/go-algorand && \
+# TODO: allow specifying the branch
+go-algorand-submodule:
+	git submodule update --init --force --remote
```
tzaffi, line 23: needed to force to stop getting stale `go-algorand` commit

README.md
```diff
@@ -1,23 +1,28 @@
+<!-- markdownlint-disable MD041 -->
```
tzaffi, line 1: linting... with vscode's `markdownlint` extenstion

api/README.md
```diff
@@ -5,9 +5,10 @@ We are using a documentation driven process.
 The API is defined using [OpenAPI v2](https://swagger.io/specification/v2/) in **indexer.oas2.yml**.
 
 ## Updating REST API
+
 The Makefile will install our fork of **oapi-codegen**, use `make oapi-codegen` to install it directly.
 
-1. Document your changes by editing **indexer.oas2.yml**
+1. Document your changes by editing **indexer.oas2.json**
```
tzaffi, line 11: I believe this is a typo

parity/json_diff.py
```diff
@@ -0,0 +1,197 @@
+from copy import deepcopy
```
tzaffi, line 1: Code to generate **algod** v. **indexer** parity reports. This could be converted into a circle test that:
1. pulls latest `go-algorand` submodule
2. runs the report
3. asserts that the results are as expected. In particular:
  * all _essential_ features of **algod** have their counterparts in **indexer** (think of `min-balance`)
  * terminology and responses remain consistent
  * etc.

parity/reports/algod2indexer_all_human.txt
```diff
@@ -0,0 +1,210 @@
+INDEXER--->definitions.Account.properties.closed-at-round:{"description":"Round during which this account was most recently closed.","type":"integer","x-algorand-format":"uint64"}
```
tzaffi, line 1: This report shows all the "essential" difference between **algod** and **indexer** in a purportedly human-friendly format. There are variants of this report that limit the scope to:
* changes of common attributes
* attributes which are not present in **algod** but exist in **indexer**
* attributes which are present in **algod** but missing from **indexer**

parity/reports/algod2indexer_all_summary.txt
```diff
@@ -0,0 +1,115 @@
+____________________      61 New Attributes       ____________________
```
tzaffi, line 1: This report shows all the "essential" difference between **algod** and **indexer** in a purportedly human-friendly format. There are variants of this report that limit the scope to:
* changes of common attributes
* attributes which are not present in **algod** but exist in **indexer**
* attributes which are present in **algod** but missing from **indexer**

However, this report shos _ALL_ the differences (the union of the above)

parity/reports/algod2indexer_full.yml
```diff
@@ -0,0 +1,167 @@
...
+    deleted:
+    - INDEXER: '{"description":"Whether or not...'
+    - ALGOD: null
+    min-balance:
+    - INDEXER: null
+    - ALGOD: '{"description":"MicroAlgo bala...'
+    sig-type:
+      description:
+      - INDEXER: '"Indicates what type of signature is used by this account, must be one of:\n* sig\n* msig\n* lsig\n* or null if unknown"'
+      - ALGOD: '"Indicates what type of signature is used by this account, must be one of:\n* sig\n* msig\n* lsig"'
```
tzaffi, line 18: seems outdated on indexer

parity/reports/algod2indexer_full.yml
```diff
@@ -0,0 +1,167 @@
...
+    - ALGOD: null
+ApplicationLogData:
+- INDEXER: '{"description":"Stores the glo...'
+- ALGOD: null
+ApplicationParams:
+  properties:
+    global-state-schema:
+      description:
+      - INDEXER: '"[\\lsch\\] global schema"'
+      - ALGOD: '"[\\gsch\\] global schema"'
```
tzaffi, line 52: looks like a typo

parity/reports/algod2indexer_full.yml
```diff
@@ -0,0 +1,167 @@
...
+- ALGOD: '{"description":"Details about ...'
+StateSchema:
+- INDEXER: '{"description":"Represents a \...'
+- ALGOD: null
+TealValue:
+  properties:
+    type:
+      description:
+      - INDEXER: '"\\[tt\\] value type."'
+      - ALGOD: '"\\[tt\\] value type. Value `1` refers to **bytes**, value `2` refers to **uint**"'
```
tzaffi, line 131: looks outdated on indexer

parity/reports/algod2indexer_dropped.yml
```diff
@@ -0,0 +1,37 @@
+Account:
+  properties:
+    min-balance:
+    - INDEXER: null
+    - ALGOD: '{"description":"MicroAlgo bala...'
```
tzaffi, line 5: gotcha `min-balance` !!!

parity/reports/algod2indexer_dropped.yml
```diff
@@ -0,0 +1,37 @@
...
+  required:
+  - min-balance
```
tzaffi, line 7: here too

parity/reports/algod2indexer_dropped.yml
```diff
@@ -0,0 +1,37 @@
...
+ApplicationParams:
+  required:
+  - creator
```
tzaffi, line 10: not sure about this one

parity/reports/algod2indexer_dropped.yml
```diff
@@ -0,0 +1,37 @@
...
+- ALGOD: '{"description":"Request data t...'
+DryrunSource:
+- INDEXER: null
+- ALGOD: '{"description":"DryrunSource i...'
+DryrunState:
+- INDEXER: null
+- ALGOD: '{"description":"Stores the TEA...'
+DryrunTxnResult:
+- INDEXER: null
+- ALGOD: '{"description":"DryrunTxnResul...'
```
tzaffi, line 25: these look legitimately algod only

parity/reports/algod2indexer_dropped.yml
```diff
@@ -0,0 +1,37 @@
...
+- ALGOD: '{"description":"An error respo...'
+ParticipationKey:
+- INDEXER: null
+- ALGOD: '{"description":"Represents a p...'
+PendingTransactionResponse:
+- INDEXER: null
+- ALGOD: '{"description":"Details about ...'
+Version:
+- INDEXER: null
+- ALGOD: '{"description":"algod version ...'
```
tzaffi, line 37: Indexer has a "version" but not as a model

parity/test_indexer_v_algod.py
```diff
@@ -0,0 +1,146 @@
...
+# Only compare swagger "definitions":
+MODELS_ONLY = True
+
+REPO_DIR = Path.cwd()
+GOAL_DIR = REPO_DIR / "third_party" / "go-algorand"
+REPORTS_DIR = REPO_DIR / "parity" / "reports"
+
+
+def tsetup():
+    exclude = [
```
tzaffi, line 23: Excluding these keys from the diff analysis

.circleci/config.yml
```diff
@@ -66,6 +66,7 @@ commands:
       - run: make test
       - run: make fakepackage
       - run: make e2e
+      - run: make parity
```
tzaffi, line 69: added to circle

parity/test_indexer_v_algod.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    excludes, indexer_swgr, algod_swgr = tsetup()
+
+    for diff_type in reports:
+        diff = generate_diff(algod_swgr, indexer_swgr, excludes, diff_type)
+        save_yaml(diff, diff_type)
+
+
+def test_parity(reports: List[str] = ASSERTIONS, save_new: bool = True):
+    excludes, indexer_swgr, algod_swgr = tsetup()
```
tzaffi, line 130:
```suggestion
    excludes, indexer_swgr, algod_swgr = tsetup()
    """
    For each report in reports:
       1.  load the pre-existing yaml report into `old_diff`
       2.  re-generate the equivalent report by comparing `algod_swgr` with `indexer_swgr`
       3. compute the `diff_of_diffs` between these two reports
       4. assert that there is no diff
    """
```

parity/test_indexer_v_algod.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    For each report in reports:
+       1. load the pre-existing yaml report into `old_diff`
+       2. re-generate the equivalent report by comparing `algod_swgr` with `indexer_swgr`
+       3. compute the `diff_of_diffs` between these two reports
+       4. assert that there is no diff
+    """
```
tzaffi, line 137: pseudocode for how the **algod** v **indexer** parity test works

idb/postgres/internal/encoding/encoding.go
```diff
@@ -376,6 +376,17 @@ func TrimAccountData(ad basics.AccountData) basics.AccountData {
...
+	// data/basics/userBalance.go::func (basics.AccountData) MinBalance(proto *config.ConsensusParams) (res MicroAlgos)
+	// then we'll need the following information:
+	// * proto.MinBalance:  	:-)
+	// * ad.Assets:				:-(
+	// * ad.AppParams:			:-(
+	// * ad.AppLocalStates:		:-(
+	// * ad.TotalAppSchema:		:-)
+	// * ad.TotalExtraAppPages: :-)
+	// but we're erasing 3 of these fields before we hand the json off to the upsert
+
```
tzaffi, line 389: 🤔

idb/postgres/internal/encoding/encoding.go
```diff
@@ -376,6 +376,17 @@ func TrimAccountData(ad basics.AccountData) basics.AccountData {
 	ad.AppLocalStates = nil
 	ad.AppParams = nil
 
+	// Uh-oh: If we want to calculate MinBalance using
+	// data/basics/userBalance.go::func (basics.AccountData) MinBalance(proto *config.ConsensusParams) (res MicroAlgos)
+	// then we'll need the following information:
+	// * proto.MinBalance:  	:-)
+	// * ad.Assets:				:-(
+	// * ad.AppParams:			:-(
```
tzaffi, line 384: can get these from `account_app`

idb/postgres/internal/encoding/encoding.go
```diff
@@ -376,6 +376,17 @@ func TrimAccountData(ad basics.AccountData) basics.AccountData {
 	ad.AppLocalStates = nil
 	ad.AppParams = nil
 
+	// Uh-oh: If we want to calculate MinBalance using
+	// data/basics/userBalance.go::func (basics.AccountData) MinBalance(proto *config.ConsensusParams) (res MicroAlgos)
+	// then we'll need the following information:
+	// * proto.MinBalance:  	:-)
+	// * ad.Assets:				:-(
```
tzaffi, line 383: can get these from `account_asset`

idb/postgres/internal/encoding/encoding.go
```diff
@@ -376,6 +376,17 @@ func TrimAccountData(ad basics.AccountData) basics.AccountData {
...
+	// * ad.AppParams:			:-(
+	// * ad.AppLocalStates:		:-(
```
tzaffi, line 385: can get these from `local_state FROM account_app a WHERE addr = `

parity/test_indexer_v_algod.py
```diff
@@ -0,0 +1,153 @@
+from pathlib import Path
```
tzaffi, line 1: API alignment test/report for indexer v. algod

parity/test_indexer_v_algod.py
```diff
@@ -0,0 +1,153 @@
...
+from .json_diff import deep_diff, prettify_diff
+
+NEW, OVERLAP, DROPPED, FULL = "new", "overlap", "dropped", "full"
+DIFF_TYPES = [NEW, OVERLAP, DROPPED, FULL]
+
+# These are the diff reports that will be run and compared/asserted against:
+ASSERTIONS = [DROPPED, FULL]
+
+# Only compare swagger "definitions":
+MODELS_ONLY = True
```
tzaffi, line 15: Currently, the test only looks at the `definitions:` part of each Swagger. 

* Do we also want to test `responses:` too?
* Anything else we want to check?

Makefile
```diff
@@ -18,10 +18,17 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 cmd/algorand-indexer/algorand-indexer: idb/postgres/internal/schema/setup_postgres_sql.go go-algorand
 	cd cmd/algorand-indexer && go build -ldflags="${GOLDFLAGS}"
 
-go-algorand:
-	git submodule update --init && cd third_party/go-algorand && \
+# TODO: allow specifying the branch
+go-algorand-submodule:
```
tzaffi, line 22: probably need to revert bc need to be more deliberate about which `go-algorand` branch gets pulled in on circle

Makefile
```diff
@@ -71,4 +78,8 @@ sign:
 test-package:
 	mule/e2e.sh
 
+# check that indexer and algod differ at the expected places
+parity:
+	pytest -sv parity
```
tzaffi, line 83: currently, `parity` is being skipped because it's not in the .PHONY

Makefile
```diff
@@ -18,10 +18,17 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 cmd/algorand-indexer/algorand-indexer: idb/postgres/internal/schema/setup_postgres_sql.go go-algorand
 	cd cmd/algorand-indexer && go build -ldflags="${GOLDFLAGS}"
 
-go-algorand:
-	git submodule update --init && cd third_party/go-algorand && \
+# TODO: allow specifying the branch
+go-algorand-submodule:
+	git submodule sync
```
tzaffi, line 23:
```suggestion
```

Makefile
```diff
@@ -18,10 +18,17 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
...
+	git submodule update --init
```
tzaffi, line 24:
```suggestion
go-algorand-submodule:
	git submodule update --init
```

Makefile
```diff
@@ -71,4 +76,9 @@ sign:
 test-package:
 	mule/e2e.sh
 
+# check that indexer and algod differ at the expected places
+# currently it is turned off since it's not in the .PHONY
```
tzaffi, line 80:
```suggestion
```

accounting/calculated.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return nil
+}
+// minBalanceProjection projects a part of a generated.Account to a basics.AccountData struct.
+// It does so for the purpose of calculating an account's minumum balance using the official
+// calculator go-algorand/data/basics/userBalance.go::MinBalance()
+// Some data which is relevant for MinBalance but duplicated is ignored. In particular:
+// * TotalAppSchema - gotten from account.AppsTotalSchema and not re-calc'ed from CreatedApps
+// * TotalExtraPages - gotten from account.AppsTotalExtraPages and not re-calc'ed from CreatedApps
+func minBalanceProjection(account *generated.Account) basics.AccountData {
```
tzaffi, line 129: There is a similar function `AccountToAccountData` in `go-algorand` that fills out all of the fields (but has some minor issues when applied to Indexer's `generated.Account`. It is also possible to try and unmarshall the `basics.Account` directly from the database as is partially done elsewhere. After some thought, I decided that this was the simplest approach - get only the essential data into a `basics.AccountData` at the most convenient location and calculate `MinBalance` from it.

api/handlers.go
```diff
@@ -13,6 +13,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
```
tzaffi, line 16: for `BlockHeader`

api/handlers.go
```diff
@@ -155,6 +156,8 @@ func (si *ServerImplementation) LookupAccountByID(ctx echo.Context, accountID st
 		IncludeAssetParams:   true,
 		Limit:                1,
 		IncludeDeleted:       boolOrDefault(params.IncludeAll),
+		// omit MinBalance when there is a rewind
+		IncludeMinBalance: boolOrDefault(params.IncludeAll) && params.Round != nil,
```
tzaffi, line 160:
```suggestion
		IncludeMinBalance: boolOrDefault(params.IncludeAll) && params.Round == nil,
```

api/indexer.oas2.json
```diff
@@ -960,7 +964,7 @@
       ],
       "properties": {
         "type": {
-          "description": "\\[tt\\] value type.",
+          "description": "\\[tt\\] value type. Value `1` refers to **bytes**, value `2` refers to **uint**",
```
tzaffi, line 967: This was spotted by the new python indexer vs. algod parity test.

api/indexer.oas2.json
```diff
@@ -1036,7 +1040,7 @@
           "$ref": "#/definitions/ApplicationStateSchema"
         },
         "global-state-schema": {
-          "description": "[\\lsch\\] global schema",
+          "description": "[\\gsch\\] global schema",
```
tzaffi, line 1043: ditto

api/handlers_test.go
```diff
@@ -752,7 +753,12 @@ func TestTimeouts(t *testing.T) {
 	}
 	transactionFunc := mostMockFunctions("Transactions")
 	applicationsFunc := mostMockFunctions("Applications")
-	accountsFunc := mostMockFunctions("GetAccounts")
+	accountsFunc := func(mockIndexer *mocks.IndexerDb, timeout <-chan time.Time) {
+		mockIndexer.
+			On("GetAccounts", mock.Anything, mock.Anything, mock.Anything).
+			WaitUntil(timeout).
+			Return(nil, uint64(0), &bookkeeping.Block{})
```
tzaffi, line 760:
```suggestion
			Return(nil, uint64(0), &bookkeeping.BlockHeader{})
```

accounting/calculated.go
```diff
@@ -0,0 +1,151 @@
...
+	proto, ok := config.Consensus[blockheader.CurrentProtocol]
+	if !ok {
+		return fmt.Errorf("cannot EnrichMinBalance as blockheader's CurrentProtocol is not known to Consensus. CurrentProtocol: %s", blockheader.CurrentProtocol)
+	}
+	raw := minBalanceProjection(account).MinBalance(&proto).Raw
+	account.MinBalance = &raw
+	return nil
+}
+
+// copy pasta the next few from api::pointer_utils.go and then renamed
```
tzaffi, line 28:
```suggestion
// inspired by api::pointer_utils.go and then renamed
```

tzaffi:
@winder and @algorandskiy have brought to my attention some important changes that are in the works to handle unlimited assets and which affect the way **min-balance** is calculated (though technically, the formula remains the same). In particular, we [can see](https://github.com/algorand/go-algorand/blob/feature/unlimited-assets/ledger/ledgercore/accountdata.go#L44) that new `Total*` fields will enter the ledger through a brand new struct `AccountData`. It may make sense for indexer to import these new ledger fields directly, in which case the minbalance calculation would use the [new stand alone MinBalance func](https://github.com/algorand/go-algorand/blob/feature/unlimited-assets/data/basics/userBalance.go#L474) something like the following:

```go
var a basics.Account // which perhaps? has as part of it a ledger.AccountBase, or something like that

... populate a ...


minBalance := basics.MinBalance(proto, a.TotalAssets, a.TotalAppSchema, a.TotalAppParams, a.TotalAppLocalStates, a.TotalExtraAppPages)
```

@winder - Do you think we should wait for these changes to the ledger to get merged into `go-algorand` before continuing with this PR? On the other hand, do you see any fairly simple and future-aware work-arounds that can let us proceed without waiting?

tzaffi:
@winder and @algorandskiy have brought to my attention some important changes that are in the works to handle unlimited assets and which affect the way **min-balance** is calculated (though technically, the formula remains the same). In particular, we [can see](https://github.com/algorand/go-algorand/blob/feature/unlimited-assets/ledger/ledgercore/accountdata.go#L44) that new `Total*` fields will enter the ledger through a brand new struct `AccountData`. It may make sense for indexer to import these new ledger fields directly, in which case the minbalance calculation would use the [new stand alone MinBalance func](https://github.com/algorand/go-algorand/blob/feature/unlimited-assets/data/basics/userBalance.go#L474) something like the following:

```go
var a basics.Account // which perhaps? has as part of it a ledger.AccountBase, or something like that

... populate a ...


minBalance := basics.MinBalance(proto, account.TotalAssets, accou
```

@winder - Do you think we should wait for these changes to the ledger to get merged into `go-algorand` before continuing with this PR? On the other hand, do you see any fairly simple and future-aware work-arounds that can let us proceed without waiting?

idb/idb.go
```diff
@@ -252,6 +252,8 @@ type AccountQueryOptions struct {
 
 	IncludeAssetHoldings bool
 	IncludeAssetParams   bool
+	// IncludeMinBalance indicates whether to calculate min-balance as well (a realatively slow operation)
+	IncludeMinBalance bool
```
winder, line 256: These are database layer options, so this option doesn't need to be here. I think we can get away with using the nil round check, or just skip/unset it in the rewind function.

idb/idb.go
```diff
@@ -173,7 +173,7 @@ type IndexerDb interface {
 	// The next multiple functions return a channel with results as well as the latest round
 	// accounted.
 	Transactions(ctx context.Context, tf TransactionFilter) (<-chan TxnRow, uint64)
-	GetAccounts(ctx context.Context, opts AccountQueryOptions) (<-chan AccountRow, uint64)
+	GetAccounts(ctx context.Context, opts AccountQueryOptions) (<-chan AccountRow, uint64, *bookkeeping.BlockHeader)
```
winder, line 176: @tolikzinovyev I'm curious what you think about this. The protocol is needed for the min balance calculation, but this approach puts accounting logic in the REST API. There are a few other bits of accounting logic outside of the eval function (sig type, created/deleted at, pending rewards calculations, maybe others). Is there a way to make this sort of accounting logic pluggable so that it can be encapsulated so that it doesn't need to be part of the DB/REST API?

cc: @AlgoStephenAkiki @shiqizng

tolikzinovyev: I think the right solution is to return `basics.AccountData`s in `IndexerDb.GetAccounts()` like it's done in other `IndexerDb` functions, and in `api` invoke `basics.AccountData.MinBalance()`.

winder: That seems reasonable, but it still has the same problem, you need ConsensusParams to pass into MinBalance.

tolikzinovyev: I wouldn't say it's a problem. `GetAccounts()` returns the current round. Similarly, it can return the current consensus version.

tzaffi: 👍

tzaffi: Look further into this issue I have some thoughts regarding @tolikzinovyev suggestions.

1. modifying `GetAccounts()` to return the current consensus version is sensible and straightforward
2. but having `GetAccounts()` return a `basics.AccountData` is more problematic. Let me explain why and get your thoughts.

We would need to modify the `<-chan AccountRow` output to return a `basics.AccountData`. We have two options for how to do this:
#### A) return it while also keeping the `models.Account` that is currently being returned
#### B) replace the `models.Account` that is being returned by this new `basics.AccountData`

Option **A** has the _advantage_ of making everything backwards compatible downstream. We only need to have the new functionality, such as `MinBalance`, opt in to this structure. It also has the _disadvantage_ of creating 2 parallel code paths for dealing with accounts. It will be difficult to know which of the two structures has the "real" account information.

Option **B** has the _advantage_ of providing the canonical account information. It has the  _disadvantage_ that current downstream consumers of `models.Account` will need to somehow consume a `basics.AccountData` instead. So we would need to either re-write the these consumers, or provide a solid adaptors that convert the models as needed. There are already similar adaptors in the `go-algorand` repo, but there are some subtle differences between `indexer` and `go-algorand` so I will not be able to use those adaptors "out of the box". Thus, we would likely also end up with parallel adaptors in each of the repos. In particular, here are the current consumers of the `models.Account` that `GetAccounts()` is currently providing:
* _account rewind functionality_: This seems to be basically broken as it cannot handle some relatively new transactions. Putting effort into writing an adaptor for this function seems to be misguided in this PR given the state of the functionality.
* endpoint `GET /v2/accounts/{account-id}`: would need to convert the models - but this seems like a reasonable goal, if we decide to embark down this path
* endpoint `GET /v2/accounts`: similar to the previous, we would need to convert the models, but it seems like a worthwhile endeavor

tzaffi: ## What am I proposing to do?

Because I believe any effort to upgrade `rewind()` is misguided in this PR, it suggests the "uglier" parallel model approach. But now that we are providing both models, it means that keeping the PR focused only on minbalance is simplest. I.e., I propose to have `GetAccounts()` output of `<-chan AccountRow` have its type modified as follows:
```go
type AccountRow struct {
	Account models.Account
        MinBalanceAccount basics.AccountData
	Error   error
}
```
the second field `MinBalanceAccount` is new. As the name suggests, the purpose of this field is to enable the calculation of `MinBalance()` so it would not be a "full" account, and the source of account truth would still be the `models.Account` field. I also propose to create new issues about ameliorating this non-ideal situation. The issues I would create would be:
1. Add adaptors to `indexer` which convert between `basics.AccountData` and `models.Account` (these would be similar to the ones already in `go-algorand`)
2. When `rewind()` gets fixed, it should use `basics.AccountData` and at the same time, `MinBalanceAccount` in `AccountRow` should be upgraded to the `Account` and have all the information that was in the old `models.Account`. The other downstream functionality would get their `models.Account` translated from a `basics.AccountData` right before returning their responses to the client.

tzaffi: ### An Even Quicker Approach

If we're aiming to get this PR out as soon as possible, an even quicker option is **C** which doesn't return any `basics.AccountData`. I could still push the calc of `MinBalance` closer to the view without a large effort. In this case, I would still create issues similar to the above.

winder: I think we need to wait for the ledger refactoring changes. Whether or not these approaches will continue working depend on changes that are yet to be made. For example, do we add the asset counts lazily or as a migration? If it's done lazily we may not always be able to compute the min balance depending on the REST API changes for the ledger refactoring (one proposal includes an option to return account data with asset/application data removed).

Regardless of that, I think we should skip the `basics.AccountData` conversion. Does it really help solve our problem? We're constructing it from metadata outside of algod, so we may end up with data inconsistencies based on how we reassemble it from Postgres.

tolikzinovyev: Agree with the first part, but ideally the database layer shouldn't construct `api` types imo. Constructing `basics.AccountData` from data in postgres is pretty straightforward, it's already done in `idb/postgres/internal/ledger_for_evaluator/`.

algoanne: I support Will's position that we wait for the ledger refactoring changes. We don't want to add this now, and then have updates to this work become a blocker for the ledger refactoring project (we have enough complex dependencies there already). 
People can continue to work without min balance being easily available for a little bit longer.

tzaffi: image

PR #829, opened 2022-01-19
Create Python Test for Generated Files

### LINKED ISSUE #405: integration test for generated files
## Summary
To ensure that development doesn't make generated files out of sync with their sources, write an integration test that checks that generation produces no change.

## Scope/Requirements
```
idb/postgres/postgres.go:7 //go:generate go run ../../cmd/texttosource/main.go postgres setup_postgres.sql reset.sql
```

## Urgency/Relative Priority
medium. editing setup_postgres.sql isn't too common.

AlgoStephenAkiki:
Resolves #405

Adds a python script that can be used to test that go generate did not
modify any files

## Summary

## Test Plan


winder (commented):
Can you change the test so that we don't need to update the md5 file? i.e. create it before running `go generate` and check that the file wasn't modified afterward.

The CI script also needs to be updated to call the new script.

winder (changes requested):
Can you change the test so that we don't need to update the md5 file? i.e. create it before running `go generate` and check that the file wasn't modified afterward.

The CI script also needs to be updated to call the new script.

Also please move this to the `misc` directory.

test/test_generate.py
```diff
@@ -0,0 +1,23 @@
...
+import sys
+from subprocess import check_output
+
+# Validates that certain files are not modified after running "go generate"
+
+# List of directories that contain generate.go files
+list_of_directories=["../idb/postgres/internal/schema"]
+
+for dir in list_of_directories:
+    os.system("/usr/local/go/bin/go generate {}".format(dir))
```
winder, line 13: Does this work without specifying the full path to `go`? It needs to be on the path for everything else so I think it's a fair assumption as long as it works.

AlgoStephenAkiki: It does not work without it unfortunately

winder (approved):

PR #833, opened 2022-01-24
Add startup check to verify existing DB data matches the algod network.

shiqizng:
## Summary
Write and check genesis hash during db load genesis. 

## Test Plan
Add a test to postgres_integration_test


idb/postgres/postgres.go
```diff
@@ -397,9 +397,29 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 // LoadGenesis is part of idb.IndexerDB
 func (db *IndexerDb) LoadGenesis(genesis bookkeeping.Genesis) error {
 	f := func(tx pgx.Tx) error {
+		// check networkdID
+		network, err := db.getMetastate(context.Background(), nil, schema.NetworkMetaStateKey)
```
tolikzinovyev, line 401: suggest making `getNetworkState()` and `setNetworkState()` similar to `?etImportState()` and `?etMigrationState()`.

idb/postgres/postgres.go
```diff
@@ -397,9 +397,29 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
+			if err != nil {
+				return fmt.Errorf("LoadGenesis() err: %w", err)
+			}
+		} else {
+			networkState, err := encoding.DecodeNetworkState([]byte(network))
+			if err != nil {
+				return fmt.Errorf("LoadGenesis() err: %w", err)
+			}
+			if networkState.NetworkID != string(genesis.Network) {
+				return fmt.Errorf("LoadGenesis() err: NetworkID does not match")
```
tolikzinovyev, line 417: Styling: "err: " means `err` is a variable. Just write `return fmt.Errorf("LoadGenesis() NetworkID does not match")`.

idb/postgres/postgres_integration_test.go
```diff
@@ -1876,3 +1876,29 @@ func TestTransactionsTxnAhead(t *testing.T) {
 		require.NoError(t, row.Error)
 	}
 }
+
+func TestNetworkIDCheckAtDBStartup(t *testing.T) {
+
```
tolikzinovyev, line 1881: empty line

tolikzinovyev (commented):
Is network ID enough? There might be some other parameters. Why not use the genesis hash?

shiqizng:
> Is network ID enough? There might be some other parameters. Why not use the genesis hash?

genesis hash isn't in bookkeeping.Genesis struct and so the hash is not available in LoadGenesis(genesis bookkeeping.Genesis).

tolikzinovyev:
> > Is network ID enough? There might be some other parameters. Why not use the genesis hash?
> 
> genesis hash isn't in bookkeeping.Genesis struct and so the hash is not available in LoadGenesis(genesis bookkeeping.Genesis). Maybe it's better to use Genesis.ID()?
> 
> ```
> // ID is the effective Genesis identifier - the combination
> // of the network and the ledger schema version
> func (genesis Genesis) ID() string {
> 	return string(genesis.Network) + "-" + genesis.SchemaID
> }
> ```

Seems like there is infrastructure for hashing `bookkeeping.Genesis`. https://github.com/algorand/go-algorand/blob/master/data/bookkeeping/genesis.go#L121 I just don't know how to invoke it.

shiqizng:
> > > Is network ID enough? There might be some other parameters. Why not use the genesis hash?
> > 
> > 
> > genesis hash isn't in bookkeeping.Genesis struct and so the hash is not available in LoadGenesis(genesis bookkeeping.Genesis). Maybe it's better to use Genesis.ID()?
> > ```
> > // ID is the effective Genesis identifier - the combination
> > // of the network and the ledger schema version
> > func (genesis Genesis) ID() string {
> > 	return string(genesis.Network) + "-" + genesis.SchemaID
> > }
> > ```
> 
> Seems like there is infrastructure for hashing `bookkeeping.Genesis`. https://github.com/algorand/go-algorand/blob/master/data/bookkeeping/genesis.go#L121 I just don't know how to invoke it.

I changed the check to use genesis hash. I removed network ID check since the hash includes the ID also.

idb/postgres/internal/encoding/encoding_test.go
```diff
@@ -4,11 +4,13 @@ import (
 	"math/rand"
 	"testing"
 
+	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger/ledgercore"
 	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/indexer/idb/postgres/internal/types"
```
tolikzinovyev, line 13: this should move below

idb/postgres/postgres.go
```diff
@@ -397,9 +398,26 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
 // LoadGenesis is part of idb.IndexerDB
 func (db *IndexerDb) LoadGenesis(genesis bookkeeping.Genesis) error {
 	f := func(tx pgx.Tx) error {
+		// check genesis hash
+		network, err := db.getNetworkState(context.Background(), nil)
```
tolikzinovyev, line 402: this should use `tx`

idb/postgres/postgres.go
```diff
@@ -397,9 +398,26 @@ func (db *IndexerDb) AddBlock(block *bookkeeping.Block) error {
...
+		if err == idb.ErrorNotInitialized {
+			networkState := types.NetworkState{
+				GenesisHash: crypto.HashObj(genesis),
+			}
+			err = db.setNetworkState(nil, &networkState)
```
tolikzinovyev, line 407: this should use `tx`

tolikzinovyev (changes requested):

tolikzinovyev (approved):
LGTM

PR #834, opened 2022-01-24
Changed OpenPostgres Function to include config

### LINKED ISSUE #826: Configurable Postgres Connection Pool Size
## Problem

It is possible to exhaust the DB connection handles, when that happens strange errors and unexpected behavior may occur. For example `FATAL: the database system is in recovery mode (SQLSTATE 57P03)`.

## Solution

Provide an indexer setting for configuring the pool size so that the local connection pool is exhausted before the connection pool. The connection pool is currently configured in the `OpenPostgres` function, we would replace `pgxpool.Connect` with `pgxpool.ConnectConfig`, and supply the `MaxConns` setting.

## Dependencies

No

AlgoStephenAkiki:
Resolves #826

Changed the OpenPostgres function to allow for setting the maximum and
minimum connection count in the connection pool.

### DIFF
--- a/cmd/algorand-indexer/main.go
+++ b/cmd/algorand-indexer/main.go
@@ -7,6 +7,7 @@ import (
 	"runtime/pprof"
 	"strings"
 
+	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/spf13/cobra"
 	//"github.com/spf13/cobra/doc" // TODO: enable cobra doc generation
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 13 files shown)


idb/dummy/dummy_factory.go
```diff
@@ -15,7 +16,7 @@ func (df dummyFactory) Name() string {
 }
 
 // Build is part of the IndexerFactory interface.
-func (df dummyFactory) Build(arg string, opts idb.IndexerDbOptions, log *log.Logger) (idb.IndexerDb, chan struct{}, error) {
+func (df dummyFactory) Build(config *pgxpool.Config, opts idb.IndexerDbOptions, log *log.Logger) (idb.IndexerDb, chan struct{}, error) {
```
winder, line 19: This interface should be DB agnostic, we shouldn't use `pgxpool` outside of the postgres implementation.

idb/idb.go
```diff
@@ -334,6 +334,10 @@ type ApplicationRow struct {
 // IndexerDbOptions are the options common to all indexer backends.
 type IndexerDbOptions struct {
 	ReadOnly bool
+	// Maximum connection number for connection pool
+	MaxConns int32
+	// Minimum connection number for connection pool
+	MinConns int32
```
winder, line 340: This is the better approach, but it looks like you didn't use it.

What is the use case for `MinConns`, do we need to set it?

cmd/algorand-indexer/main.go
```diff
@@ -74,20 +75,41 @@ var rootCmd = &cobra.Command{
...
+
+		// If the maximum is less than the minimum, just make them the same
+		if postgresMaxConns < postgresMinConns {
+			postgresMaxConns = postgresMinConns
+		}
+
+		postgresConfig.MaxConns = postgresMaxConns
+		postgresConfig.MinConns = postgresMinConns
+
+		db, ch, err := idb.IndexerDbByName("postgres", postgresConfig, opts, logger)
```
winder, line 112: This logic should be in `OpenPostgres`

winder (changes requested):
You can implement this without changing the `IndexerDbByName` interface. Could you take another pass and make sure the connection pool settings are passed into the DB and postgres specific work is done in that package?

PR #837, opened 2022-01-24
Changed configuration for postgres

### LINKED ISSUE #826: Configurable Postgres Connection Pool Size

AlgoStephenAkiki:
Resolves #826 

Changed database interface to include options and pushed implementation into postgres only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ There are different settings:
 
 One can set the maximum number of connections allowed in the local connection pool by using the `--max-conn` setting.  It is recommended to set this number to be below the database server connection pool limit.
  
-If the maximum number of connections/active queries is reached, subsequent connections will wait until a connection becomes available, or timeout according to the read-timeout setting
+If the maximum number of connections/active queries is reached, subsequent connections will wait until a connection becomes available, or timeout according to the read-timeout setting.
 
 # Settings
 
--- a/README.md
+++ b/README.md
@@ -127,6 +127,9 @@ There are different settings:
 ## Connection Pool Settings
 
 One can set the maximum number of connections allowed in the local connection pool by using the `--max-conn` setting.  It is recommended to set this number to be below the database server connection pool limit.
+ 
+If the maximum number of connections/active queries is reached, subsequent connections will wait until a connection becomes available, or timeout according to the read-timeout setting
+
 
 # Settings
 
--- a/README.md
+++ b/README.md
@@ -130,7 +130,6 @@ One can set the maximum number of connections allowed in the local connection po
  
 If the maximum number of connections/active queries is reached, subsequent connections will wait until a connection becomes available, or timeout according to the read-timeout setting
 
-
 # Settings
 
 Settings can be provided from the command line, a configuration file, or an environment variable
--- a/README.md
+++ b/README.md
@@ -117,11 +117,16 @@ When `--token your-token` is provided, an authentication header is required. For
 The `/metrics` endpoint is configured with the `--metrics-mode` option and configures if and how [Prometheus](https://prometheus.io/) formatted metrics are generated.
 
 There are different settings:
+
 | Setting | Description |
 | ------- | ----------- |
 | ON      | Metrics for each REST endpoint in addition to application metrics. |
 | OFF     | No metrics endpoint. |
 | VERBOSE | Separate metrics for each combination of query parameters. This option should be used with caution, there are many combinations of query parameters which could cause extra memory load depending on usage patterns. |
+
+## Connection Pool Settings
+
+One can set the maximum number of connections allowed in the local connection pool by using the `--max-conn` setting.  It is recommended to set this number to be below the database server connection pool limit.
 
 # Settings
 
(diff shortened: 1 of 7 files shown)


cmd/algorand-indexer/daemon.go
```diff
@@ -34,6 +34,8 @@ var (
 	tokenString      string
 	writeTimeout     time.Duration
 	readTimeout      time.Duration
+	maxConn          int32
```
shiqizng, line 37: should these be uint32 to disallow negative values?

AlgoStephenAkiki: I chose int32 because the postgres config units are int32...Should I change them here and cast them when creating the Postgres config?

winder: Unless there's a reason to allow negative values, I think it would be better to use uint32 here and let the flag parser enforce the negative bounds checking for us.

Sometimes interfaces use uint32 to avoid casts when using other built-ins... for example:
```
if len(connections) >= maxConn {
    ...
}
```

shiqizng (dismissed):

cmd/algorand-indexer/daemon.go
```diff
@@ -141,6 +145,7 @@ func init() {
 	daemonCmd.Flags().StringVarP(&metricsMode, "metrics-mode", "", "OFF", "configure the /metrics endpoint to [ON, OFF, VERBOSE]")
 	daemonCmd.Flags().DurationVarP(&writeTimeout, "write-timeout", "", 30*time.Second, "set the maximum duration to wait before timing out writes to a http response, breaking connection")
 	daemonCmd.Flags().DurationVarP(&readTimeout, "read-timeout", "", 5*time.Second, "set the maximum duration for reading the entire request")
+	daemonCmd.Flags().Uint32VarP(&maxConn, "max-conn", "", 4, "set the maximum connections allowed in the connection pool")
```
winder, line 148: 4 is the default? Seems small, I think making it optional and setting it to 0 would be better.

Please update the readme, we should mention that we recommend people should set this to a number smaller than their Postgres server's configuration.

winder (changes requested):
Looks good, just missing a README update and a test. Since this is all encapsulated in `OpenPostgres` the unit tests should be able to exercise the connection pool error.

cmd/algorand-indexer/daemon.go
AlgoStephenAkiki: 4 is the default number of the postgres pool but we can make it optional.

winder (changes requested):
This needs a test, everything else looks good.

cmd/algorand-indexer/daemon.go
```diff
@@ -141,6 +145,7 @@ func init() {
 	daemonCmd.Flags().StringVarP(&metricsMode, "metrics-mode", "", "OFF", "configure the /metrics endpoint to [ON, OFF, VERBOSE]")
 	daemonCmd.Flags().DurationVarP(&writeTimeout, "write-timeout", "", 30*time.Second, "set the maximum duration to wait before timing out writes to a http response, breaking connection")
 	daemonCmd.Flags().DurationVarP(&readTimeout, "read-timeout", "", 5*time.Second, "set the maximum duration for reading the entire request")
+	daemonCmd.Flags().Uint32VarP(&maxConn, "max-conn", "", 0, "set the maximum connections allowed in the connection pool")
```
winder, line 148: Is this accurate?
```suggestion
	daemonCmd.Flags().Uint32VarP(&maxConn, "max-conn", "", 0, "set the maximum connections allowed in the connection pool, if the maximum is reached subsequent connections will be queued and wait until a connection becomes available, or timeout according to the read-timeout setting")
```

AlgoStephenAkiki: The subsequent connections I believe will block, which is a little different than queued.  But the rest seems accurate.

winder: "queued and wait" is kinda the same thing as block? How about just saying wait?

AlgoStephenAkiki: Sure

README.md
```diff
@@ -117,12 +117,19 @@ When `--token your-token` is provided, an authentication header is required. For
...
 | ------- | ----------- |
 | ON      | Metrics for each REST endpoint in addition to application metrics. |
 | OFF     | No metrics endpoint. |
 | VERBOSE | Separate metrics for each combination of query parameters. This option should be used with caution, there are many combinations of query parameters which could cause extra memory load depending on usage patterns. |
 
+## Connection Pool Settings
+
+One can set the maximum number of connections allowed in the local connection pool by using the `--max-conn` setting.  It is recommended to set this number to be below the database server connection pool limit.
+ 
+If the maximum number of connections/active queries is reached, subsequent connections will wait until a connection becomes available, or timeout according to the read-timeout setting
```
winder, line 131:
```suggestion
If the maximum number of connections/active queries is reached, subsequent connections will wait until a connection becomes available, or timeout according to the read-timeout setting.
```

winder (approved):

PR #842, opened 2022-01-26
add genesis hash check to import

shiqizng:
## Summary
This is a follow up to #833 which check genesis hash only if db is not initialized. This PR fixes the problem by check genesis hash during every db startup.
## Test Plan
Add a test to postgres_integration_test to validate InitialImport() returns error when there is genesis hash conflict.


importer/helper.go
```diff
@@ -257,3 +244,38 @@ func (paths *blockTarPaths) Swap(i, j int) {
 	(*paths)[i] = (*paths)[j]
 	(*paths)[j] = t
 }
+
+// GetReader creates a genesis reader given the inputs
+func GetReader(genesisJSONPath string, client *algod.Client, l *log.Logger) io.Reader {
```
winder, line 249: Nice pattern! The only suggestion is to include what is being read somehow, like "GetGenesisFile". A `type GenesisFileReader io.Reader` could also help clarify what is going on and help link this function to InitialImport.

importer/helper.go
```diff
@@ -189,36 +186,26 @@ func loadGenesis(db idb.IndexerDb, in io.Reader) (err error) {
...
-func InitialImport(db idb.IndexerDb, genesisJSONPath string, client *algod.Client, l *log.Logger) bool {
+func InitialImport(db idb.IndexerDb, genesisReader io.Reader, l *log.Logger) (bool, error) {
 	_, err := db.GetNextRoundToAccount()
-
 	// Exit immediately or crash if we don't see ErrorNotInitialized.
 	if err != idb.ErrorNotInitialized {
-		maybeFail(err, l, "getting import state, %v", err)
-		return false
+		if err != nil {
+			maybeFail(err, l, "getting import state, %v", err)
```
winder, line 194: maybeFail calls exit, so you don't need the if/else. Just call maybeFail first.

winder: Alternatively, you can change this to return an error instead of exiting as you did with checkGenesisHash. You can decide what you prefer here.

tolikzinovyev: I think returning an error is definitely better. Below, `else` is also not needed. The whole block can move left.

importer/helper.go
```diff
@@ -189,36 +186,26 @@ func loadGenesis(db idb.IndexerDb, in io.Reader) (err error) {
...
-		maybeFail(err, l, "unable to fetch genesis from algod")
-		genesisReader = strings.NewReader(genesisString)
-	} else {
-		l.Fatal("Neither genesis file path or algod client provided for initial import.")
-	}
-
 	err = loadGenesis(db, genesisReader)
-	maybeFail(err, l, "%s: could not load genesis json, %v", genesisJSONPath, err)
-	return true
+	maybeFail(err, l, "could not load genesis json, %v", err)
```
tolikzinovyev, line 207: Similarly, we should return `err` if it's not `nil`.

importer/helper.go
```diff
@@ -257,3 +244,38 @@ func (paths *blockTarPaths) Swap(i, j int) {
...
+		genesisReader = strings.NewReader(genesisString)
+	} else {
+		l.Fatal("Neither genesis file path or algod client provided for initial import.")
+	}
+	return genesisReader
+}
+
+func checkGenesisHash(db idb.IndexerDb, genesisReader io.Reader, l *log.Logger) error {
+	network, err := db.GetNetworkState()
+	maybeFail(err, l, "unable to fetch network state from db")
```
tolikzinovyev, line 271: errors should be returned here too imo instead of crashing

tolikzinovyev (commented):
good catch

winder (dismissed):
LGTM

importer/helper.go
```diff
@@ -188,37 +185,28 @@ func loadGenesis(db idb.IndexerDb, in io.Reader) (err error) {
...
 	// Exit immediately or crash if we don't see ErrorNotInitialized.
 	if err != idb.ErrorNotInitialized {
-		maybeFail(err, l, "getting import state, %v", err)
-		return false
+		if err != nil {
+			return false, fmt.Errorf("getting import state, %v", err)
+		}
+		err = checkGenesisHash(db, genesisReader)
+		if err != nil {
+			l.WithError(err).Errorf("%v", err)
```
tolikzinovyev, line 198: this is probably not needed now

importer/helper.go
```diff
@@ -257,3 +245,44 @@ func (paths *blockTarPaths) Swap(i, j int) {
 	(*paths)[i] = (*paths)[j]
 	(*paths)[j] = t
 }
+
+// GetGenesisFile creates a reader from given the genesis file
```
tolikzinovyev, line 249:
```suggestion
// GetGenesisFile creates a reader from the given genesis file
```

importer/helper.go
```diff
@@ -257,3 +245,44 @@ func (paths *blockTarPaths) Swap(i, j int) {
...
+	} else {
+		l.Fatal("Neither genesis file path or algod client provided for initial import.")
+	}
+	return genesisReader
+}
+
+func checkGenesisHash(db idb.IndexerDb, genesisReader io.Reader) error {
+	network, err := db.GetNetworkState()
+	if err != nil {
+		return fmt.Errorf("unable to fetch network state from db %v", err)
```
tolikzinovyev, line 273: it's better to use `%w` instead of `%v`, so that `err` can be extracted later on

tolikzinovyev (approved):

PR #856, opened 2022-01-31
Fix network validation during database upgrade case.

shiqizng:
## Summary
This PR fixes the following issue received during 2.8.3 indexer upgrade,
`{"error":"unable to fetch network state from db GetNetworkState() err: accounting not initialized","level":"error","msg":"importer.EnsureInitialImport() error","time":"2022-01-31T01:35:47Z"}`
NetworkMetaStateKey is a new addition to metastate table. So, it needs to be set when it returns GetNetworkState() returns ErrorNotInitialized.

## Test Plan
Integration test is updated to validate that when network metadata is correctly initialized before GetNetworkState() calls.

### DIFF
--- a/importer/helper.go
+++ b/importer/helper.go
@@ -267,10 +267,6 @@ func GetGenesisFile(genesisJSONPath string, client *algod.Client, l *log.Logger)
 }
 
 func checkGenesisHash(db idb.IndexerDb, genesisReader io.Reader) error {
-	network, err := db.GetNetworkState()
-	if err != nil {
-		return fmt.Errorf("unable to fetch network state from db %w", err)
-	}
 	var genesis bookkeeping.Genesis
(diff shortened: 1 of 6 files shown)


importer/helper.go
```diff
@@ -280,6 +276,16 @@ func checkGenesisHash(db idb.IndexerDb, genesisReader io.Reader) error {
 	if err != nil {
 		return fmt.Errorf("error decoding genesis, %w", err)
 	}
+	network, err := db.GetNetworkState()
+	if err == idb.ErrorNotInitialized {
```
winder, line 280: If you use `errors.Is(err, idb.ErrorNotInitialized)` here, you can keep the wrapped error message above.

winder (approved):

PR #864, opened 2022-02-03
block-generator: Improve payment algorithm to allow sending many more transactions.

winder:
## Summary

Update the block-generator payment algorithm to cycle through accounts if the genesis account runs out of algos.

## Test Plan

Accounts were running out of algos during a 1hr run of the jumbo payment scenario, now it isn't.

### DIFF
--- a/cmd/block-generator/generator/generate.go
+++ b/cmd/block-generator/generator/generate.go
@@ -467,12 +467,14 @@ func (g *generator) generatePaymentTxnInternal(selection TxTypeID, round uint64,
 	case paymentTx:
 		receiveIndex = rand.Uint64() % g.numAccounts
 	case paymentAcctCreateTx:
+		// give new accounts extra algos for sending other transactions (optin/optout)
+		amount = minbal * 100
 		g.balances = append(g.balances, 0)
+		receiveIndex = g.numAccounts
 		// increment at the end in case it needs to be referenced later.
 		defer func() {
 			g.numAccounts++
 		}()
-		receiveIndex = g.numAccounts - 1
 	}
 
 	// Select a sender from genesis account
--- a/cmd/block-generator/generator/generate.go
+++ b/cmd/block-generator/generator/generate.go
@@ -451,13 +451,13 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
 	return g.generatePaymentTxnInternal(selection.(TxTypeID), round, intra)
 }
 
-const minbal = uint64(100000)
+const minBal = uint64(100000)
 
 func (g *generator) generatePaymentTxnInternal(selection TxTypeID, round uint64, intra uint64) (transactions.SignedTxn, transactions.ApplyData, error) {
 	defer g.recordData(track(selection))
 
 	// amounts
-	amount := minbal
+	amount := minBal
 	fee := uint64(1000)
 	total := amount + fee
 
@@ -468,7 +468,7 @@ func (g *generator) generatePaymentTxnInternal(selection TxTypeID, round uint64,
 		receiveIndex = rand.Uint64() % g.numAccounts
 	case paymentAcctCreateTx:
 		// give new accounts extra algos for sending other transactions (optin/optout)
-		amount = minbal * 100
+		amount = minBal * 100
 		g.balances = append(g.balances, 0)
 		receiveIndex = g.numAccounts
 		// increment at the end in case it needs to be referenced later.
@@ -480,7 +480,7 @@ func (g *generator) generatePaymentTxnInternal(selection TxTypeID, round uint64,
 	// Select a sender from genesis account
 	sendIndex := (g.numPayments + g.paymentOffset) % g.config.NumGenesisAccounts
 	// if the genesis account has insufficient balance... start checking others
-	for n := uint64(1); g.balances[sendIndex] < (total + minbal); n++ {
+	for n := uint64(1); g.balances[sendIndex] < (total + minBal); n++ {
 		g.paymentOffset++
 		sendIndex = (g.numPayments + g.paymentOffset) % g.numAccounts
 		fmt.Printf(" (%d) the sender account does not have enough algos for the transfer. idx %d, payment number %d\n", n, sendIndex, g.numPayments)
--- a/cmd/block-generator/generator/generate.go
+++ b/cmd/block-generator/generator/generate.go
@@ -459,7 +459,6 @@ func (g *generator) generatePaymentTxnInternal(selection TxTypeID, round uint64,
 	// amounts
 	amount := uint64(1)
 	fee := uint64(1000)
-	total := amount + fee
 
 	// Select a receiver
 	var receiveIndex uint64
@@ -476,6 +475,7 @@ func (g *generator) generatePaymentTxnInternal(selection TxTypeID, round uint64,
 			g.numAccounts++
 		}()
 	}
+	total := amount + fee
 
 	// Select a sender from genesis account
 	sendIndex := (g.numPayments + g.paymentOffset) % g.config.NumGenesisAccounts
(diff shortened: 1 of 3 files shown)


cmd/block-generator/generator/generate.go
```diff
@@ -450,33 +451,44 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
 	return g.generatePaymentTxnInternal(selection.(TxTypeID), round, intra)
 }
 
+const minbal = uint64(100000)
```
shiqizng, line 454: nit: please caps for const.

shiqizng (dismissed):

cmd/block-generator/generator/generate.go
winder: I don't want to export it, but I probably should have at least used camelcase

tolikzinovyev:
Are you using a config with genesis_account_balance = 10^15? With that number assuming 10K TPS, accounts shouldn't run out of money at least for one day. Also, I'd rather decrease the payment amount; it's simpler and doesn't change the nature of transactions.

winder:
> Are you using a config with genesis_account_balance = 10^15? With that number assuming 10K TPS, accounts shouldn't run out of money at least for one day. Also, I'd rather decrease the payment amount; it's simpler and doesn't change the nature of transactions.

Good call, I wanted to make sure new accounts got extra algod but didn't consider reducing the amount for the other transactions.

I think you already updated the genesis account balance.

tolikzinovyev:
> > Are you using a config with genesis_account_balance = 10^15? With that number assuming 10K TPS, accounts shouldn't run out of money at least for one day. Also, I'd rather decrease the payment amount; it's simpler and doesn't change the nature of transactions.
> 
> Good call, I wanted to make sure new accounts got extra algod but didn't consider reducing the amount for the other transactions.
> 
> I think you already updated the genesis account balance.

I did update it, but are you sure you are not using a custom config?

cmd/block-generator/generator/generate.go
```diff
@@ -450,33 +451,44 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
...
+		defer func() {
+			g.numAccounts++
+		}()
 	}
 
-	// Always send from a genesis account.
-	sendIndex := g.numPayments % g.config.NumGenesisAccounts
+	// Select a sender from genesis account
+	sendIndex := (g.numPayments + g.paymentOffset) % g.config.NumGenesisAccounts
+	// if the genesis account has insufficient balance... start checking others
```
tolikzinovyev, line 482: this is unnecessary now, right?

winder: Depending on the configuration this could still be needed. Not quite as obvious as killing the program, but the print statement still makes it pretty clear when you're using this fallback.

The checked-in scenarios shouldn't have a problem even for very long tests.

cmd/block-generator/generator/generate.go
```diff
@@ -450,33 +451,44 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
...
+	for n := uint64(1); g.balances[sendIndex] < (total + minBal); n++ {
+		g.paymentOffset++
+		sendIndex = (g.numPayments + g.paymentOffset) % g.numAccounts
+		fmt.Printf(" (%d) the sender account does not have enough algos for the transfer. idx %d, payment number %d\n", n, sendIndex, g.numPayments)
```
tolikzinovyev, line 486: this might not apply to `sendIndex` since it was just changed

cmd/block-generator/generator/generate.go
```diff
@@ -450,33 +451,44 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
...
+		defer func() {
+			g.numAccounts++
+		}()
 	}
+	total := amount + fee
 
-	// Always send from a genesis account.
-	sendIndex := g.numPayments % g.config.NumGenesisAccounts
+	// Select a sender from genesis account
+	sendIndex := (g.numPayments + g.paymentOffset) % g.config.NumGenesisAccounts
```
tolikzinovyev, line 481: why is `paymentOffset` used here?

cmd/block-generator/generator/generate.go
```diff
@@ -450,33 +451,44 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
 	return g.generatePaymentTxnInternal(selection.(TxTypeID), round, intra)
 }
 
+const minBal = uint64(100000)
```
tolikzinovyev, line 454: is it possible to use the value from the protocol config?

cmd/block-generator/generator/generate.go
```diff
@@ -450,33 +451,44 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
...
+		}()
 	}
+	total := amount + fee
 
-	// Always send from a genesis account.
-	sendIndex := g.numPayments % g.config.NumGenesisAccounts
+	// Select a sender from genesis account
+	sendIndex := (g.numPayments + g.paymentOffset) % g.config.NumGenesisAccounts
+	// if the genesis account has insufficient balance... start checking others
+	for n := uint64(1); g.balances[sendIndex] < (total + minBal); n++ {
```
tolikzinovyev, line 483: this loop might be very slow, I'm not sure I would trust performance measurements if we got here

cmd/block-generator/generator/generate.go
tolikzinovyev: I'm not sure what configurations you are referring to. There should be enough microalgos for any desirable configuration.

cmd/block-generator/generator/generate.go
winder: I was thinking we could have gotten into a situation where the genesis accounts were exhausted and we were now sending payments from non-genesis accounts. In that scenario keeping track of an offset would avoid checking the same non-genesis accounts over and over again.

After the other changes, I think it would take a very long to hit this condition, and by then there are probably many other problems with the generator that I'm not yet aware of.

cmd/block-generator/generator/generate.go
```diff
@@ -42,13 +42,21 @@ const (
...
 )
 
+// consensus parameters.
+var (
+	minBal uint64
+	fee    uint64
+)
+
+func init() {
+	minBal = config.Consensus[protocol.ConsensusFuture].MinBalance
```
tolikzinovyev, line 56: it might be better to move this to `generatePaymentTxnInternal()` where `g.protocol` can be accessed

cmd/block-generator/generator/generate.go
```diff
@@ -453,30 +461,37 @@ func (g *generator) generatePaymentTxn(round uint64, intra uint64) (transactions
...
+			g.numAccounts++
+		}()
 	}
+	total := amount + fee
 
-	// Always send from a genesis account.
+	// Select a sender from genesis account
 	sendIndex := g.numPayments % g.config.NumGenesisAccounts
+	// if the genesis account has insufficient balance... start checking others
+	for g.balances[sendIndex] < (total + minBal) {
```
tolikzinovyev, line 487: use `if`?

cmd/block-generator/generator/generate.go
winder: I moved fee to be a constant to avoid referencing the consensus params in those places, and took this suggestion for min balance. Its tempting to take it all out and just let indexer crash with the overspend error, since generator doesn't do anything special in that case now.

tolikzinovyev: It's just "future" is specified in 3 places now, which is not optimal. I would add consensus params struct to the generator struct, so that all parameters can easily be referenced.

winder: good idea, added the params to the generator struct.

tolikzinovyev (approved):

PR #871, opened 2022-02-08
Include import-validator as a hidden algorand-indexer subcommand.

winder:
## Summary

Add a `./algorand-indexer import-validator` subcommand which behaves the same as the current `import-validator` utility.

## Test Plan

Manual testing:
1. Start postgres in a terminal: `docker run -it --rm --name some-postgres -p 5555:5432 -e POSTGRES_PASSWORD=algorand -e POSTGRES_USER=algorand -e POSTGRES_DB=algorand postgres`
2. Initialize a ledger directory with the current import-validator: `./import-validator --postgres "user=algorand password=algorand host=localhost dbname=algorand port=5555 sslmode=disable" --algod-net https://<URL> --algod-token <token> --algod-ledger ledger`
3. Kill import-validator and start the new subcommand: `./algorand-indexer import-validator --postgres "user=algorand password=algorand host=localhost dbname=algorand port=5555 sslmode=disable" --algod-net https://<URL> --algod-token <token> --algod-ledger ../import-validator/ledger`


tolikzinovyev:
What's the rationale?

winder:
Bundling in the most useful utilities should make it easier to perform release testing and ensure the utilities match the algorand-indexer version. I'm planning to include `block-generator` and `validator` as well.

winder:
> What's the rationale?

@tolikzinovyev Bundling in the most useful utilities should make it easier to perform release testing and ensure the utilities match the algorand-indexer version. There is also some automation I want to add, which will be simpler if I don't have to rebuild the utilities when a new release is detected. I'm planning to include `block-generator` and `validator` next.

tolikzinovyev:
could you give one specific example where this is useful?

winder:
> could you give one specific example where this is useful?

I would like to setup the `import-validator` on an EC2 instance, along with a cron job that fetches the latest release. With this tool in the latest release this is simple to automate. If it isn't in the release the script would also need a full development environment and would have to guess about which code to use (usually the tip of master, but not necessarily).

There is some release automation with similar issues, they become more straight-forward when the utilities are bundled together:
1) Grabbing a new release and starting a new indexer deployment is simple, but once caught up how do we run `validator`?
2) The performance benchmarks can get the latest version, but when if the block generator also had changed?

cmd/import-validator/core/service.go
```diff
@@ -0,0 +1,359 @@
...
+// ImportValidatorArgs required parameters for the import validator servies.
+type ImportValidatorArgs struct {
+	AlgodAddr       string
+	AlgodToken      string
+	AlgodLedger     string
+	PostgresConnStr string
+}
+
+// Run is a blocking call that runs the import validator service.
+func Run(args ImportValidatorArgs) {
```
tolikzinovyev, line 40: nit: put it at the bottom

tolikzinovyev (approved):
The code looks good, but imo it seems like too much automation. You could have a script that takes a git hash as input and does everything else.

Issue #374, opened 2021-03-03
Rewind: Remove or Implement /v2/assets/{asset-id}/balances 'round' query parameter


winder:
## Summary
/v2/assets/{asset-id}/balances have a **round** parameter in the spec which is defined as "Include results for the specified round."

The parameter is currently unused so we should either remove the parameter or implement the feature. Considering that nobody has noticed this parameter, we should probably remove it.

nullun:
I was just about to file a new issue on this round parameter not working correctly! I was hoping to use it to retrieve a snapshot of account balances for a particular historic round. If there's no demand for it I'll look at alternatives, but I think it would be nice to have.

jiujiteiro:
+1 for implementing. I saw the parameter in the spec and was hoping to use as well to pull historical data on ASAs.

winder:
This unimplemented parameter is removed by https://github.com/algorand/indexer/pull/872

PR #872, opened 2022-02-08
REST API changes for unlimited assets

cce:
## Summary

Adds new REST endpoints and "exclude" parameter for account information endpoint, similar to https://github.com/algorand/go-algorand/pull/3542. Relies on #870

Update endpoint: GET /v2/accounts/{address}
- Supports new query parameter “exclude”, which can be a combination of `all`, `created-apps`, `created-assets`, `apps-local-state`, `assets`
- Removes empty “creator” string from AssetHolding type — was not implemented (removed associated TODO)

Update endpoint: GET /v2/assets/{asset-id}/balances 
- Remove "round" query parameter — I noticed support is not implemented, and the parameter was being ignored

New endpoints:
GET /v2/accounts/{address}/created-apps
GET /v2/accounts/{address}/created-assets
GET /v2/accounts/{address}/assets
GET /v2/accounts/{address}/apps-local-state
- Supported query parameters: asset-id / application-id (look up a single asset), include-all (include deleted), limit, next (for pagination)

## Test Plan

Extend existing tests, add additional tests to exercise features of all new API endpoints.


idb/idb.go
```diff
@@ -306,9 +309,12 @@ type AssetRow struct {
...
 type AssetBalanceQuery struct {
-	AssetID  uint64
-	AmountGT *uint64 // only rows > this
-	AmountLT *uint64 // only rows < this
+	AssetID   uint64
+	AssetIDGT uint64
+	AmountGT  *uint64 // only rows > this
+	AmountLT  *uint64 // only rows < this
+
+	Address []byte
```
tolikzinovyev, line 317: this should be `basics.Address`

idb/idb.go
```diff
@@ -332,12 +338,27 @@ type AssetBalanceRow struct {
...
-// ApplicationRow is metadata relating to one application in an application query.
+// ApplicationRow is metadata and global state (AppParams) relating to one application in an application query.
 type ApplicationRow struct {
 	Application models.Application
 	Error       error
 }
 
+// ApplicationQuery is a parameter object used for query local and global application state.
+type ApplicationQuery struct {
+	Address                  []byte
```
tolikzinovyev, line 349: this should be `basics.Address`

idb/postgres/postgres.go
```diff
@@ -1978,21 +1968,29 @@ func (db *IndexerDb) buildAccountQuery(opts idb.AccountQueryOptions) (query stri
...
 		}
-		// app
-		query += `, qapp AS (SELECT app.creator as addr, json_agg(app.index) as papps, json_agg(app.params) as ppa, json_agg(app.created_at) as app_created_at, json_agg(app.closed_at) as app_closed_at, json_agg(app.deleted) as app_deleted FROM app JOIN qaccounts ON qaccounts.addr = app.creator WHERE coalesce(app.deleted, false) = false GROUP BY 1)`
-		// app localstate
-		query += `, qls AS (SELECT la.addr, json_agg(la.app) as lsapps, json_agg(la.localstate) as lsls, json_agg(la.created_at) as ls_created_at, json_agg(la.closed_at) as ls_closed_at, json_agg(la.deleted) as ls_deleted FROM account_app la JOIN qaccounts ON qaccounts.addr = la.addr WHERE coalesce(la.deleted, false) = false GROUP BY 1)`
+		if opts.IncludeAppParams {
+			// app
+			query += `, qapp AS (SELECT app.creator as addr, json_agg(app.index) as papps, json_agg(app.params) as ppa, json_agg(app.created_at) as app_created_at, json_agg(app.closed_at) as app_closed_at, json_agg(app.deleted) as app_deleted FROM app JOIN qaccounts ON qaccounts.addr = app.creator WHERE coalesce(app.deleted, false) = false GROUP BY 1)`
```
tolikzinovyev, line 1988: `coalesce` is not needed anymore since `deleted` is `NOT NULL`

`WHERE NOT deleted`

idb/postgres/postgres.go
```diff
@@ -2169,7 +2189,11 @@ func (db *IndexerDb) AssetBalances(ctx context.Context, abq idb.AssetBalanceQuer
 	if len(whereParts) > 0 {
 		query += " WHERE " + strings.Join(whereParts, " AND ")
 	}
-	query += " ORDER BY addr ASC"
+	if abq.Address != nil {
+		query += " ORDER BY assetid ASC"
```
tolikzinovyev, line 2193: perhaps this should be unconditional?

idb/postgres/postgres.go
```diff
@@ -2362,6 +2386,113 @@ func (db *IndexerDb) yieldApplicationsThread(rows pgx.Rows, out chan idb.Applica
...
+		}
+		var rec idb.AppLocalStateRow
+		rec.AppLocalState.Id = index
+		rec.AppLocalState.OptedInAtRound = created
+		rec.AppLocalState.ClosedOutAtRound = closed
+		rec.AppLocalState.Deleted = deleted
+
+		ls, err := encoding.DecodeAppLocalState(statejson)
+		if err != nil {
+			rec.Error = fmt.Errorf("app=%d json err, %v", index, err)
```
tolikzinovyev, line 2480: nit: `%v` -> `%w`

idb/idb.go
```diff
@@ -332,12 +338,27 @@ type AssetBalanceRow struct {
...
+	ApplicationID            uint64
+	ApplicationIDGreaterThan uint64
+	IncludeDeleted           bool
+	Limit                    uint64
+}
+
+// AppLocalStateRow is metadata and local state (AppLocalState) relating to one application in an application query.
+type AppLocalStateRow struct {
+	AppLocalState models.ApplicationLocalState
```
tolikzinovyev, line 358: I really think the database layer shouldn't construct and return REST representation.

api/handlers.go
```diff
@@ -145,6 +145,34 @@ func (si *ServerImplementation) MakeHealthCheck(ctx echo.Context) error {
...
+			opts.IncludeAppParams = false
+		case "assets":
+			opts.IncludeAssetHoldings = false
+		case "created-assets":
+			opts.IncludeAssetParams = false
+		case "apps-local-state":
+			opts.IncludeAppLocalState = false
+		case "created-apps":
+			opts.IncludeAppParams = false
+		case "none", "":
```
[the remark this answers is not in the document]
cce: I couldn't find anywhere that the swagger-generated code creates or checks against the enum list, so I just added them as a convenience, in case they were provided, but I could make them explicit.. or get rid of this ..

idb/idb.go
cce: I had assumed it was better to keep the existing pattern in idb.go for []byte addresses than introduce an inconsistency, wdyt?

idb/postgres/postgres.go
cce: oh wow, OK, I can update a lot of coalesce(X.deleted, false) clauses throughout the query builder then. Just found #520 where this change was first made

cce: this means I can convert `json_agg(coalesce(x.deleted, false))` everywhere to just `json_agg(x.deleted)` yes?

idb/postgres/postgres.go
cce: there are existing code paths to look up all the balances for a specific asset that don't set this.

idb/idb.go
cce: Using the generated models is how AccountRow and ApplicationRow does it, so I added AppLocalStateRow to match ApplicationRow. However the TxnRow, AssetRow and AssetBalanceRow provide their own encoding. I have no strong feelings either way whether the new type I introduced here should have gone one or the other way ... @winder?

idb/idb.go
tolikzinovyev: I know this will make an inconsistency, but I think it's ok as long as the code is moving in the right direction.

idb/postgres/postgres.go
tolikzinovyev: looks good!

api/converter_utils.go
```diff
@@ -548,6 +548,30 @@ func assetParamsToAssetQuery(params generated.SearchForAssetsParams) (idb.Assets
...
+		}
+		appGreaterThan = agt
+	}
+
+	return idb.ApplicationQuery{
+		ApplicationID:            uintOrDefault(params.ApplicationId),
+		ApplicationIDGreaterThan: appGreaterThan,
+		Address:                  addr,
+		IncludeDeleted:           boolOrDefault(params.IncludeAll),
+		Limit:                    min(uintOrDefaultValue(params.Limit, defaultApplicationsLimit), maxApplicationsLimit),
```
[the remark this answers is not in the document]
winder: Configurable limits would be good. We decided to truncate the max limit early on, I don't see a need to change the design for this PR.

api/handlers.go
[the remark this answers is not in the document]
cce: deleted!

api/converter_utils.go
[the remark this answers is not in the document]
cce: OK, I have moved them to configurable limits in e41d91fd2e8ae08bc363abd30fd58c2591694cc4

api/indexer.oas2.json
```diff
@@ -455,9 +651,6 @@
           {
             "$ref": "#/parameters/next"
           },
-          {
```
tolikzinovyev, line 458: was this also never implemented?

cce: Yes, while looking at which API endpoints supported the "Round" parameter, I noticed that params.Round was never used in the [implementation of /v2/assets/{asset-id}/balances](https://github.com/algorand/indexer/blob/develop/api/handlers.go#L448-L486) but listed as a parameter, so I got rid of it. CC @winder

api/handlers.go
```diff
@@ -742,6 +917,47 @@ func (si *ServerImplementation) fetchAssetBalances(ctx context.Context, options
...
+		var assetbalchan <-chan idb.AssetBalanceRow
+		assetbalchan, round = si.db.AssetBalances(ctx, options)
+
+		for row := range assetbalchan {
+			if row.Error != nil {
+				return row.Error
+			}
+
+			addr := basics.Address{}
+			if len(row.Address) != len(addr) {
```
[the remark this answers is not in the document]
cce: I started trying this, from changing `decodeAddress` to return a basics.Address, but then abandoned it because I didn't want to change things a lot of different places. I think I will leave this to someone in the future to make an across-the-board change rather than have different checks/byte-to-address conversions take place in different places through the handles and DB code.

api/indexer.oas2.json
winder: Thanks. This closes #374 

cc: @algoanne

idb/postgres/postgres.go
```diff
@@ -1107,37 +1107,22 @@ func (db *IndexerDb) yieldAccountsThread(req *getAccountsRequest) {
...
-				&appParamIndexes, &appParams, &appCreatedBytes, &appClosedBytes, &appDeletedBytes, &localStateAppIds, &localStates,
-				&localStateCreatedBytes, &localStateClosedBytes, &localStateDeletedBytes,
-			)
-		} else {
-			err = req.rows.Scan(
-				&addr, &microalgos, &rewardstotal, &createdat, &closedat, &deleted, &rewardsbase, &keytype, &accountDataJSONStr,
-				&appParamIndexes, &appParams, &appCreatedBytes, &appClosedBytes, &appDeletedBytes, &localStateAppIds, &localStates,
-				&localStateCreatedBytes, &localStateClosedBytes, &localStateDeletedBytes,
-			)
+		// build list of columns to scan using include options like buildAccountQuery
```
winder, line 1110: nice trick :+1:

idb/postgres/postgres.go
```diff
@@ -1719,6 +1720,69 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
 	return out, round
 }
 
+func (db *IndexerDb) checkAccountResourceLimit(ctx context.Context, tx pgx.Tx, opts idb.AccountQueryOptions) error {
```
winder, line 1723: You can skip the checks if all of the include flags are already false.

idb/postgres/postgres.go
```diff
@@ -2189,6 +2284,113 @@ func (db *IndexerDb) yieldApplicationsThread(rows pgx.Rows, out chan idb.Applica
 	}
 }
 
+// AppLocalState is part of idb.IndexerDB
+func (db *IndexerDb) AppLocalState(ctx context.Context, filter idb.ApplicationQuery) (<-chan idb.AppLocalStateRow, uint64) {
+	out := make(chan idb.AppLocalStateRow, 1)
+
+	query := `SELECT app, addr, localstate, created_at, closed_at, deleted FROM account_app `
+
+	const maxWhereParts = 30
```
winder, line 2293:
```suggestion
	const maxWhereParts = 4
```

idb/postgres/postgres.go
```diff
@@ -2066,40 +2161,40 @@ func (db *IndexerDb) yieldAssetBalanceThread(rows pgx.Rows, out chan<- idb.Asset
...
 	out := make(chan idb.ApplicationRow, 1)
-	if filter == nil {
-		out <- idb.ApplicationRow{Error: fmt.Errorf("no arguments provided to application search")}
-		close(out)
-		return out, 0
-	}
 
 	query := `SELECT index, creator, params, created_at, closed_at, deleted FROM app `
 
 	const maxWhereParts = 30
```
winder, line 2169:
```suggestion
	const maxWhereParts = 4
```

winder (changes requested):
Looks really nice overall. I don't think I had any major feedback here

cmd/algorand-indexer/daemon.go
```diff
@@ -148,6 +159,19 @@ func init() {
...
+	daemonCmd.Flags().Uint32VarP(&defaultAccountsLimit, "default-accounts-limit", "", 100, "set the default Limit parameter for querying accounts, if none is provided")
+	daemonCmd.Flags().Uint32VarP(&maxAssetsLimit, "max-assets-limit", "", 1000, "set the maximum allowed Limit parameter for querying assets")
+	daemonCmd.Flags().Uint32VarP(&defaultAssetsLimit, "default-assets-limit", "", 100, "set the default Limit parameter for querying assets, if none is provided")
+	daemonCmd.Flags().Uint32VarP(&maxBalancesLimit, "max-balances-limit", "", 10000, "set the maximum allowed Limit parameter for querying balances")
+	daemonCmd.Flags().Uint32VarP(&defaultBalancesLimit, "default-balances-limit", "", 1000, "set the default Limit parameter for querying balances, if none is provided")
+	daemonCmd.Flags().Uint32VarP(&maxApplicationsLimit, "max-applications-limit", "", 1000, "set the maximum allowed Limit parameter for querying applications")
+	daemonCmd.Flags().Uint32VarP(&defaultApplicationsLimit, "default-applications-limit", "", 100, "set the default Limit parameter for querying applications, if none is provided")
```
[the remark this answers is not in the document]
winder: Could you add these new options to the table in the README?

Makefile
```diff
@@ -11,6 +11,8 @@ GOLDFLAGS += -X github.com/algorand/indexer/version.CompileTime=$(shell date -u
 GOLDFLAGS += -X github.com/algorand/indexer/version.GitDecorateBase64=$(shell git log -n 1 --pretty="%D"|base64|tr -d ' \n')
 GOLDFLAGS += -X github.com/algorand/indexer/version.ReleaseVersion=$(shell cat .version)
 
+COVERPKG := $(shell go list ./...  | grep -v '/cmd/' | egrep -v '(testing|test|mocks)$$' |  paste -s -d, - )
```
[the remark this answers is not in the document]
cce: This makes the coverage reports account for cross-package coverage. Basically I wanted quereis made by api/handlers_e2e_test.go to show up in coverage for the idb/postgres package

api/server.go
```diff
@@ -37,6 +37,35 @@ type ExtraOptions struct {
 
 	// ReadTimeout is the maximum duration for reading the entire request, including the body.
 	ReadTimeout time.Duration
+
+	// MaxAccountsAPIResults is the maximum number of combined AppParams, AppLocalState, AssetParams,
+	// and AssetHolding resources per address that can be returned by the /v2/accounts endpoints.
+	// If an address exceeds this number, a 400 error is returned. Zero means unlimited.
+	MaxAccountsAPIResults uint64
```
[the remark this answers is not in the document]
cce: This is the same name as used by algod — I had asked @barnjamin if the word "creatables" was widely used and it is not. Resources is ambiguous but kind of OK by me. Agreed "Max Accounts API Results" is confusing especially since you have other various configurable limits in the indexer now. Maybe `MaxNestedAccountResources` ...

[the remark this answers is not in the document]
winder: I agree with you. API results just seems like an alias for limit, but I'm not a huge fan of my suggestions either. Maybe MaxAccountNestedObjects?

cmd/algorand-indexer/daemon.go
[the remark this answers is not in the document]
cce: OK, I did, plus I noticed write-timeout and read-timeout were not in the README Settings section and added them.

api/server.go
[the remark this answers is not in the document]
cce: That seems good.. will switch

api/handlers_e2e_test.go
```diff
@@ -87,30 +115,582 @@ func TestApplicationHandler(t *testing.T) {
...
+		exclude   []string
+		errStatus int
+	}{
+		{address: test.AccountA, exclude: []string{"all"}},
+		{address: test.AccountA, exclude: []string{"created-assets", "created-apps", "apps-local-state", "assets"}},
+		{address: test.AccountB, exclude: []string{"assets", "apps-local-state"}},
+		{address: test.AccountA, exclude: []string{"created-assets"}, errStatus: http.StatusBadRequest},
+		{address: test.AccountA, exclude: []string{"created-apps"}, errStatus: http.StatusBadRequest},
+		{address: test.AccountA, exclude: []string{"apps-local-state"}, errStatus: http.StatusBadRequest},
+		{address: test.AccountA, exclude: []string{"assets"}, errStatus: http.StatusBadRequest},
```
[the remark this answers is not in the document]
cce: It does below on lines 488-498 — if there is a bad request error, it decodes the error response and checks the counts inside match

api/handlers.go
```diff
@@ -40,31 +40,9 @@ type ServerImplementation struct {
...
-const defaultAssetsLimit = 100
 
-// Asset Balances
-const maxBalancesLimit = 10000
-const defaultBalancesLimit = 1000
-
-// Applications
-const maxApplicationsLimit = 1000
-const defaultApplicationsLimit = 100
+	opts ExtraOptions
```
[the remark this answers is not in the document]
cce: I figured, maybe sometime later you will want to use this flag to additionally disable rewind for some other reason, and thought any change was kind of out of scope

idb/postgres/postgres.go
cce: thanks, updated

PR #873, opened 2022-02-09
Initial Runtime data structures

AlgoStephenAkiki:
Resolves algorand/go-algorand#3581

Added a runtime data structure that can be used to disable endpoints and
query parameters if needed.

Currently the data structure enables everything, so no functional change
has occurred to the user.


api/disabled_parameters_test.go
```diff
@@ -0,0 +1,447 @@
...
+
+func (er *testingErrorReporter) Reset() {
+	er.ErrorFCalledCount = 0
+}
+
+type testingEchoContext struct {
+	InternalQueryParams  url.Values
+	InternalFormParams   url.Values
+	ForceFormParamsError bool
+}
```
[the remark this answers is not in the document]
winder: It looks like you can force a form error: https://github.com/labstack/echo/blob/fdacff0d93d54e8065ac7f5072041146f054a201/context_test.go#L579

api/handlers.go
```diff
@@ -148,6 +150,18 @@ func (si *ServerImplementation) MakeHealthCheck(ctx echo.Context) error {
...
+	switch rc {
+	case verifyIsGood:
+		// Do nothing
+		break
+	case verifyFailedParameter:
+		return badRequest(ctx, fmt.Sprintf("provided disabled parameter: %s", dpName))
+
+	case verifyFailedEndpoint:
+		return badRequest(ctx, "endpoint is disabled")
+	}
```
winder, line 163: Is there a way to compare the swagger "path" with the context path?

If yes, I think there is a way to implement this as a middleware. That would let us skip this boilerplate and handle it globally.

api/server.go
```diff
@@ -77,12 +77,19 @@ func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError
 		middleware = append(middleware, middlewares.MakeAuth("X-Indexer-API-Token", options.Tokens))
 	}
 
+	swag, err := generated.GetSwagger()
```
winder, line 80: Technically there is also `common.GetSwagger()`, but I don't think the Health endpoint has any parameters.

I only bring it up because you included the boilerplate code on MakeHealthCheck but there's no way for it to be part of the DisableMap.

api/disabled_parameters.go
```diff
@@ -0,0 +1,151 @@
+package api
+
+import (
+	"fmt"
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/labstack/echo/v4"
```
winder, line 6:
```suggestion
	"fmt"
	
	"github.com/getkin/kin-openapi/openapi3"
	"github.com/labstack/echo/v4"
```

api/disabled_parameters.go
```diff
@@ -0,0 +1,151 @@
...
+
+				disabledParameter := DisabledParameter{
+					Name:     pref.Value.Name,
+					Disabled: false,
+				}
+				// TODO how to enable it to be disabled
+
+				if pref.Value.Required {
+					disabledList.RequiredParams = append(disabledList.RequiredParams, disabledParameter)
+					fmt.Printf(" -> Param (Req): %s\n", pref.Value.Name)
```
winder, line 53: Use the logger for these, they could be verbose messages but probably not necessary for the standard output.

api/disabled_parameters.go
```diff
@@ -0,0 +1,151 @@
...
+	"fmt"
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/labstack/echo/v4"
+)
+
+// DisabledParameter holds the name and disabled status of a parameter
+type DisabledParameter struct {
+	Name     string
+	Disabled bool
+}
```
winder, line 13: This could be `map[string]bool` instead of a struct. With that format, you could iterate over the set parameters and check if they're enabled or not.

The current algorithm iterates over the disabled parameters and checks if they were used. In the worst case, something like a transaction search could have something like 20 checks even if the user didn't set any parameters.

api/disabled_parameters.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (dl *DisabledList) verify(ctx echo.Context, log DisabledParameterErrorReporter) (VerifyRC, string) {
+
+	// If any of the Required Params are disabled we have to fail the whole
+	// endpoint
+	for _, dp := range dl.RequiredParams {
+		if dp.Disabled {
+			return verifyFailedEndpoint, ""
+		}
+	}
```
winder, line 116: Instead of storing `RequiredParams` could you store an `EndpointDisabled` flag?

winder (changes requested):
Looks great, this approach is a lot nicer than I realized from your description. I had some specific suggestions that might reduce the boilerplate even more, let me know what you think.

api/disabled_parameters_test.go
```diff
@@ -0,0 +1,447 @@
...
+	require.Equal(t, 1, er.ErrorFCalledCount)
+	require.Empty(t, rstr)
+
+}
+
+// TestFailingQueryParam tests that disabled parameters provided via
+// the QueryParams() function of the context are appropriately handled
+func TestFailingQueryParam(t *testing.T) {
+	dm := NewDisabledMap()
+
```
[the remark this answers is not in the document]
AlgoStephenAkiki: Done

api/disabled_parameters.go
AlgoStephenAkiki: Done

api/disabled_parameters.go
AlgoStephenAkiki: Done

api/disabled_parameters.go
AlgoStephenAkiki: Simply removed them...unnecessary really

api/server.go
AlgoStephenAkiki: MakeHealthCheck doesn't have the checking code...will it ever need it?

api/disabled_parameters_test.go
```diff
@@ -0,0 +1,447 @@
...
+	"net/url"
+	"testing"
+
+	"github.com/labstack/echo/v4"
+	"github.com/stretchr/testify/require"
+)
+
+type testingErrorReporter struct {
+	ErrorFCalledCount int
+}
```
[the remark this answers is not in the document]
AlgoStephenAkiki: I do like the error reporter defined here because I get to count how many times it was called as opposed to the log message (which might change in the future).

api/handlers.go
AlgoStephenAkiki: I couldn't find a way that didn't involve codegen changes but we could talk more on a call to see if I missed something.

api/server.go
winder: No, we can ignore it there

api/disabled_parameters.go
```diff
@@ -0,0 +1,149 @@
...
+		if queryValue != "" {
+			// If the query value is non-zero, and it was disabled, we should return false
+			return verifyFailedParameter, paramName
+		}
+
+		if formErr != nil {
+			continue
+		}
+
+		formValue := formParams.Get(paramName)
```
shiqizng, line 141: can queryParams or formParams be nil?

api/handlers.go
```diff
@@ -553,6 +691,17 @@ func (si *ServerImplementation) LookupTransaction(ctx echo.Context, txid string)
...
```
shiqizng, line 704: would it be better to put this check in a helper method to reduce repeated code?

api/disabled_parameters.go
```diff
@@ -0,0 +1,149 @@
...
+				parameterIsDisabled := false
+				if !parameterIsDisabled {
+					// If the parameter is not disabled, then we don't need
+					// to do anything
+					continue
+				}
+
+				if pref.Value.Required {
+					// If an endpoint config required parameter is disabled, then the whole endpoint is disabled
+					endpointConfig.EndpointDisabled = true
```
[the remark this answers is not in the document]
AlgoStephenAkiki: Technically I could, but I think having all information available to the program might help us down the road when we want to display it to the user.  This runs only once during startup so the performance cost isn't that great and we get greater flexibility when we need it.

api/disabled_parameters.go
AlgoStephenAkiki: I checked the implementations, there seem to be nil checks in there so I don't think that is possible.

api/handlers.go
AlgoStephenAkiki: I tried going through some iterations of that but I couldn't really get to a solution I liked very much.  Do you have some suggestions?  The problem is that you always want to end up differentiating between a failed parameter and a failed endpoint so you always end up with some sort of switch statement.  

You could do something where it returns true or false, but then you have to handle the error generated by badRequest.

shiqizng: maybe something like, 
`if rc>0 {
return getErrorMsg(rc)
}
`
where getErrorMsg(rc) returns the different badRequest messages.

winder: You can do this middleware/plugin pattern used by go's HTTP (and echo). The generated echo code abstracts this a bit so I don't think it would be as clean. The middleware version would be similar but requires figuring out the operationID based on the context. Problematic because swagger has things like `/v2/assets/{assetid}` and echo uses `/v2/assets/:assetid:`.

The basic pattern looks something like this, where you chain the handlers together:
```
func verifyHandler(operationID, log logger.Log, next echo.Handler) echo.Handler {
    return func(e echo.Context) error {
        if rc := verify(operationID, e, log); rc != 0 {
            switch { ...returns here... }
        }
        return next(e)
    }
}
```

Something similar where you call badRequest in a helper then return the error code or run the rest of the handler would be functionally identical but less idiomatic

api/disabled_parameters_test.go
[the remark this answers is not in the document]
winder: The null logger "hook" has all log entries, so you can count the size of that array

api/handlers.go
AlgoStephenAkiki: I've implemented this although I had to use reflection because each of the functions had a different interface.  Make sure to "hide whitespace" when reviewing in github.

api/disabled_parameters_test.go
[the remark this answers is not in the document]
AlgoStephenAkiki: Done

api/handlers.go
```diff
@@ -147,502 +147,504 @@ func (si *ServerImplementation) MakeHealthCheck(ctx echo.Context) error {
...
+		Account:      accounts[0],
+	})
 }
 
 // SearchForAccounts returns accounts matching the provided parameters
 // (GET /v2/accounts)
 func (si *ServerImplementation) SearchForAccounts(ctx echo.Context, params generated.SearchForAccountsParams) error {
-	return si.verifyHandler("SearchForAccounts", ctx, func() error {
+	if err := si.verifyHandler("SearchForAccounts", ctx); err != nil {
+		return si.handleVerifyError(err, ctx)
```
winder, line 208: I don't think this is needed anymore, you can use `return badRequest(ctx, err.Error())`

AlgoStephenAkiki: Done

api/disabled_parameters_test.go
```diff
@@ -0,0 +1,161 @@
...
+		tstruct.setFormValues(&f)
+
+		ctx := ctxFactory(e, &f, tstruct)
+
+		logger, hook := test.NewNullLogger()
+
+		err := Verify(dm, "K1", *ctx, logger)
+
+		require.Equal(t, tstruct.expectedError, err)
+		require.Equal(t, tstruct.expectedErrorCount, len(hook.AllEntries()))
```
winder, line 88: Nit: `require.Len(t, hook.AllEntries(), tstruct.expectedErrorCount)`

winder (approved):

PR #884, opened 2022-02-16
RFC - Remove .version and tag hard dependencies

onetechnical:
## Summary

Not to merge yet - for consideration.

Having hard requirements around .version and tags for builds is a little inflexible, so I propose we relax these standards a little. Moving a step closer to not rebuilding binaries between releases, I've explored the idea of removing the .version file and relaxing tag hard restrictions here. You can set VERSION (for the binary build, defaults to `git describe --tags`) and RELVERSION (defaults to VERSION, though could be e.g. `TAG-rc1`).

We still need to separate the compilation from the repackaging. For example, if you build a binary with a tag of, say, 2.9.0-1-gabcd123, and package it as 2.9.0-rc1, then your tarball will need to be named 2.9.0-rc1, as that is how you would normally expect to unpack it. However, if you promote this, then the tarball would need to be re-packed and re-signed.

## Test Plan

Did a few test builds from the command line just to verify it worked.


Makefile
```diff
@@ -1,15 +1,16 @@
...
 OS_TYPE		?= $(shell $(SRCPATH)/mule/scripts/ostype.sh)
-ARCH			?= $(shell $(SRCPATH)/mule/scripts/archtype.sh)
-PKG_DIR		= $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(VERSION)
+ARCH		?= $(shell $(SRCPATH)/mule/scripts/archtype.sh)
+PKG_DIR		= $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(RELVERSION)
 
 # TODO: ensure any additions here are mirrored in misc/release.py
 GOLDFLAGS += -X github.com/algorand/indexer/version.Hash=$(shell git log -n 1 --pretty="%H")
-GOLDFLAGS += -X github.com/algorand/indexer/version.Dirty=$(if $(filter $(strip $(shell git status --porcelain|wc -c)), "0"),,true)
+GOLDFLAGS += -X github.com/algorand/indexer/version.Dirty=$(if $(filter $(strip $(shell git status --porcelain|wc -c)), 0),,true)
```
onetechnical, line 10: I think this is a bug. Fixed it here.

Makefile
```diff
@@ -1,15 +1,16 @@
...
+VERSION		?= $(shell git describe --tags)
+RELVERSION	?= $(VERSION)
 OS_TYPE		?= $(shell $(SRCPATH)/mule/scripts/ostype.sh)
-ARCH			?= $(shell $(SRCPATH)/mule/scripts/archtype.sh)
-PKG_DIR		= $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(VERSION)
+ARCH		?= $(shell $(SRCPATH)/mule/scripts/archtype.sh)
+PKG_DIR		= $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(RELVERSION)
 
 # TODO: ensure any additions here are mirrored in misc/release.py
 GOLDFLAGS += -X github.com/algorand/indexer/version.Hash=$(shell git log -n 1 --pretty="%H")
```
winder, line 9: Huh, looks like we have the hash from this already. So we can skip `shell git describe --tags`

Makefile
```diff
@@ -36,13 +37,7 @@ check: go-algorand
...
 	rm -rf $(PKG_DIR)
 	mkdir -p $(PKG_DIR)
-	misc/release.py --host-only --outdir $(PKG_DIR)
-
-# used in travis test builds; doesn't verify that tag and .version match
-fakepackage: go-algorand
-	rm -rf $(PKG_DIR)
-	mkdir -p $(PKG_DIR)
-	misc/release.py --host-only --outdir $(PKG_DIR) --fake-release
+	misc/release.py --host-only --outdir $(PKG_DIR) --version $(VERSION) --relversion $(RELVERSION)
```
winder, line 40: Why is the version passed in here? Filename?

misc/release.py
```diff
@@ -243,9 +230,9 @@ def main():
     if args.host_only:
         hostos, hostarch = hostOsArch()
         logger.info('will only run %s %s', hostos, hostarch)
-    with open('.version') as fin:
-        version = fin.read().strip()
-    ldflags = compile_version_opts(version, allow_mismatch=args.fake_release)
+    version = args.version
+    relversion = args.relversion
+    ldflags = compile_version_opts(version)
     for goos, goarch, debarch in osArchArch:
```
winder, line 236: This looks like cross-compiling stuff, we should make sure we don't set GOOS/GOARCH anymore

Makefile
```diff
@@ -1,15 +1,16 @@
 SRCPATH		:= $(shell pwd)
-VERSION		:= $(shell $(SRCPATH)/mule/scripts/compute_build_number.sh)
+VERSION		?= $(shell git describe --tags)
```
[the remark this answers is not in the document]
winder: My suggestion to use a script:
```
if [ branch starts with "release/" ] return string after "release/"
else return "git describe --tags
```

PR #893, opened 2022-02-23
Ensure query is canceled after account rewind.

winder:
## Summary

Make sure the transaction query is canceled when exiting the AccountAtRound function.

## Test Plan

Manual testing. Query psql for active queries. Verified a specific problematic query before and after this change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,7 @@ commands:
           command: go get -u golang.org/x/lint/golint
 
   run_tests:
+    no_output_timeout: 15m
     steps:
       - run: test -z `go fmt ./...`
       - run: make lint
@@ -110,6 +111,7 @@ commands:
       - run: make e2e
 
   run_tests_nightly:
+    no_output_timeout: 15m
     steps:
       - run: test -z `go fmt ./...`
       - run: make lint
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,25 +99,27 @@ commands:
           command: go get -u golang.org/x/lint/golint
 
   run_tests:
-    no_output_timeout: 15m
     steps:
       - run: test -z `go fmt ./...`
       - run: make lint
       - run: make check
       - run: make integration
-      - run: make test
+      - run:
+          command: make test
+          no_output_timeout: 15m
       - run: make test-generate
       - run: make fakepackage
       - run: make e2e
 
   run_tests_nightly:
-    no_output_timeout: 15m
     steps:
       - run: test -z `go fmt ./...`
       - run: make lint
       - run: make check
       - run: make integration
-      - run: make test
+      - run:
+          command: make test
+          no_output_timeout: 15m
       - run: make test-generate
       - run: make fakepackage
       - run: make e2e
--- a/accounting/rewind.go
+++ b/accounting/rewind.go
@@ -102,18 +102,15 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
 	ctx2, cf := context.WithCancel(ctx)
 	defer cf()
 	txns, r := db.Transactions(ctx2, tf)
+	defer func() {
+		cf()
+		for range txns {
+		}
+	}()
 	if r < account.Round {
 		err = ConsistencyError{fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round)}
 		return
 	}
-	return processTransactions(account, addr, round, txns)
-}
-
-func processTransactions(account models.Account, addr basics.Address, round uint64, txns <-chan idb.TxnRow) (acct models.Account, err error) {
-	defer func() {
-		for range txns {
-		}
-	}()
 	txcount := 0
 	for txnrow := range txns {
 		if txnrow.Error != nil {
--- a/accounting/rewind.go
+++ b/accounting/rewind.go
@@ -100,8 +100,10 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
 		MaxRound: account.Round,
 	}
 	ctx2, cf := context.WithCancel(ctx)
+	// In case of a panic before the next defer, call cf() here.
 	defer cf()
 	txns, r := db.Transactions(ctx2, tf)
+	// In case of an error, make sure the context is cancelled, and the channel is cleaned up.
 	defer func() {
 		cf()
 		for range txns {
--- a/accounting/rewind.go
+++ b/accounting/rewind.go
@@ -99,7 +99,13 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
 		MinRound: round + 1,
 		MaxRound: account.Round,
 	}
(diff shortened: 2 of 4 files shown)


accounting/rewind.go
```diff
@@ -99,7 +99,13 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
 		MinRound: round + 1,
 		MaxRound: account.Round,
 	}
+	ctx, cf := context.WithCancel(ctx)
 	txns, r := db.Transactions(ctx, tf)
+	defer func() {
+		cf()
+		for range txns {
+		}
+	}()
```
shiqizng, line 108: why not just defer cf()? what is line 106 for?

winder: I'm not 100% sure, but I think that `txns` is hung while trying to write the next result into it, so we need to discard the final entries in order for it to terminate.

accounting/rewind.go
```diff
@@ -99,7 +99,13 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
 		MinRound: round + 1,
 		MaxRound: account.Round,
 	}
+	ctx, cf := context.WithCancel(ctx)
```
tsachiherman, line 102: I think that the code is fine, but please use a different variable name for `ctx` to avoid variable shadowing.
Also, I'd suggest you'll defer the cf before the `db.Transactions` call. ( i.e. so that if the `db.Transactions` call get panicked, the context would still get destroyed )

AlgoStephenAkiki (approved):

accounting/rewind.go
tsachiherman: It's ok that you don't want to use it afterward.. the naming would just make sure the reader would understand it's a different object.

winder: using `defer cf()` is a little problematic because the logic would now depend on the evaluation order of defer. I think splitting the function in half should make the order clearer.

winder: Actually, splitting up the function forces them to be backwards. It occurs to me now that I can simply call cf multiple times.

tolikzinovyev (changes requested):
I think we need to make sure to drain the channel in `api/handler.go fetchAccounts()` as well. There, if rewind fails, we just exit the range loop.

accounting/rewind.go
```diff
@@ -99,17 +99,21 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
...
+	txns, r := db.Transactions(ctx2, tf)
 	if r < account.Round {
 		err = ConsistencyError{fmt.Sprintf("queried round r: %d < account.Round: %d", r, account.Round)}
 		return
 	}
+	return processTransactions(account, addr, round, txns)
+}
+
+func processTransactions(account models.Account, addr basics.Address, round uint64, txns <-chan idb.TxnRow) (acct models.Account, err error) {
+	defer func() {
```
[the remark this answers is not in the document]
winder: You're right, I got this backwards. Changing it back.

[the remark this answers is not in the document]
tolikzinovyev: Oh, never mind, this will wait until `txns` is closed. But still less efficient.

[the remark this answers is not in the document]
winder: Does cf() need to be called first in order to make sure we stop receiving new transactions? In testing, both seem to work properly, but I'm not sure how.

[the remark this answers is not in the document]
tolikzinovyev: Yes. Otherwise, you will read all transactions and then canceling the context will have no effect.

tsachiherman (approved):

accounting/rewind.go
```diff
@@ -99,7 +99,14 @@ func AccountAtRound(ctx context.Context, account models.Account, round uint64, d
 		MinRound: round + 1,
 		MaxRound: account.Round,
 	}
-	txns, r := db.Transactions(ctx, tf)
+	ctx2, cf := context.WithCancel(ctx)
+	defer cf()
```
tolikzinovyev, line 103: Isn't this a noop? The later defer will be called first.

winder: This was on Tsachi's suggestion. If db.Transactions were to panic it ensures we still cleanup.

tolikzinovyev: We don't have panics in indexer code, @tsachiherman .

tolikzinovyev: Even if we had, indexer would just crash it that case. So cleanup would be unnecessary.

winder: I believe the http server catches panics, so cleaning up resources would be useful.

tolikzinovyev:
@winder have you figured out why CI is failing?

accounting/rewind.go
tolikzinovyev: could you add a comment for the first `cf()` then?

winder:
> @winder have you figured out why CI is failing?

I think there are 2 issues.
1) We have an e2e test that uses algod's e2e subs. It was changed recently to include the python scripts, which had been missing previously.
2) There seems to be a new issue with the go-algorand build taking a long time. I haven't seen that before and am wondering if the circle CI resources are somehow overloaded and the build is taking a long time.

tolikzinovyev:
> > @winder have you figured out why CI is failing?
> 
> I think there are 2 issues.
> 
>     1. We have an e2e test that uses algod's e2e subs. It was changed recently to include the python scripts, which had been missing previously.
> 
>     2. There seems to be a new issue with the go-algorand build taking a long time. I haven't seen that before and am wondering if the circle CI resources are somehow overloaded and the build is taking a long time.

From the logs, `make test` is failing. So it's neither of those two issues.

accounting/rewind_test.go
```diff
@@ -67,10 +67,12 @@ func TestStaleTransactions1(t *testing.T) {
...
 	}
 
-	var outCh <-chan idb.TxnRow
+	ch := make(chan idb.TxnRow)
+	var outCh <-chan idb.TxnRow = ch
+	close(ch)
 
 	db := &mocks.IndexerDb{}
-	db.On("GetSpecialAccounts").Return(transactions.SpecialAddresses{}, nil)
+	db.On("GetSpecialAccounts", mock.Anything).Return(transactions.SpecialAddresses{}, nil)
```
winder, line 75: I'm not sure why this one was working. It should have started failing once we added the context to this function.

.circleci/config.yml
```diff
@@ -115,7 +117,9 @@ commands:
       - run: make lint
       - run: make check
       - run: make integration
-      - run: make test
+      - run:
+          command: make test
+          no_output_timeout: 15m
```
winder, line 122: `go test` times out after 10 minutes. Change the CircleCI timeout to something larger than that so that the timeout error is properly reported.

accounting/rewind.go
winder: done

tolikzinovyev (approved):

PR #899, opened 2022-02-24
fix failing e2e test

shiqizng:
## Summary
make e2e fails with nil pointer error with a new s3 artifact, 
```
output from 'go' 'run' 'cmd/e2equeries/main.go' '-pg' 'host=e2e-db port=5432 user=algorand password=algorand dbname=indexer_db sslmode=disable' '-q'
panic: runtime error: invalid memory address or nil pointer dereference
e2e_1     | [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf94ac9]
```
This PR fixes this issue. 

## Test Plan
run make e2e to validate test completes without error.

### DIFF
--- a/cmd/e2equeries/main.go
+++ b/cmd/e2equeries/main.go
@@ -44,7 +44,9 @@ func main() {
 	rowchan, _ := db.Transactions(context.Background(), rekeyTxnQuery)
 	for txnrow := range rowchan {
 		maybeFail(txnrow.Error, "err rekey txn %v\n", txnrow.Error)
-		rekeyedAuthAddrs = append(rekeyedAuthAddrs, txnrow.Txn.Txn.RekeyTo)
+		if txnrow.Txn != nil {
+			rekeyedAuthAddrs = append(rekeyedAuthAddrs, txnrow.Txn.Txn.RekeyTo)
+		}
 	}
 
 	// some rekeys get rekeyed back; some rekeyed accounts get deleted, just want to find at least one
@@ -63,7 +65,7 @@ func main() {
 	}
 	if !foundRekey {
 		// this will report the error in a handy way
-		printAccountQuery(db, idb.AccountQueryOptions{EqualToAuthAddr: rekeyedAuthAddrs[0][:], Limit: 1})
+		printAccountQuery(db, idb.AccountQueryOptions{EqualToAuthAddr: rekeyedAuthAddrs[0][:], Limit: 0})
 	}
 
 	// find an asset with > 1 account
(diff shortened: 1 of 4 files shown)


misc/e2elive.py
```diff
@@ -59,7 +59,7 @@ def main():
         s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
         tarname = 'net_done.tar.bz2'
         tarpath = os.path.join(tempdir, tarname)
-        firstFromS3Prefix(s3, bucket, 'indexer/e2e3', tarname, outpath=tarpath)
+        firstFromS3Prefix(s3, bucket, 'indexer/e2e4', tarname, outpath=tarpath)
```
winder, line 62: This is also specified in `misc/buildtestdata.sh`

cmd/e2equeries/main.go
```diff
@@ -44,7 +44,9 @@ func main() {
 	rowchan, _ := db.Transactions(context.Background(), rekeyTxnQuery)
 	for txnrow := range rowchan {
 		maybeFail(txnrow.Error, "err rekey txn %v\n", txnrow.Error)
-		rekeyedAuthAddrs = append(rekeyedAuthAddrs, txnrow.Txn.Txn.RekeyTo)
+		if txnrow.Txn != nil {
+			rekeyedAuthAddrs = append(rekeyedAuthAddrs, txnrow.Txn.Txn.RekeyTo)
+		}
```
winder, line 49: I'm not exactly sure whats happening here, but perhaps we need to check both
```suggestion
                t := txnrow.Txn
		if txnrow.RootTxn != nil {
		        t = txnrow.RootTxn
		}
		rekeyedAuthAddrs = append(rekeyedAuthAddrs, t.Txn.RekeyTo)
```

cmd/e2equeries/main.go
```diff
@@ -63,7 +65,7 @@ func main() {
 	}
 	if !foundRekey {
 		// this will report the error in a handy way
-		printAccountQuery(db, idb.AccountQueryOptions{EqualToAuthAddr: rekeyedAuthAddrs[0][:], Limit: 1})
+		printAccountQuery(db, idb.AccountQueryOptions{EqualToAuthAddr: rekeyedAuthAddrs[0][:], Limit: 0})
```
winder, line 68: What error is this supposed to be reporting?

winder: I think I see it. This triggers the "came up short" error?

shiqizng: yes

winder (approved):

PR #900, opened 2022-02-24
Updates for unlimited assets

cce:
## Summary

Companion updates for https://github.com/algorand/go-algorand/pull/3652

Adds new REST endpoints and "exclude" parameter for account information endpoint, similar to https://github.com/algorand/go-algorand/pull/3542. 

Update endpoint: GET /v2/accounts/{address}
- Supports new query parameter “exclude”, which can be a combination of `all`, `created-apps`, `created-assets`, `apps-local-state`, `assets`, `none`
- Removes empty “creator” string from AssetHolding type — was not implemented (removed associated TODO)

Update endpoint: GET /v2/assets/{asset-id}/balances 
- Remove "round" query parameter — I noticed support is not implemented, and the parameter was being ignored

New endpoints:
GET /v2/accounts/{address}/created-apps
GET /v2/accounts/{address}/created-assets
GET /v2/accounts/{address}/assets
GET /v2/accounts/{address}/apps-local-state
- Supported query parameters: asset-id / application-id (look up a single asset), include-all (include deleted), limit, next (for pagination)

## Test Plan

New tests added, existing tests should pass.


api/indexer.oas2.json
```diff
@@ -455,9 +651,6 @@
           {
             "$ref": "#/parameters/next"
           },
-          {
-            "$ref": "#/parameters/round"
```
algonautshant, line 459: Will this impact sdk backwards compatibility? Is this removed for both algod and indexer? Please comment.

api/indexer.oas2.json
```diff
@@ -1126,10 +1339,6 @@
           "description": "Asset ID of the holding.",
           "type": "integer"
         },
-        "creator": {
-          "description": "Address that created this asset. This is the address where the parameters for this asset can be found, and also the address where unwanted asset units can be sent in the worst case.",
-          "type": "string"
-        },
```
algonautshant, line 1132: Will this impact sdk backwards compatibility? Please comment.

api/indexer.oas2.json
cce: Yes, anyone who was providing this argument with an SDK will have a compilation error, and anyone who was using without the SDK will receive a 400 (since the backend will not know what "round" is for). After talking with @winder and @algoanne I believe this is preferable to the current behavior — where the round parameter is being silently ignored, the latest round's data is returned, but the caller thinks they are getting a historical value from an old round.

api/indexer.oas2.json
cce: Yes, similarly, if anyone was expecting this value using the SDK it will no longer compile. However, this value was always being returned as an empty string and was never implemented. So anyone who has ever written code that retrieved this field should have already found it empty and useless, and likely was not depending on it.

mxmauro:
I would like to suggest a change in the new implementation of `GET /v2/accounts/{address}`: Not to return an error if the number of apps/assets is high and instead add a message/flag to the response indicating that there are more items to instruct the user to call the new endpoints. May be an extra flag in the query.

Reason:

Assuming most of users won't have a lot of opted-in assets/apps, any UI like wallets would be able to fill the whole information with just one call to the backend.

If an error is returned, the UI must call the endpoint with exclude all and then the new apis (one for each type of information they want to retrieve) in order to get the first page of assets/apps they use.

cce:
We were concerned that for backwards compatibility reasons, giving callers a "trimmed down" account for some cases, but a full one in others, would leave them with the false impression that the account was empty, especially if they were using an older SDK.

Since most users won't have a lot of opted-in assets, like you said the error case is rare, but clear. Optimizing the number of calls for the rare error case wasn't a priority when we talked this through.

One hope we had, and considered making a requirement, was that developers would adopt exclude=all everywhere, and then only pull information about assets/apps when truly necessary. This would be the most error-proof / future-proof way to look up account information and saves the indexer a lot of unnecessary work.

For those that choose not to adopt exclude=all for their initial lookup (e.g. to check counts of opted-in assets/apps) and are willing to "risk" an optimistic exclude=none, do you think it is too onerous to have them implement the 400, then re-request flow?

cce:
Oh I re-read. So an extra flag that says "give me a trimmed version if it goes over the limit." I guess that is "exclude=sometimes" `:P` or some better term.. that would address the backwards compatibility concern.

Makefile
```diff
@@ -45,7 +47,7 @@ fakepackage: go-algorand
 	misc/release.py --host-only --outdir $(PKG_DIR) --fake-release
 
 test: idb/mocks/IndexerDb.go cmd/algorand-indexer/algorand-indexer
-	go test ./... -coverprofile=coverage.txt -covermode=atomic
+	go test -coverpkg=$(COVERPKG) ./... -coverprofile=coverage.txt -covermode=atomic
```
tsachiherman, line 50: It doesn't look like this relates to this PR, but I can't find a good reason not to have it.

winder: I'm not sure where my comment went, but I asked this question as well. It relates to reporting code coverage for new tests, where the test was written in another package.

api/handlers.go
```diff
@@ -698,7 +852,31 @@ func (si *ServerImplementation) fetchApplications(ctx context.Context, params ge
...
+func (si *ServerImplementation) fetchAppLocalStates(ctx context.Context, params idb.ApplicationQuery) ([]generated.ApplicationLocalState, uint64, error) {
+	var als []generated.ApplicationLocalState
+	var round uint64
+	err := callWithTimeout(ctx, si.log, si.timeout, func(ctx context.Context) error {
+		var results <-chan idb.AppLocalStateRow
+		results, round = si.db.AppLocalState(ctx, params)
+
+		for result := range results {
+			if result.Error != nil {
+				return result.Error
```
tsachiherman, line 868: I don't know if that would be an issue, but you might want to drain the channel before returning.

winder (approved):
Aside from the merge conflict, this still looks good!

Makefile
cce: I was trying to ensure I had good coverage of my changes to the postgres.go query builder exercised by handlers_e2e_test.go in a different package and was frustrated they weren't showing up, so I made that change. With a small repo like indexer it didn't cost anything to add the additional coverpkg arguments.

api/handlers.go
winder: This should be OK because the channel should be closed after writing an error.

tsachiherman (approved):

PR #902, opened 2022-02-28
Allow Viewing of Disabled Params

AlgoStephenAkiki:
Resolves #3583

Add a new subcommand "algorand-indexer api-config" to show current
config

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,50 @@ When `--token your-token` is provided, an authentication header is required. For
 ~$ curl localhost:8980/transactions -H "X-Indexer-API-Token: your-token"
 ```
 
+## Disabling Parameters 
+
+The Indexer has the ability to selectively enable or disable parameters for endpoints.  Disabling a "required" parameter will result in the entire endpoint being disabled while disabling an "optional" parameter will cause an error to be returned only if the parameter is provided.
+
+### Viewing the Current Configuration
+
+The Indexer has a default set of disabled parameters.  To view the disabled parameters issue:
+```
+~$ algorand-indexer api-config
+```
+
+This will output ONLY the disabled parameters in a YAML configuration.  To view all parameters (both enabled and disabled) issue:
+
+```
+~$ algorand-indexer api-config all
+```
+
+### Interpreting The Configuration
+
+Below is a sample output of what the output of `algorand-indexer api-config` will look like:
+
+```
+/v2/accounts:
+    optional:
+        - currency-greater-than: disabled
+        - currency-less-than: disabled
+/v2/assets/{asset-id}/transactions:
+    optional:
+        - note-prefix: disabled
+        - tx-type: disabled
+        - sig-type: disabled
+        - before-time: disabled
+        - after-time: disabled
+        - currency-greater-than: disabled
+        - currency-less-than: disabled
+        - address-role: disabled
+        - exclude-close-to: disabled
+        - rekey-to: disabled
+    required:
+        - asset-id: disabled
+```
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -112,6 +112,50 @@ When `--token your-token` is provided, an authentication header is required. For
...
+
+This will output ONLY the disabled parameters in a YAML configuration.  To view all parameters (both enabled and disabled) issue:
+
+```
+~$ algorand-indexer api-config all
+```
+
+### Interpreting The Configuration
+
+Below is a sample output of what the output of `algorand-indexer api-config` will look like:
```
winder, line 134:
```suggestion
Below is a snippet of the output from `algorand-indexer api-config`:
```

api/disabled_parameters.go
```diff
@@ -7,8 +7,103 @@ import (
...
+	// representation:
+	//  /v2/accounts/{account-id}:
+	//    required:
+	//      - account-id : enabled
+	//  /v2/accounts:
+	//    optional:
+	//      - auth-addr : enabled
+	//      - next: disabled
+	Data map[string]map[string][]map[string]string
+}
```
winder, line 27: But The type here seems very similar to what we already have for DisableMap. So there might be a way to avoid this similar but different type. This would become pretty useful when it comes time to parse the config file because you would use `yaml.Unmarshal` to do most of the work.

I think it's OK to leave it like this now, but maybe consider it again when you start parsing the file.

api/server.go
```diff
@@ -50,7 +50,7 @@ func (e ExtraOptions) handlerTimeout() time.Duration {
 }
 
 // Serve starts an http server for the indexer API. This call blocks.
-func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, fetcherError error, log *log.Logger, options ExtraOptions) {
+func Serve(ctx context.Context, disabledMapConfig *DisabledMapConfig, serveAddr string, db idb.IndexerDb, fetcherError error, log *log.Logger, options ExtraOptions) {
```
winder, line 53: Can this go in `ExtraOptions`?

cmd/algorand-indexer/api_config.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		var displayDisabledMapConfig *api.DisplayDisabledMap
+		// Show a limited subset
+		if len(args) == 0 {
+			displayDisabledMapConfig = api.MakeDisplayDisabledMapFromConfig(swag, disabledMapConfig, true)
+
+		} else {
+			// This is the only acceptable option
+			if args[0] != "all" {
```
winder, line 41: We should adjust the options so that they are included in the documentation. Perhaps instead of "all" it is a bool `--all` flag. Is this going to be a problem with the other command that are planned?

cmd/algorand-indexer/main.go
```diff
@@ -23,6 +24,8 @@ import (
 	"github.com/algorand/indexer/version"
 )
 
+var disabledMapConfig *api.DisabledMapConfig
```
winder, line 27: Move to ExtraOptions

winder (changes requested):
Some small changes needed, and an optional suggestion to consider.

cmd/algorand-indexer/api_config.go
```diff
@@ -58,3 +57,8 @@ var apiConfigCmd = &cobra.Command{
 
 	},
 }
+
+func init() {
+	apiConfigCmd.Flags().BoolVarP(&showAllDisabled, "all", "", false, "show all api config parameters, enabled and disabled")
+	apiConfigCmd.Flags().Lookup("all").NoOptDefVal = "true"
```
winder, line 63: I don't think you need this, `false` is already set as the default.

winder (approved):

PR #908, opened 2022-03-04
Fix/metrics

shiqizng:
## Summary
Metrics changes from monitoring dashboard updates broke the metrics calculation in block generator. This PR fixes this error, `error writing metrics (imported_tx_per_block): metric incomplete or not found: imported_tx_per_block`

## Test Plan
Execute block-generator runner to validate the error is gone and check that metrics report is complete.


winder:
I added a small change to avoid hard coding the metric name in the processing code. The TPS calculation is also incorrect, I didn't notice earlier but switching from counter to a gauge means that we lose data if scraping doesn't happen for every block.

Here is something that seems to work, but it must break the monitoring dashboard.

shiqizng:
closing this issue. it's not needed anymore after #913 is merged.

PR #912, opened 2022-03-07
Enable Validation and Acceptance of New Parameter Configuration

AlgoStephenAkiki:
Resolves #3584

- Allows for the supplying of configuration files for the daemon and
api-config subcommands.
- Properly validates the schema and contents of the supplied config file
- Enables default parameters to be disabled, allows integration tests to
  run with all parameters enabled

### DIFF
--- a/api/disabled_parameters_test.go
+++ b/api/disabled_parameters_test.go
@@ -17,7 +17,7 @@ import (
 )
 
 func TestToDisabledMapConfigFromFile(t *testing.T) {
-	expectedValue := &DisabledMapConfig{Data: map[string]map[string][]string{
+	expectedValue := DisabledMapConfig{Data: map[string]map[string][]string{
 		"/sampleEndpoint": {http.MethodGet: {"p2"}},
 	}}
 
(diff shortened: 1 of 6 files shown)


api/disabled_parameters_test.go
```diff
@@ -14,6 +16,165 @@ import (
...
+	}}
+
+	configFile := filepath.Join("test_resources", "mock_disabled_map_config.yaml")
+
+	// Nil pointer for openapi3.swagger because we don't want any validation
+	// to be run on the config (they are made up endpoints)
+	loadedConfig, err := MakeDisabledMapConfigFromFile(nil, configFile)
+	require.NoError(t, err)
+
+	require.True(t, reflect.DeepEqual(*expectedValue, *loadedConfig))
```
winder, line 31: Did you find that `DeepEqual` was required here? I believe testify uses that internally so it wouldn't be required. Also, you can avoid the pointer in expectedValue above so that you don't need to dereference it here. 
```suggestion
	require.NotNil(t, loadedConfig)
	require.Equal(t, expectedValue, *loadedConfig)
```

api/disabled_parameters_test.go
```diff
@@ -14,6 +16,165 @@ import (
...
+		})
+	}
+
+}
+
+func TestSchemaCheck(t *testing.T) {
+	type testingStruct struct {
+		name        string
+		ddm         *DisplayDisabledMap
+		expectError bool
```
winder, line 80: If you replace bool with a string, you can verify the error message also. Testing against a substring is sometimes convenient for making things more flexible, like `require.Contains(t, "p1, p3", err.Error())`. There are also some fancy things you can do with `require.ErrorIs` and `require.ErrorAs` to find/compare error types, but I don't think they would add much to this test.

winder (changes requested):
Looks good, I just had a couple small suggestions.

winder (approved):
Thanks for the updates, looks good!

PR #916, opened 2022-03-09
use ErrorResponse instead of AccountsErrorResponse

cce:
## Summary

As discussed with @winder it would make it easier for SDKs not to have to distinguish between different error response types — this merges the AccountsErrorResponse fields into the "data" object in ErrorResponse.

## Test Plan

Update tests that were looking for the old fields in AccountsErrorResponse.


winder (approved):
LGTM

tsachiherman (approved):

PR #917, opened 2022-03-09
Feature flag for query param configuration.

winder:
## Summary

Disable this feature for the next release.


api/disabled_parameters_test.go
```diff
@@ -77,7 +77,7 @@ func TestSchemaCheck(t *testing.T) {
 	type testingStruct struct {
 		name        string
 		ddm         *DisplayDisabledMap
-		expectError string
+		expectError []string
```
winder, line 80: This was flaky because the error string is generated according to maps being returned in a randomized order. So sometimes the tests with 2 values were swapped in the error string.

AlgoStephenAkiki (approved):

PR #922, opened 2022-03-10
Fix missing algorand-indexer_VERSION_armhf.deb.sig

algobarb:
## Summary

Fix missing algorand-indexer_VERSION_armhf.deb.sig because the signer skips over the armhf file to sign.

## Test Plan

Will need to test it on the next indexer release.
Makes sure the armhf file gets signed and uploaded to S3 in the armhf pipeline.

### DIFF
--- a/mule/sign.sh
+++ b/mule/sign.sh
@@ -21,7 +21,7 @@ make_hashes () {
     for ext in ${EXTENSIONS[*]}
     do
         if ls *"$VERSION"*."$ext"; then
-            HASHFILE="hashes_${OS_TYPE}_${ARCH}_${VERSION}_${ext}"
+            HASHFILE="hashes_${OS_TYPE}_${ARCH}_${VERSION}_${ext/./}"
             {
                 md5sum *"$VERSION"*."$ext" ;
                 shasum -a 256 *"$VERSION"*."$ext" ;


mule/sign.sh
```diff
@@ -21,7 +21,7 @@ make_hashes () {
     for ext in ${EXTENSIONS[*]}
     do
         if ls *"$VERSION"*."$ext"; then
-            HASHFILE="hashes_${OS_TYPE}_${ARCH}_${VERSION}_${ext}"
+            HASHFILE="hashes_${OS_TYPE}_${ARCH}_${VERSION}_${ext/./}"
```
algobarb, line 24: @onetechnical I followed the description on my ticket, but is this correct?

winder: It looks like this removes the period from `tar.bz2`? Seems fine to me if that's the intent.

onetechnical, line 24: Stylistically, I think replacing it with underscore makes it match the deb hashfiles:

```suggestion
            HASHFILE="hashes_${OS_TYPE}_${ARCH}_${VERSION}_${ext/./_}"
```

onetechnical (changes requested):

mule/sign.sh
algobarb: I made a change so it replaces all `.` with `_` in the extensions. `HASHFILE="hashes_${OS_TYPE}_${ARCH}_${VERSION}_${ext//./_}"`

onetechnical (approved):

PR #930, opened 2022-03-15
Shiqi/partitioning

shiqizng:
## Summary
This PR adds table partitioning for txn table. 

## Test Plan
Add migration unit test. Run performance tests using pgbench and block-generator.


urtho:
Not sure this is the right place to ask.

1) Why keep a separate txn_by_tixid index when primary key now contains txnid as the first subkey ?
2) If the primary key for txn does not contain round as the first subkey then queries for a block or round range will no longer use indexes. Also scanning for rounds in reverse order would be impossible. 
3) What is the rationale for changing the primary key on txn table ? IMHO this will slow a lot of things down unless there is a new dedicated index for round . 
4) `INSERT INTO txn_part SELECT * FROM txn ORDER BY round, intra` would not slow down the import but allow for a nicer physical layout - with recent rounds grouped together .

Am running partitioned TXN by round only (not txnid, not intra), without changing the primary key. This allows for faster single round read while still boosting scans for round ranges.

shiqizng:
> Not sure this is the right place to ask.
> 
> 1. Why keep a separate txn_by_tixid index when primary key now contains txnid as the first subkey ?
> 2. If the primary key for txn does not contain round as the first subkey then queries for a block or round range will no longer use indexes. Also scanning for rounds in reverse order would be impossible.
> 3. What is the rationale for changing the primary key on txn table ? IMHO this will slow a lot of things down unless there is a new dedicated index for round .
> 4. `INSERT INTO txn_part SELECT * FROM txn ORDER BY round, intra` would not slow down the import but allow for a nicer physical layout - with recent rounds grouped together .
> 
> Am running partitioned TXN by round only (not txnid, not intra), without changing the primary key. This allows for faster single round read while still boosting scans for round ranges.

Hi, primary keys are not changed, see file changes.

winder:
> Not sure this is the right place to ask.
> 
>     1. Why keep a separate txn_by_tixid index when primary key now contains txnid as the first subkey ?
> 
>     2. If the primary key for txn does not contain round as the first subkey then queries for a block or round range will no longer use indexes. Also scanning for rounds in reverse order would be impossible.
> 
>     3. What is the rationale for changing the primary key on txn table ? IMHO this will slow a lot of things down unless there is a new dedicated index for round .
> 
>     4. `INSERT INTO txn_part SELECT * FROM txn ORDER BY round, intra` would not slow down the import but allow for a nicer physical layout - with recent rounds grouped together .
> 
> 
> Am running partitioned TXN by round only (not txnid, not intra), without changing the primary key. This allows for faster single round read while still boosting scans for round ranges.

It sounds like there is some confusion about the changes, could you be more specific about why you think the primary key has changed?

I'm not sure what you mean by the txnid is contained as the first subkey. Our intention was to use the same `(round, intra)` primary key, but partitioned. This is what we've done.

The primary key must to contain `round` and `intra` (the offset into the round), otherwise it wouldn't be unique.

Regarding (4), I believe the order is implicit. But if it's left out, wouldn't the DB be allowed to optimize according to the physical layout on its own?

urtho:
> It sounds like there is some confusion about the changes, could you be more specific about why you think the primary key has changed?
> 

sorry. have not noticed https://github.com/algorand/indexer/commits?author=shiqizng which removes txid from primary key. 

> I'm not sure what you mean by the txnid is contained as the first subkey. Our intention was to use the same `(round, intra)` primary key, but partitioned. This is what we've done.
> 
> The primary key must to contain `round` and `intra` (the offset into the round), otherwise it wouldn't be unique.
> 
> Regarding (4), I believe the order is implicit. But if it's left out, wouldn't the DB be allowed to optimize according to the physical layout on its own?

unless there is explicit order by in the select statement the insert will get rounds in randomly

PR #933, opened 2022-03-16
Set max account API resource limit to 1000.

winder:
## Summary

Change default for maxAPIResourcesPerAccount, fix a bug.

### DIFF
--- a/api/handlers_e2e_test.go
+++ b/api/handlers_e2e_test.go
@@ -534,7 +534,7 @@ func TestAccountMaxResultsLimit(t *testing.T) {
 		includeDeleted bool
 		errStatus      int
 	}{
-		{address: test.AccountA, exclude: []string{}},
+		{address: test.AccountA, exclude: []string{}, errStatus: http.StatusBadRequest},
 		{address: test.AccountA, exclude: []string{"all"}},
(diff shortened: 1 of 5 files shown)


cce (approved):

api/handlers_e2e_test.go
```diff
@@ -522,6 +534,7 @@ func TestAccountMaxResultsLimit(t *testing.T) {
 		includeDeleted bool
 		errStatus      int
 	}{
+		{address: test.AccountA, exclude: []string{}},
```
cce, line 537: This test is using maxResults = 14 so you are going to hit a 400 here. (because AccountA has 20 resources)
```suggestion
		{address: test.AccountA, exclude: []string{}, errStatus: http.StatusBadRequest},
```

api/handlers_e2e_test.go
```diff
@@ -576,6 +589,7 @@ func TestAccountMaxResultsLimit(t *testing.T) {
 		errStatus  int
 		errAddress basics.Address
 	}{
+		{exclude: []string{}},
```
cce, line 592: Similarly, the maxResults parameter is set to 14 here too.
```suggestion
		{exclude: []string{}, errStatus: http.StatusBadRequest},
```

PR #935, opened 2022-03-16
Consider TxType when selecting transaction amount currency type.

winder:
## Summary

## Test Plan

New unit test.


api/converter_utils.go
```diff
@@ -624,6 +615,15 @@ func (si *ServerImplementation) transactionParamsToTransactionFilter(params gene
...
 
+	// filter Algos or Asset but not both.
+	if filter.AssetID != 0 || filter.TypeEnum == idb.TypeEnumAssetTransfer {
+		filter.AssetAmountLT = params.CurrencyLessThan
+		filter.AssetAmountGT = params.CurrencyGreaterThan
+	} else {
+		filter.AlgosLT = params.CurrencyLessThan
+		filter.AlgosGT = params.CurrencyGreaterThan
+	}
+
```
shiqizng, line 626: this block is just being moved and it doesn't change the method's behavior, right?

winder: I changed the first condition, but needed to move it down so that `filter.TypeEnum` is initialized.

AlgoStephenAkiki (approved):

shiqizng (approved):

PR #947, opened 2022-03-29
add linting

shiqizng:
## Summary

Add linting to project.

### DIFF
--- a/.golangci-warnings.yml
+++ b/.golangci-warnings.yml
@@ -1,0 +1,79 @@
+run:
+  timeout: 5m
+  tests: true
+
+linters:
+  disable-all: true
+  enable:
+    - staticcheck
+    - structcheck
+    - typecheck
+    - varcheck
+    - deadcode
+    - gosimple
+    - unused
+    - partitiontest
+
+
+linters-settings:
+  custom:
+    partitiontest:
+      path: cmd/partitiontest_linter/plugin.so
+      description: This custom linter checks files that end in '_test.go', specifically functions that start with 'Test' and have testing argument, for a line 'partitiontest.ParitionTest(<testing arg>)'
+      original-url: github.com/algorand/go-algorand/cmd/partitiontest_linter
+
+severity:
+  default-severity: warning
+
+issues:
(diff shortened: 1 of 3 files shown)


.golangci-warnings.yml
```diff
@@ -0,0 +1,56 @@
...
+    - deadcode
+    - gosimple
+    - unused
+
+severity:
+  default-severity: warning
+
+issues:
+  # use these new lint checks on code since #2574
+  new-from-rev:  99ab66fdadcd77fef4048bb2aa86cae1f06c6e08
```
cce, line 21: You'll want to remove this if you have already fixed all the linting issues in this codebase — this references a commit on the go-algorand repo that I used to ignore all lint issues before this commit. (Because it was too much work to fix all the issues in the code before that date)

cce: oh I see 99ab66fdadcd77fef4048bb2aa86cae1f06c6e08 is a commit on indexer, well, never mind

.golangci.yml
```diff
@@ -0,0 +1,43 @@
...
+    - govet
+    - ineffassign
+    - misspell
+
+severity:
+  default-severity: error
+
+issues:
+  # use these new lint checks on code since #2574
+  new-from-rev: 988ad6384783701d0b78bed903ff1bc28277f05c
```
cce, line 20: Is it a lot of work to make indexer clean for the error-level linters you've picked out? just curious.

winder:
@cce should we pull in this change also? https://github.com/algorand/go-algorand/pull/3734

Looks like there were some side effects on go-algorand that might not apply to Indexer.

.golangci-warnings.yml
shiqizng: I should update the description.

.golangci.yml
shiqizng: if we start from this commit, there are just a couple of errors.

cce:
sure, the gci importer is great!

winder (approved):
LGTM, lets try it out

PR #948, opened 2022-03-29
Use external docker container to speed up tests.

winder:
## Summary

Setup a docker container to reuse across tests instead of depending on gnomock to start and stop the container for every test.

## Test Plan

Existing tests.

### DIFF
--- a/idb/postgres/internal/testing/testing.go
+++ b/idb/postgres/internal/testing/testing.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"os"
 	"testing"
 
 	"github.com/jackc/pgx/v4/pgxpool"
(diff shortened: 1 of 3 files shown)


AlgoStephenAkiki (approved):

idb/postgres/internal/testing/testing.go
```diff
@@ -14,15 +15,23 @@ import (
...
+var testpg string
+
+func init() {
+	flag.StringVar(&testpg, "test-pg", "", "postgres connection string; resets the database")
+	if testpg == "" {
+		// Note: tests across packages run in parallel, so this does not work
+		// very well unless you use "-p 1".
+		testpg = os.Getenv("TEST_PG")
+	}
+}
```
shiqizng, line 27: does the process exit with error if testpg is empty, which happens if neither test-pg or TEST_PG is set?

winder: If TEST_PG isn't set it returns an empty string, so it would be a no-op.

shiqizng (approved):

algojack (approved):

PR #957, opened 2022-04-15
vacuum doc

shiqizng:
## Summary
This PR adds a documentation on using Postgres vacuum.

### DIFF
--- a/docs/Vacuum.md
+++ b/docs/Vacuum.md
@@ -1,0 +1,33 @@
+## PostgresDB Vacuum
+
+See [official documentation](https://www.postgresql.org/docs/current/routine-vacuuming.html#VACUUM-BASICS) for more details
+
+Each update or delete query creates dead tuples. And running vacuum recovers disk space occupied by these tuples.
+However, it creates a substantial amount of I/O traffic, which can cause poor performance for other active sessions. 
+
+There are two types of vacuum: **autovacuum** or **vacuum** full.
+In the default configuration, autovacuuming is enabled and the related configuration parameters are appropriately set.
+
+
+### Autovacuum (Standard vacuum)
+
+- can run in parallel with production database operations
+- does not recover as much space full vacuum
+
+### Vacuum full
+- cannot be done in parallel with other use of the table;
+- runs much more slowly
+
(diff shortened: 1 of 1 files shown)


docs/Vacuum.md
```diff
@@ -0,0 +1,33 @@
...
+
+Vacuum can be [configured](https://www.postgresql.org/docs/current/runtime-config-autovacuum.html) to be more or less aggressive
+
+The vacuum threshold is defined as:
+
+vacuum threshold = vacuum base threshold + vacuum scale factor * number of tuples
+
+Insert threshold is defined as:
+
+vacuum insert threshold = vacuum base insert threshold + vacuum insert scale factor * number of tuples
```
winder, line 33: Changed the number of pound symbols
PostgresDB -> PostgreSQL
Fixed a typo
Added one more bullet to vacuum full
```suggestion
# PostgrSQL Vacuum

See [official documentation](https://www.postgresql.org/docs/current/routine-vacuuming.html#VACUUM-BASICS) for more details

Each update or delete query creates dead tuples. And running vacuum recovers disk space occupied by these tuples.
However, it creates a substantial amount of I/O traffic, which can cause poor performance for other active sessions. 

There are two types of vacuum: **autovacuum** or **vacuum** full.
In the default configuration, autovacuuming is enabled and the related configuration parameters are appropriately set.


## Autovacuum (Standard vacuum)

- can run in parallel with production database operations
- does not recover as much space full vacuum

## Vacuum full
- cannot be done in parallel with other use of the table
- runs much more slowly
- is sometimes able to recover space that autovacuum cannot

*Since vacuum full is much more costly,  the usual goal of routine vacuuming is to standard vacuum often enough to avoid full vacuum.*

## Vacuum configuration

Vacuum can be [configured](https://www.postgresql.org/docs/current/runtime-config-autovacuum.html) to be more or less aggressive

The vacuum threshold is defined as:

vacuum threshold = vacuum base threshold + vacuum scale factor * number of tuples

Insert threshold is defined as:

vacuum insert threshold = vacuum base insert threshold + vacuum insert scale factor * number of tuples
```

winder (changes requested):
Looks good, just had a couple small suggestions

winder (approved):

PR #958, opened 2022-04-15
Cleanup README, add index management section, clarify transaction result order. 

winder:
## Summary

Add a section to the documentation with some useful indices. In the future we can add to it, and when "slow queries" are reported we can use it as a reference.

The docker-compose demo was broken, so I removed it along with updating the README.

## Test Plan

N/A


README.md
```diff
@@ -286,11 +255,22 @@ sudo systemctl start algorand-indexer
...
-Load balancing: 
+## Load balancing
 If indexer is deployed with a clustered database using multiple readers behind a load balancer, query discrepancies are possible due to database replication lag. Users should check the `current-round` response field and be prepared to retry queries when stale data is detected.
 
+## Custom indices
+Different application workloads will require different custom indices in order to make queries perform well. More information is available in [PostgresqlIndexes.md](docs/PostgresqlIndexes.md).
+
+## Transaction results order
+
+The order transactions are returned it depends on whether or not an account address filter is used.
```
Eric-Warehime, line 268:
```suggestion
The order transactions are returned in depends on whether or not an account address filter is used.
```

Eric-Warehime (approved):

api/indexer.oas2.json
```diff
@@ -679,7 +679,7 @@
     },
     "/v2/assets/{asset-id}/transactions": {
       "get": {
-        "description": "Lookup transactions for an asset.",
+        "description": "Lookup transactions for an asset. Transactions are returned oldest to newest.",
```
shiqizng, line 682: is there a reason the order is inconsistent with /v2/accounts/{account-id}/transactions?

shiqizng (approved):

api/indexer.oas2.json
winder: Yes, basically, the txn/account index is optimized the "wallet" scenario and returns things newest to oldest. All other queries are done oldest to newest. The README goes into a little more detail about it.

PR #965, opened 2022-04-20
Modified Rollback to occur before ctx

### LINKED ISSUE #952: Indexer APIs with callWithTimeout have ctx race that leads to forced postgres connection close
Noticed that my indexers are creating and destroying postgres connections at the rate of almost 50% of the request rate.
Indexer behaves like there was no connection pooling.  Wireshark confirmed that properly acked ROLLBACK command still causes pgxpool to terminate the connection.

Added a this little diagnostic tweak:

```go
go func() {
        db.yieldAccountsThread(req)
        close(req.out)
        if rerr := tx.Rollback(ctx); rerr != nil {
            db.log.Printf("rollback error: %s", rerr)
        }
    }()
```

And my logs are now full of:

`{"level":"info","msg":"rollback error: timeout: context canceled","time":"2022-04-06T22:21:15Z"}`

This means that the CTX got cancelled before the rollback was confirmed and according to the pgx source this forces connection close. 

This seems harmless but also totally negates the gains from having a connection pool. 

And the reason seems to be a race between 
`close(req.out)` and `tx.Rollback(ctx)` 

`close(req.out)` causes  `defer cancel()` on the ctx in `callWithTimeout` to fire (nondeterministically) ahead of actual rollback .

Doing the tx.Rollback before the close(req.out) should solve the problem at a cost of a slight delay to the API response. Or rollback might be called with a new, unrealated, deadline ctx. 

This is on 2.10.0 , linux 64bit . Not able to reproduce in the lab! But happens to all productions servers.

AlgoStephenAkiki:
Resolves #952

Modified postgres.go to make it so that Rollback is called before
closing a returned channel, eliminating a race condition in
the callWithTimeout function

## Summary

## Test Plan

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -1672,11 +1672,11 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
 	round, err := db.getMaxRoundAccounted(ctx, tx)
 	if err != nil {
 		err = fmt.Errorf("account round err %v", err)
-		out <- idb.AccountRow{Error: err}
 		// Because we return a channel into a "callWithTimeout" function,
 		// We need to make sure that rollback is called before close()
 		// otherwise we can end up with a situation where "callWithTimeout"
 		// will cancel our context, resulting in connection pool churn
+		out <- idb.AccountRow{Error: err}
 		if rerr := tx.Rollback(ctx); rerr != nil {
 			db.log.Printf("rollback error: %s", rerr)
 		}
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -852,10 +852,6 @@ func (db *IndexerDb) Transactions(ctx context.Context, tf idb.TransactionFilter)
 
 	round, err := db.getMaxRoundAccounted(ctx, tx)
 	if err != nil {
-		// Because we return a channel into a "callWithTimeout" function,
-		// We need to make sure that rollback is called before close()
-		// otherwise we can end up with a situation where "callWithTimeout"
-		// will cancel our context, resulting in connection pool churn
 		if rerr := tx.Rollback(ctx); rerr != nil {
 			db.log.Printf("rollback error: %s", rerr)
 		}
@@ -1672,10 +1668,6 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
 	round, err := db.getMaxRoundAccounted(ctx, tx)
 	if err != nil {
 		err = fmt.Errorf("account round err %v", err)
-		// Because we return a channel into a "callWithTimeout" function,
-		// We need to make sure that rollback is called before close()
-		// otherwise we can end up with a situation where "callWithTimeout"
-		// will cancel our context, resulting in connection pool churn
 		out <- idb.AccountRow{Error: err}
 		if rerr := tx.Rollback(ctx); rerr != nil {
 			db.log.Printf("rollback error: %s", rerr)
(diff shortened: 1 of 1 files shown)


idb/postgres/postgres.go
```diff
@@ -850,15 +852,27 @@ func (db *IndexerDb) Transactions(ctx context.Context, tf idb.TransactionFilter)
 
 	round, err := db.getMaxRoundAccounted(ctx, tx)
 	if err != nil {
-		tx.Rollback(ctx)
+		// Because we return a channel into a "callWithTimeout" function,
+		// We need to make sure that rollback is called before close()
+		// otherwise we can end up with a situation where "callWithTimeout"
+		// will cancel our context, resulting in connection pool churn
+		if rerr := tx.Rollback(ctx); rerr != nil {
+			db.log.Printf("rollback error: %s", rerr)
```
tolikzinovyev, line 860: should this use `%v` or `%w`?

idb/postgres/postgres.go
```diff
@@ -1659,8 +1673,14 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
 	if err != nil {
 		err = fmt.Errorf("account round err %v", err)
 		out <- idb.AccountRow{Error: err}
+		// Because we return a channel into a "callWithTimeout" function,
```
tolikzinovyev, line 1676: I would put this one line above in case the caller cancels the context right after receiving an error on the channel.

tolikzinovyev (commented):
Looks good, but I would really consider calling `Rollback(context.Background)` to simplify interaction between `postgres` and `util.go`.

idb/postgres/postgres.go
AlgoStephenAkiki: Since it isn't returning/wrapping the error, I think it is fine here to convert to string.

AlgoStephenAkiki: Is there some documentation or standard that recommends %v over %s in this scenario?

tolikzinovyev: Not sure, probably `%s` is fine.

idb/postgres/postgres.go
```diff
@@ -1658,9 +1672,15 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
 	round, err := db.getMaxRoundAccounted(ctx, tx)
 	if err != nil {
 		err = fmt.Errorf("account round err %v", err)
+		// Because we return a channel into a "callWithTimeout" function,
+		// We need to make sure that rollback is called before close()
+		// otherwise we can end up with a situation where "callWithTimeout"
+		// will cancel our context, resulting in connection pool churn
 		out <- idb.AccountRow{Error: err}
```
tolikzinovyev, line 1679: This should be below `Rollback()`. There are are many similar cases below.

AlgoStephenAkiki: Didn't you say earlier that it should be above AccountRow?

tolikzinovyev: First we should rollback and then write an error to the channel and close the channel.

idb/postgres/postgres.go
```diff
@@ -1659,8 +1669,10 @@ func (db *IndexerDb) GetAccounts(ctx context.Context, opts idb.AccountQueryOptio
 	if err != nil {
 		err = fmt.Errorf("account round err %v", err)
 		out <- idb.AccountRow{Error: err}
+		if rerr := tx.Rollback(ctx); rerr != nil {
```
tolikzinovyev, line 1672: Could you move it below `close(out)` or above `err = ...` so as to not break the flow? There are some similar cases below.

tolikzinovyev (approved):

winder (approved):

Issue #486, opened 2021-05-04
Omit expected fetcher error from health output.


winder:
## Summary
Some of the recent refactoring / improvements has negatively impacted the health output.

1) The improved errors output includes errors from fetcher. On startup there is always one expected error "ledger does not have entry _", and it is now displayed briefly before being cleared out after successfully continuing.

~2) We no longer update the round accounted value when there are no transactions, so the health round can become out of date. This is especially visible in a private network with sandbox where most rounds will be empty.~

Example of indexer health displayed when starting sandbox:
```
{
  "data": {
    "migration-required": false,
    "read-only-node": true
  },
  "db-available": true,
  "errors": [
    "fetcher error: HTTP 500 Internal Server Error: {\"message\":\"ledger does not have entry 1 (latest 0, committed 0)\"}\n"
  ],
  "is-migrating": false,
  "message": "0",
  "round": 0
}
```

## Scope/Requirements
1) Suppress the fetcher error.
2) Update the round accounted number even for empty rounds.

## Urgency/Relative Priority
This is pretty visible to users and not very hard to fix. We should probably do it.

barnjamin:
https://github.com/algorand/indexer/issues/483 maybe toss this one on there as well?

algojack:
This is also showing up in our cicd pipeline, and sandbox up commands. Would be nice to get this fixed so we don't waste time trying to troubleshoot it every time something else breaks.

PR #966, opened 2022-04-20
Do not show block not found error after fetcher has finished catchup.

Eric-Warehime:
## Summary

Addresses feedback in https://github.com/algorand/sandbox/pull/110 for making the version command global again.  
Also addresses #486 by not setting the error status during catchup when receiving algod client errors.
## Test Plan

Tested via repro using sandbox. Fresh sandbox up commands no longer show algod error in the health status command which were produced by the final catchup lookup.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -124,7 +124,6 @@ func (bot *fetcherImpl) catchupLoop(ctx context.Context) error {
 			if ctx.Err() != nil {
 				return fmt.Errorf("catchupLoop() fetch err: %w", err)
 			}
-			bot.setError(err)
 			bot.log.WithError(err).Errorf("catchup block %d", bot.nextRound)
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


fetcher/fetcher.go
```diff
@@ -124,7 +124,6 @@ func (bot *fetcherImpl) catchupLoop(ctx context.Context) error {
 			if ctx.Err() != nil {
 				return fmt.Errorf("catchupLoop() fetch err: %w", err)
 			}
-			bot.setError(err)
```
AlgoStephenAkiki, line 127: So we just want to permanently not set the error?  Just confirming if this is intended functionality.

Eric-Warehime: Yes, my intention is to log errors returned from Algod client but not set them.

The error returned when the catchup loop has caught up is https://github.com/algorand/go-algorand-sdk/blob/develop/client/v2/common/common.go#L79 just a generic typedef for error, so as far as I can tell there is not a great way to differentiate between an aglod error that is expected because we caught up versus one this stems from a problem.

AlgoStephenAkiki (approved):

winder (approved):

fetcher/fetcher.go
winder: If there was some other, more severe error, it would be triggered moments later in the regular block following routine.

PR #967, opened 2022-04-20
New block fetcher unit tests.

Eric-Warehime:
## Summary

Adds unit tests to fetcher. I wasn't able to get full coverage at the moment, but the framework is there for us to expand these in the future.

I mocked the algod client and returned errors in places in order to validate the behavior is as expected.

This also includes 2 commits from #966 since `TestFetcherImplCatchupLoopBlockError` would fail without that.

## Test Plan

Unit testing.


AlgoStephenAkiki (approved):

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+func TestFetcherImplAlgodReturnsClient(t *testing.T) {
+	aclient := mockAClient(t, &AlgodHandler{})
+	fetcher := &fetcherImpl{aclient: aclient, log: logrus.New()}
+	require.Equal(t, aclient, fetcher.Algod(), "Algod client returned from fetcherImpl does not match expected instance.")
+}
+
+func TestFetcherImplSetError(t *testing.T) {
+	aclient := mockAClient(t, &AlgodHandler{})
+	fetcher := &fetcherImpl{aclient: aclient, log: logrus.New()}
```
winder, line 60: logrus ships with a mock logger, I don't think it's needed here but maybe it would be in the future: https://pkg.go.dev/github.com/sirupsen/logrus/hooks/test#NewNullLogger

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,123 @@
...
+	expectedErr := fmt.Errorf("foobar")
+	fetcher.setError(expectedErr)
+	require.Equal(t, expectedErr.Error(), fetcher.Error(), "Error produced by setError was not reflected in Error output.")
+}
+
+func TestFetcherImplProcessQueueHandlerError(t *testing.T) {
+	mockAlgodHandler := &AlgodHandler{}
+	aclient := mockAClient(t, mockAlgodHandler)
+	ctx, cf := context.WithCancel(context.Background())
+	defer cf()
```
winder, line 70: I guess this is to ensure fetcher terminates after the test is finished?

If so, I don't see how it could work. If there is no error, wouldn't `fetcher.Run` go on forever?

Perhaps `context.WithTimeout` would be a better safeguard?

What happens if you don't use either, and just call `fetcher.Run(context.Background())`?

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,123 @@
...
+		req := args.Get(1).(*http.Request)
+		path := req.URL.Path
+		if strings.Contains(path, "v2/blocks/") {
+			if passingCalls <= 0 {
+				respWriter.WriteHeader(http.StatusInternalServerError)
+			} else {
+				block := new(bookkeeping.Block)
+				respWriter.WriteHeader(http.StatusOK)
+				respWriter.Write(protocol.Encode(block))
+				passingCalls--
```
winder, line 116: nit: I think it's pretty uncommon for `new` to be the right way to declare something. It's all personal preferences, but I would probably opt for one of these:
```suggestion
				respWriter.WriteHeader(http.StatusOK)
				respWriter.Write(protocol.Encode(&bookkeeping.Block{}))
				passingCalls--
```
```suggestion
				var block bookkeeping.Block
				respWriter.WriteHeader(http.StatusOK)
				respWriter.Write(protocol.Encode(&block))
				passingCalls--
```
```suggestion
				data := protocol.Encode(&bookkeeping.Block{})
				respWriter.WriteHeader(http.StatusOK)
				respWriter.Write(data)
				passingCalls--
```

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/rpcs"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
```
winder, line 16: nit: we group imports by builtin / 3rd party / 1st party. I think the CI would have warned about this if it was working properly on forks.

winder (changes requested):
These are great! Just need to fix the import order.

fetcher/fetcher_test.go
Eric-Warehime: Good to know! I agree, I think it's not needed for these tests, but we should look at it in the future.

fetcher/fetcher_test.go
Eric-Warehime: You're right, these are completely unnecessary. I removed these and just used `context.Background()` for the tests.

winder (approved):

PR #974, opened 2022-04-25
Skip validator errors resulting from the account resource limit.

winder:
## Summary

Do not generate errors if validator detects a mismatch between algod and indexer due to the account resource limit.

## Test Plan

Manual testing.

```
$ ./validator --algod-token _ --indexer-token _ --indexer-url "" --algod-url "" --addr IACII2KW5LSVKZKSJ5MMC77745CSVROOV47CEZQ7K75Q7TFKWXK5HNCK5Q 2>err.log

0        : _

Number of errors: [0 / 1]
Skipped (account-limit): 1
Retry count: 0
Checks per second: 1.701474
Test duration: 00:00:00
```
```
$ cat err.log 
===================================================================
2022-04-25 11:34:30 AM
Account: IACII2KW5LSVKZKSJ5MMC77745CSVROOV47CEZQ7K75Q7TFKWXK5HNCK5Q
Error #: 0
Retries: 0
Rounds Match: false
Address skipped: too many asset and/or accounts to return
```

### DIFF
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -647,7 +647,7 @@ func (si *ServerImplementation) maxAccountsErrorToAccountsErrorResponse(maxErr i
 		"total-created-apps":    maxErr.TotalAppParams,
 	}
 	return generated.ErrorResponse{
-		Message: "Result limit exceeded",
+		Message: ErrResultLimitReached,
 		Data:    &extraData,
 	}
 }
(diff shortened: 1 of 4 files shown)


tolikzinovyev:
Why not increase the limit when validating?

winder:
> Why not increase the limit when validating?

That would also be good, and this change wouldn't prevent doing that as well. Regardless of whether or not the server is suitably configured, this change means the report is not going to have false positives. Whether or not anything was skipped (and which accounts were skipped) is included in the report / error log.

tolikzinovyev (approved):

PR #975, opened 2022-04-25
Add --error-log-file validator option.

winder:
## Summary

New validator option to automate redirecting errors to a log file. The new `--error-log-file err.log` option is fundamentally identical to including `2> err.log` to the process call.

## Test Plan

Manual testing

### DIFF
--- a/cmd/validator/core/command.go
+++ b/cmd/validator/core/command.go
@@ -21,6 +21,7 @@ func init() {
 		threads      int
 		processorNum int
 		printCurl    bool
+		errorLogFile string
 	)
 
 	ValidatorCmd = &cobra.Command{
@@ -28,7 +29,7 @@ func init() {
 		Short: "validator",
 		Long:  "Compare algod and indexer to each other and report any discrepencies.",
 		Run: func(cmd *cobra.Command, _ []string) {
-			run(config, addr, threads, processorNum, printCurl)
+			run(config, errorLogFile, addr, threads, processorNum, printCurl)
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/validator/core/command.go
```diff
@@ -57,6 +59,19 @@ func run(config Params, addr string, threads int, processorNum int, printCurl bo
 		ErrorLog.Fatalf("indexer-url parameter is required.")
 	}
 
+	if errorLogFile != "" {
+		_, err := os.Stat(errorLogFile)
+		if !os.IsNotExist(err) {
+			ErrorLog.Fatalf("Error log already exists: %s", errorLogFile)
```
algojack, line 65: so on a restart it will fail or am i misunderstanding this?

algojack: also double negative in !os.IsNotExist

algojack:
why do we need this code if it's the same as 2>err.log?

cmd/validator/core/command.go
winder: Yes, I didn't think it makes any sense to restart. If you ran a long test you risk deleting results on accident. If you appending to the end then the second test results are buried deep inside a large file.

Open to suggestions if this behavior isn't desired.

For the double negative, the go file interfaces are kinda weird and the double negative is the only sure way I know of to verify that the correct condition is being checked.

https://stackoverflow.com/a/12518877/204023

winder:
> why do we need this code if it's the same as 2>err.log?

Convenience. I usually lookup the syntax for redirecting the error stream to double check that I got it right.

AlgoStephenAkiki (approved):

PR #943, opened 2022-03-22
build: update to go 1.16

tsachiherman:
## Summary

Update to use go 1.16, as well as update the dependent sub-module.

## Test Plan

Use existing tests.

PR #980, opened 2022-04-28
build: Update for Mac M1 support

Eric-Warehime:
## Summary

Adds M1 support
* CPATH/LIBRARY_PATH updated for new homebrew installation location
* archtype.sh updated to include m1 arch
* UTF-8 unit tests were failing, so I've updated those

## Test Plan

Tests run locally on my M1 machine.
Will need to update CI to include M1 arm64 test support.


Eric-Warehime:
Requires https://github.com/algorand/go-algorand/pull/3920 since we depend on some go-algorand pkgs in indexer as well.

util/util_test.go
```diff
@@ -22,7 +21,7 @@ func TestPrintableUTF8OrEmpty(t *testing.T) {
 		},
 		{
 			"Asset 510285544",
-			"8J+qmSBNb25leQ==",
```
winder, line 25: Why did this need to change?

Eric-Warehime: Not sure of the exact root cause, but from what I can tell, that string is `🪙 Money, want` which returns a printable string (failing the test which expects it not to be printable) while the new string `🪙 Money, want\n` includes newline and is not printable (as anticipated) on the newline character.

It looks like the most recent update to the unicode bits in go is https://github.com/golang/go/commit/8ec5a052ec8d4e628353fb5099b29341fae9b3a4 which started in 1.16. That could be the root cause if we were expecting the moon symbol to fail since we're on 1.13. But I haven't dug that deep on it.

Eric-Warehime: This also reminded me that I forgot to update golang in this PR. Updating with that.

Eric-Warehime: https://github.com/algorand/indexer/pull/943 also includes changes to this test so it must be failing on golang1.16 as well.

cmd/import-validator/core/service.go
```diff
@@ -255,14 +257,31 @@ func checkModifiedState(db *postgres.IndexerDb, l *ledger.Ledger, block *bookkee
...
+				case basics.AppCreatable:
+					var appResource ledgercore.AppResource
+					appResource, err1 = l.LookupApplication(block.Round(), address, basics.AppIndex(creatable.Index))
+					if err1 != nil {
+						err1 = fmt.Errorf("checkModifiedState() lookup application resource err1: %w", err1)
+						return
+					}
+					resourcesForAddress[creatable] = normalizeAccountResource(appResource)
+				default:
+					err1 = fmt.Errorf("checkModifiedState() unexpected creatable type %v", creatable.Type)
```
winder, line 282: @tolikzinovyev could you review this section? It seems reasonable, but should we call `normalizeAccountResource(ar)` for the `AssetCreatable` case?

winder:
`api/Makefile` - The oapi-codegen version should be changed to v1.3.7, and `go get` changed to `go install`
`Makefile` - Use `go install` instead of `go get` to install mockery.
`.circleci/config.yml` - Use `go install` instead of `go get` to install golint

cmd/import-validator/core/service.go
tolikzinovyev: I would only suggest to do one of the following:

1. Keep `normalizeAccountResource()` unchanged and call it after the `switch`.
2. Change `normalizeAccountResource()` into `normalizeAppResource()` that would accept and return `ledgercore.AppResource`.

Eric-Warehime: Went with changing `normalizeAccountResource` to `convertAppResource` which will return an `AccountResource` with normalized fields. I think it's more properly named and avoids calling the function when it's not necessary for `AssetResource`s.

winder (approved):
LGTM

cmd/import-validator/core/service.go
Eric-Warehime: @tolikzinovyev Let me know if you have any feedback on these changes. 😄

tolikzinovyev: Looks fine!

winder (approved):

PR #982, opened 2022-04-29
Developer Tools: indexer block processor

shiqizng:
## Summary
add a block processor module. It writes a validated block to a local ledger.  

## Test Plan
Unit tests for the module

### DIFF
--- a/processor/blockprocessor/block_processor.go
+++ b/processor/blockprocessor/block_processor.go
@@ -20,6 +20,17 @@ func MakeProcessor(ledger *ledger.Ledger, handler func(block *ledgercore.Validat
 	if ledger == nil {
 		return nil, fmt.Errorf("MakeProcessor(): local ledger not initialized")
 	}
+	if handler != nil && ledger.Latest() == 0 {
+		blk, err := ledger.Block(0)
+		if err != nil {
+			return nil, fmt.Errorf("MakeProcessor() err: %w", err)
+		}
+		vb := ledgercore.MakeValidatedBlock(blk, ledgercore.StateDelta{})
+		err = handler(&vb)
+		if err != nil {
+			return nil, fmt.Errorf("MakeProcessor() handler err: %w", err)
+		}
+	}
 	return &blockProcessor{ledger: ledger, nextRoundToProcess: uint64(ledger.Latest() + 1), handler: handler}, nil
 }
 
--- a/processor/blockprocessor/block_processor_test.go
+++ b/processor/blockprocessor/block_processor_test.go
@@ -100,10 +100,13 @@ func TestFailedProcess(t *testing.T) {
 	handler := func(vb *ledgercore.ValidatedBlock) error {
 		return fmt.Errorf("handler error")
 	}
-	pr, _ = block_processor.MakeProcessor(l, handler)
+	_, err = block_processor.MakeProcessor(l, handler)
+	assert.Contains(t, err.Error(), "MakeProcessor() handler err")
+	pr, _ = block_processor.MakeProcessor(l, nil)
 	txn = test.MakePaymentTxn(0, 10, 0, 1, 1, 0, test.AccountA, test.AccountA, basics.Address{}, basics.Address{})
 	block, err = test.MakeBlockForTxns(genesisBlock.BlockHeader, &txn)
 	assert.Nil(t, err)
+	pr.SetHandler(handler)
 	rawBlock = rpcs.EncodedBlockCert{Block: block, Certificate: agreement.Certificate{}}
 	err = pr.Process(&rawBlock)
 	assert.Contains(t, err.Error(), "Process() handler err")


processor/processor.go
```diff
@@ -0,0 +1,242 @@
...
+			return fmt.Errorf("Process() err: %w", err)
+		}
+		defer ledgerForEval.Close()
+		resources, _ := prepareEvalResources(&ledgerForEval, &cert.Block)
+		if err != nil {
+			panic(fmt.Errorf("Process() eval err: %w", err))
+		}
+
+		delta, _, err :=
+			ledger.EvalForIndexer(ledgerForEval, &cert.Block, proto, resources)
```
winder, line 95: Did you look into using StartEvaluator instead of EvalForIndexer? It's interesting that by implementing the `LedgerForEval` interface we could theoretically swap between the local ledger and postgres, but I'm not sure if we would ever want to do that.

processor/processor.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/rpcs"
+	"github.com/algorand/indexer/accounting"
+)
+
+type Processor interface {
+	Process(cert *rpcs.EncodedBlockCert) error
+	SetHandler(handler func(block *ledgercore.ValidatedBlock) error)
+	GetLastProcessedRound() uint64
```
winder, line 22: I guess I propogated that initialization error by suggesting this interface. It should be much easier by changing to this format.
```suggestion
	NextRoundToProcess() uint64
```

processor/ledger_for_evaluator.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	"fmt"
+
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/ledger"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+)
+
+// LedgerForEvaluator implements the indexerLedgerForEval interface from
```
winder, line 12: Skipping over this file for now. Hopefully we can use existing go-algorand code to handle fetching the inputs.

winder (commented):
Looks like a good start

processor/processor.go
```diff
@@ -0,0 +1,103 @@
...
+		return fmt.Errorf("Process() err: block has invalid round")
+	}
+
+	if cert.Block.Round() == basics.Round(0) {
+		// Block 0 is special, we cannot run the evaluator on it
+		err := processor.ledger.AddBlock(cert.Block, agreement.Certificate{})
+		if err != nil {
+			return fmt.Errorf("error adding round  %v to local ledger", err.Error())
+		}
+		processor.nextRoundToProcess = uint64(cert.Block.Round()) + 1
```
AlgoStephenAkiki, line 77: Why not move this and the one on line 93 to after the overall if statement?

Eric-Warehime (commented):
I'm still becoming familiar with the idiomatic way to do unit testing in Go, so it's possible there's a more preferred style here. But I think in general it would be nice to get more coverage of the failure scenarios by splitting out the tests more. For example, `TestInitFailureInvalidLedgerPath` or `TestProcessFailureInvalidGenesisAddress`.

I've left a few comments in different places with some ideas on how we can test individual failure cases in both `Init` and `Process`. 
In cases where it seems useful we can add things to our test utils which will help with generating blocks/ledgers with states that are generally useful for validating failure scenarios in different parts of indexer.

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+
+	paysetgroups, err := cert.Block.DecodePaysetGroups()
+	if err != nil {
+		return fmt.Errorf("Process() decode payset groups err: %w", err)
+	}
+
+	for _, group := range paysetgroups {
+		err = blkeval.TransactionGroup(group)
+		if err != nil {
+			return fmt.Errorf("Process() apply transaction group err: %w", err)
```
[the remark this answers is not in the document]
shiqizng: I'll add a few basic test cases since most failure cases should be covered in ledger's unit tests already.

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	//validated block
+	vb, err := blkeval.GenerateBlock()
+	if err != nil {
+		return fmt.Errorf("Process() validated block err: %w", err)
+	}
+	//	write to ledger
+	err = processor.ledger.AddValidatedBlock(*vb, agreement.Certificate{})
```
tolikzinovyev, line 95: We should either pass `cert.Certificate` here or have a `bookkeeping.Block` parameter in `Process()`.

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/rpcs"
+)
+
+// Processor is the block processor interface
+type Processor interface {
+	Process(cert *rpcs.EncodedBlockCert) error
+	SetHandler(handler func(block *ledgercore.ValidatedBlock) error)
+	NextRoundToProcess() uint64
+	Init(path string, genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block) error
```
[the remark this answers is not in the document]
tolikzinovyev: Additionally, I think `MakeProcessor()` should accept only `ledger *ledger.Ledger` and `handler`. The rationale is easier testability. You will be able to pass an in-memory `Ledger` instead of creating a temporary directory on disk (pass `true` to `ledger.OpenLedger()`).

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+	processor.nextRoundToProcess = uint64(cert.Block.Round()) + 1
+	return nil
+}
+
+func (processor *processorImpl) SetHandler(handler func(block *ledgercore.ValidatedBlock) error) {
+	processor.handler = handler
+}
+func (processor *processorImpl) NextRoundToProcess() uint64 {
```
tolikzinovyev, line 112: nit: add empty line

processor/processor_test.go
```diff
@@ -0,0 +1,71 @@
+package processor
```
tolikzinovyev, line 1: It's better to do black-box testing. To achieve it, use `package processor_test`.

processor/processor.go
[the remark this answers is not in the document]
tolikzinovyev: https://en.wikipedia.org/wiki/Dependency_injection

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/ledger"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/rpcs"
+)
+
+// Processor is the block processor interface
+type Processor interface {
```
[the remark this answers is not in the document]
shiqizng: using an interface would allow for different implementations of Processor.

processor/processor.go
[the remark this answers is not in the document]
shiqizng: I'm not sure `MakeProcessor(ledger *ledger.Ledger, handler func(block *ledgercore.ValidatedBlock))` is what we want. This definition seems to be having block fetcher and addBlock to create a ledger instance. Shouldn't processor be owning the local ledger?

processor/processor.go
tolikzinovyev, line 95: I think writing to ledger should happen after calling the handler. In case, we crash between the two, we want to be able to resume.

processor/processor.go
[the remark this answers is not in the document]
tolikzinovyev: Cool! I think another good benefit is that the user will not be able to instantiate a processor object without using a "constructor" function (e.g., `Processor{}`).

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+	ledger             *ledger.Ledger
+}
+
+// MakeProcessor creates a block processor
+func MakeProcessor() Processor {
+	return &processorImpl{}
+}
+
+// Init block processor. Create ledger if it wasn't initialized before
+func (processor *processorImpl) Init(ledgerPath string, genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block) error {
```
winder, line 37: Since these pointers are de-referenced, you should verify that they aren't nil.

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return fmt.Errorf("openLedger() open err: %w", err)
+	}
+	processor.ledger = ledger
+	processor.nextRoundToProcess = 1
+	return nil
+}
+
+// Process a raw algod block
+func (processor *processorImpl) Process(cert *rpcs.EncodedBlockCert) error {
```
winder, line 64: nil parameter check.

processor/processor.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Processor is the block processor interface
+type Processor interface {
+	Process(cert *rpcs.EncodedBlockCert) error
+	SetHandler(handler func(block *ledgercore.ValidatedBlock) error)
+	NextRoundToProcess() uint64
+	Init(path string, genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block) error
+}
+
+type processorImpl struct {
```
winder, line 25: Could you rename this to `blockProcessor` and move the implementation into a new package?

processor/processor.go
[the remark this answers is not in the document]
winder: I think we can have the best of both approaches with two constructors:
* `MakeBlockProcessor(path string, genesis ...)` - self contained when part of a pipeline. 
* `makeBlockProcessor(ledger *ledger.Ledger)` - easy to inject the ledger for tests.

processor/processor_test.go
```diff
@@ -0,0 +1,71 @@
...
+	defer os.RemoveAll(dir)
+	assert.Nil(t, err)
+
+	processor := processorImpl{}
+	err = processor.Init(dir, &genesis, &genesisBlock)
+	assert.Nil(t, err)
+	assert.NotNil(t, processor.ledger)
+}
+
+func TestProcess(t *testing.T) {
```
winder, line 28: Not that this test is especially complicated, but Tolik's suggestion should make it even simpler.

The separate Init tests can be done against against `MakeBlockProcessor`.
Process tests can focus on the logic using an in-memory ledger.

processor/processor_test.go
winder: What is the benefit to black box testing vs just using the public methods?

One pattern I like is having private helpers (i.e. for dependency injection) to make testing easier, while having simple public functions.

winder (changes requested):
A couple more things in addition to what others said.

processor/processor_test.go
tolikzinovyev: In general, white box tests are bad because they are fragile. When internal implementation of a unit changes, the tests also need to change. However, if the tests only invoke the public API, which rarely changes, changes in code don't break the tests.

processor/processor.go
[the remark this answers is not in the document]
tolikzinovyev: I think this approach works for testing right now but it's not the best. 

1. it forces tests to be white box
2. if you write tests for a different package that depend on this package, you will not be able to use an in-memory ledger
3. in general this approach is inferior for a reason unrelated to testing; imagine you need to construct an object that itself constructs 10 objects with 5 parameters each. Instead of passing 50 parameters to the main object, it's better to construct 10 smaller objects and pass all of them to the main object.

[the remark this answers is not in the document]
shiqizng: MakeBlockProcessor(path string, genesis ...) is really creating a local ledger, maybe I can rename it to CreateLedger and make it a helper method in`util`? and use MakeBlockProcessor(ledger *ledger.Ledger, handler func(block *ledgercore.ValidatedBlock)) in `processor`?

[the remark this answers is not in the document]
tolikzinovyev: Maybe make `func MakeInitState(genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block) ledgercore.InitState` in this package? Only if you are going to test it though. Otherwise, I think it can go in `cmd/algorand-indexer/daemon.go`.

processor/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// MakeProcessor creates a block processor
+func MakeProcessor(ledger *ledger.Ledger, handler func(block *ledgercore.ValidatedBlock) error) processor.Processor {
+	return &blockProcessor{ledger: ledger, nextRoundToProcess: 1, handler: handler}
+}
+
+// Process a raw algod block
+func (processor *blockProcessor) Process(cert *rpcs.EncodedBlockCert) error {
+	if processor.ledger == nil {
```
tolikzinovyev, line 25: I think this check is not needed, but if you really want it, it should probably be in `MakeProcessor()`.

processor/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return fmt.Errorf("Process() validated block err: %w", err)
+	}
+	if processor.handler != nil {
+		err = processor.handler(vb)
+		if err != nil {
+			return fmt.Errorf("Process() handler err: %w", err)
+		}
+	}
+	//	write to ledger
```
tolikzinovyev, line 63: nit: Comments are inconsistent. Here there are two spaces after `//`, above there is zero. One space is best I think.

processor/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,78 @@
...
+// Process a raw algod block
+func (processor *blockProcessor) Process(cert *rpcs.EncodedBlockCert) error {
+	if processor.ledger == nil {
+		return fmt.Errorf("Process() err: local ledger not initialized")
+	}
+	if cert == nil {
+		return fmt.Errorf("Process() err: cannot process a nil block")
+	}
+	if uint64(cert.Block.Round()) != processor.nextRoundToProcess {
+		return fmt.Errorf("Process() err: block has invalid round")
```
tolikzinovyev, line 32: If you want to be consistent with other indexer code, `err: ...` describes the value of the variable `err`. Here, there is no variable `err`, so writing `err: ` is not appropriate.

Instead you can have
```
return fmt.Errorf("Process() invalid round cert.Block.Round(): %d processor.nextRoundToProcess: %d", cert.Block.Round(), processor.nextRoundToProcess)
```

processor/blockprocessor/block_processor_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/rpcs"
+	block_processor "github.com/algorand/indexer/processor/blockprocessor"
+	"github.com/algorand/indexer/util/test"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProcess(t *testing.T) {
+	//initialize local ledger
```
tolikzinovyev, line 21: One space after `//` would be the best.

processor/blockprocessor/block_processor_test.go
```diff
@@ -0,0 +1,127 @@
...
+	genesis := test.MakeGenesis()
+	genesisBlock := test.MakeGenesisBlock()
+	initState, err := test.CreateInitState(&genesis, &genesisBlock)
+	if err != nil {
+		log.Panicf("test init err: %v", err)
+	}
+	logger := logging.NewLogger()
+	l, err := ledger.OpenLedger(logger, "local_ledger", true, initState, config.GetDefaultLocal())
```
tolikzinovyev, line 29: I would create `func makeTestLedger(t *testing.T) *ledger.Ledger` and use it in tests: `pr := block_processor.MakeProcessor(makeTestLedger(t), handler)`. This way boring initialization can be reused across tests.

processor/blockprocessor/block_processor_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// incorrect round
+	txn := test.MakePaymentTxn(0, 10, 0, 1, 1, 0, test.AccountA, test.AccountA, test.AccountA, test.AccountA)
+	block, err := test.MakeBlockForTxns(genesisBlock.BlockHeader, &txn)
+	block.BlockHeader.Round = 10
+	assert.Nil(t, err)
+	rawBlock := rpcs.EncodedBlockCert{Block: block, Certificate: agreement.Certificate{}}
+	err = pr.Process(&rawBlock)
+	assert.Contains(t, err.Error(), "Process() err: block has invalid round")
+
+	//	non-zero balance after close remainder to sender address
```
tolikzinovyev, line 90: One space after `//` would be the best.

processor/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,78 @@
...
+			return fmt.Errorf("Process() apply transaction group err: %w", err)
+		}
+	}
+
+	//validated block
+	vb, err := blkeval.GenerateBlock()
+	if err != nil {
+		return fmt.Errorf("Process() validated block err: %w", err)
+	}
+	if processor.handler != nil {
```
tolikzinovyev, line 57: You could add a comment about calling the handler before writing to ledger.

processor/processor.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/rpcs"
+)
+
+// Processor is the block processor interface
+type Processor interface {
+	Process(cert *rpcs.EncodedBlockCert) error
+	SetHandler(handler func(block *ledgercore.ValidatedBlock) error)
```
tolikzinovyev, line 11: Do we still need this function?

processor/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,78 @@
...
+
+type blockProcessor struct {
+	handler            func(block *ledgercore.ValidatedBlock) error
+	nextRoundToProcess uint64
+	ledger             *ledger.Ledger
+}
+
+// MakeProcessor creates a block processor
+func MakeProcessor(ledger *ledger.Ledger, handler func(block *ledgercore.ValidatedBlock) error) processor.Processor {
+	return &blockProcessor{ledger: ledger, nextRoundToProcess: 1, handler: handler}
```
tolikzinovyev, line 20: shouldn't `nextRoundToProcess` be queried from `ledger`?

util/test/account_testutil.go
```diff
@@ -596,3 +598,21 @@ func MakeGenesisBlock() bookkeeping.Block {
 		},
 	}
 }
+
+// CreateInitState makes an initState
+func CreateInitState(genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block) (ledgercore.InitState, error) {
```
tolikzinovyev, line 603: I can see why you wanted to put it in `util`. Perhaps it would be better to avoid two copies. But I'm not very particular.

processor/processor.go
```diff
@@ -0,0 +1,13 @@
+package processor
+
+import (
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/rpcs"
+)
+
+// Processor is the block processor interface
+type Processor interface {
+	Process(cert *rpcs.EncodedBlockCert) error
```
tolikzinovyev, line 10: maybe rename `cert` to `block` or `blockcert`?

tolikzinovyev (commented):
I left some more comments.

Also, I'm not sure about all the `nil` checks. The function's contract can just say this argument must be non-`nil`. The function can still check if it's `nil` as a safeguard, but definitely testing such checks is unnecessary. It's a shame Go doesn't have const references.

processor/processor.go
tolikzinovyev: I think in practice it will never be needed. But you can add it back when it is!

util/test/account_testutil.go
shiqizng: I'm moving it to util.

winder (approved):
LGTM

PR #987, opened 2022-05-02
validator: skip comparison of deleted accounts.

winder:
## Summary

The recent validator change causes all HTTP errors to log an error, previously they were ignored and zero values would be compared. In the case of a deleted account, the zero values are equal. This previously ignored case must now be specifically skipped.

## Test Plan

Manual testing.

### DIFF
--- a/api/error_messages.go
+++ b/api/error_messages.go
@@ -23,7 +23,7 @@ const (
 	errFailedSearchingApplication      = "failed while searching for application"
 	errFailedLookingUpHealth           = "failed while getting indexer health"
 	errNoApplicationsFound             = "no application found for application-id"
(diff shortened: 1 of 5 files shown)


algobarb:
How come we had to change the error_message to capital `ErrNoAccountsFound`? Is it so it can be called with `api.ErrNoAccountsFound`?

winder:
> How come we had to change the error_message to capital `ErrNoAccountsFound`? Is it so it can be called with `api.ErrNoAccountsFound`?

yes exactly. Believe it or not, this is how you control visibility across packages in go.

algobarb (approved):
LGTM

PR #989, opened 2022-05-03
Add support for new block header: TxnRoot SHA256

Aharonee:


winder (commented):
This looks fine, but lets keep the PR open until we can update the submodule and handle the TODO

Eric-Warehime:
@Aharonee @algoidan Would be great to get your feedback on these changes.

Aharonee:
> @Aharonee @algoidan Would be great to get your feedback on these changes.

Looks good to me

bertani:
Just a note about the fact that while the mainnet block production has already switched to block headers containing the new sha256 txnroot field few days ago, the indexer still not having it can cause serious difficulties to external systems who need to independently verify block validation

Aharonee:
@bertani This change to the `BlockHeader` is still under ConsensusFuture so it hasn't taken affect on the mainnet as of yet.
https://github.com/algorand/go-algorand/blob/91095fa6bdccb11ec133c38da68fdd92fa611b6c/config/consensus.go#L1152

bertani:
@Aharonee then my bad, I had assumed that single the block header calculation started failing over the last week, it was due to this change, it must be for a different reason then

api/generated/common/types.go
```diff
@@ -339,6 +339,9 @@ type Block struct {
 	// \[txn\] TransactionsRoot authenticates the set of transactions appearing in the block. More specifically, it's the root of a merkle tree whose leaves are the block's Txids, in lexicographic order. For the empty block, it's 0. Note that the TxnRoot does not authenticate the signatures on the transactions, only the transactions themselves. Two blocks with the same transactions but in a different order and with different signatures will have the same TxnRoot.
 	TransactionsRoot []byte `json:"transactions-root"`
 
+	// \[txn256\] TransactionsRootSHA256 is an auxiliary TransactionRoot, built using a vector commitment instead of a merkle tree, and SHA256 hash function instead of the default SHA512_256. It is used mostly for StateProofs.
+	TransactionsRootSha256 []byte `json:"transactions-root-sha256"`
```
algoidan, line 343: should we give `TransactionsRootSha256` and `TransactionsRoot` the same names as in algod?

algoidan (approved):
LGTM

have some minor comments

api/generated/common/types.go
Eric-Warehime: I don't think algod defines the fields for a block--instead preferring to leave them as simply an opaque object which is returned. So in that sense these fields would not exist for algod to my knowledge.

api/indexer.oas3.yml
```diff
@@ -1168,6 +1168,12 @@
             "pattern": "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$",
             "type": "string"
           },
+          "transactions-root-sha256": {
+            "description": "\\[txn256\\] TransactionsRootSHA256 is an auxiliary TransactionRoot, built using a vector commitment instead of a merkle tree, and SHA256 hash function instead of the default SHA512_256. It is used mostly for StateProofs.",
```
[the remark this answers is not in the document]
Eric-Warehime: Updated.

api/generated/common/types.go
Aharonee: That field is not defined in `algod` as is, but it directly corresponds to the new fields in the `BlockHeader`.
I chose not to rename these fields (as I've renamed `TxnRoot -> TxnCommitments struct`) to stay backwards compatible to the previously defined type.

Aharonee: We could rename them and keep the `json` keys as they are though;
```
TransactionsRoot -> TransactionCommitmentsNativeSha512_256
TransactionsRootSha256 -> TransactionCommitmentsSha256
```

Eric-Warehime: There is a field in the code generator which is supposed to allow us to diverge the json from the generated class name. However, it doesn't seem to be respected by the code generator when I change/run it, and also having the two diverge in that way seems confusing from a user perspective.

I think the path forward I'd like to take here is to just maintain the existing `transactions-root` and add in the `transactions-root-shaa256` as we are currently doing here.

winder:
Closing in favor of #1002

winder (approved):

Eric-Warehime (approved):

Eric-Warehime:
Need to update https://github.com/algorand/indexer/blob/develop/api/test_resources/state_proof.response to include the expected changes.

PR #990, opened 2022-05-03
Asset id check

barnjamin:
During #977 we removed an implicit check for the options having asset id set.


winder (approved):

algobarb:
Thanks for learning and finding the fix @barnjamin . Can we merge against `develop` instead of `master`?

PR #995, opened 2022-05-10
CI: switch to codecov github action

Eric-Warehime:
## Summary

Add a codecov config file to directly configure upload steps instead of using the bash uploader since that has been deprecated.


algojack:
I know you are still working on this, but once you are ready if you can please replace upload_coverage on line 64 in .circleci/config.yml too, and delete the upload_coverage.sh script

Eric-Warehime:
I looked into why the coverage is changing in this PR (causing it to fail as intended when coverage decreases) even though we're not actually touching the code being tested.

I suspect the answer is hidden in [this bash script](https://github.com/codecov/codecov-bash/blob/43c5d3c7147b91674dd69c3e0e0becc48eba04bd/codecov) though I haven't identified the exact cause. It is re-writing the coverage file before uploading (you can see the `# path=fixes` section in a coverage file produced by the bash script like [this one](https://codecov.io/api/gh/algorand/indexer/download/build?path=v4/raw/2022-05-03/D85DC428BE251CC55F29BB7A161B78A1/2a3bde9be776b807395d622a4a55743ec4ac308d/4b63801d-38ba-4dbb-be2b-54fe30ca6ce3.txt) which does not exist in the [uploaded file](https://codecov.io/api/gh/algorand/indexer/download/build?path=v4/raw/2022-05-12/D85DC428BE251CC55F29BB7A161B78A1/36f0c77763c85280c62810ce47e7af960dc5202f/5d5c5c69-786b-4440-a48e-e4a9fd1d2e92.txt) from the new codecov uploader. This leads me to believe there is some altering of data happening affecting which files we're tracking.

Regardless of the root cause of the above, the difference we're seeing is that the previously untracked files `config/config.go`, `util/metrics/metrics.go`, and `util/diff.go` are now being reported in new coverage reports which is causing the coverage to drop. I think we can either ignore that failure and merge this or I can update the codecov config to not fail on coverage drops in this PR and submit a followup PR to enable the failure which should show a 0% drop and pass.

winder:
What do you think about setting a coverage target for new code only? We can discuss the target, I'd consider something like 75% to start but maybe we can go higher?

https://docs.codecov.com/docs/common-recipe-list#ensure-all-code-is-covered

Eric-Warehime:
Discussed this morning and we've agreed to track the coverage % in the patch, and to initially prevent it from blocking PRs on failure. In the future as we get more complete coverage we will enable it to block PRs on failure.

winder (approved):

cce:
updated title, this uses a circleci orb, not a github action

PR #999, opened 2022-05-12
Developer Tools: block processor integration

shiqizng:
## Summary
Integrate block fetcher and addBlock with block processor

## Test Plan
Existing e2e test passes, run indexer daemon to validate block importing works.


shiqizng:
This is will be part of another PR.

PR #1002, opened 2022-05-17
state proofs: Initial commit for stateproofs model

Eric-Warehime:
## Summary

I'm opening this PR to integrate the indexer models w/ the new changes being worked on for state proofs. Much of the implementation code for that is located in a branch in go-algorand at the moment (https://github.com/algorand/go-algorand/tree/feature/stateproofs) so this PR will probably have to remain open until that is merged into go-algorand master.


api/converter_utils.go
```diff
@@ -407,6 +408,99 @@ func signedTxnWithAdToTransaction(stxn *transactions.SignedTxnWithAD, extra rowD
 		}
 
 		application = &a
+	case protocol.StateProofTx:
+		partPath := make([]generated.GenericDigest, len(stxn.Txn.StateProof.PartProofs.Path))
+		for idx, part := range stxn.Txn.StateProof.PartProofs.Path {
+			partPath[idx] = generated.GenericDigest(part)
```
winder, line 414: I believe you need to make a copy of `part` because the pointer is reused by the for loop.

winder: Here's an example where we do something similar: https://github.com/algorand/indexer/pull/575/files

idb/postgres/internal/writer/writer.go
```diff
@@ -3,6 +3,7 @@ package writer
 import (
 	"context"
 	"fmt"
+	"github.com/algorand/go-algorand/protocol"
```
winder, line 6: nit: fix import grouping

idb/postgres/internal/writer/writer.go
```diff
@@ -159,7 +160,7 @@ func getSigTypeDeltas(payset []transactions.SignedTxnInBlock) (map[basics.Addres
 	res := make(map[basics.Address]sigTypeDelta, len(payset))
 
 	for i := range payset {
-		if payset[i].Txn.RekeyTo == (basics.Address{}) {
+		if payset[i].Txn.RekeyTo == (basics.Address{}) && payset[i].Txn.Type != protocol.StateProofTx {
```
winder, line 163: Did you consider pushing this logic down into `idb.SignatureType`? It could return `None`.

Maybe it's fine like this, makes it clear that there is a special case relating to state proofs.

misc/e2elive.py
```diff
@@ -57,7 +57,7 @@ def main():
         from botocore.config import Config
         from botocore import UNSIGNED
         s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
-        tarname = 'net_done.tar.bz2'
+        tarname = 'feature-stateproofs.tar.bz2'
```
winder, line 60: stateproofs is merged! We can change this:
```suggestion
        tarname = 'rel-nightly.tar.bz2'
```

idb/postgres/internal/writer/writer.go
Eric-Warehime: Yeah I actually was trying fit it into that function first, but it returns `(SigType, error)` where `SigType("")` is always an error because we essentially have an unsigned Txn in the block. It just seemed to be more confusing to either have an empty SigType that isn't an error in there, or one that is so I just decided to bypass checking the sig type altogether for StateProofs.

api/indexer.oas2.json
```diff
@@ -2228,6 +2258,215 @@
...
+      }
+    },
+    "StateProofSigSlot": {
+      "type": "object",
+      "properties": {
+        "signature": {
+          "description": "",
+          "$ref": "#/definitions/StateProofSignature"
+        },
+        "lower-sig-weight": {
```
algoidan, line 2360: nits: I was not able to find an exact name for this field. And I'm not sure this is accurate. can we rename it to `LValue`?

api/converter_utils.go
```diff
@@ -407,6 +408,101 @@ func signedTxnWithAdToTransaction(stxn *transactions.SignedTxnWithAD, extra rowD
...
+							},
+							Path:      &proofPath,
+							TreeDepth: uint64Ptr(uint64(revToConv.SigSlot.Sig.Proof.TreeDepth)),
+						},
+						VerifyingKey: &verifyKey,
+					},
+				},
+			}
+		}
+		proof := generated.StateProof{
```
algoidan, line 471: I think we should also copy `PositionsToReveal` slice

algoidan (commented):
Overall looks good
thanks!
There is only one issue which I wasn't sure

api/indexer.oas2.json
algoidan: I guess it is okay. thanks.

api/converter_utils.go
winder: @Eric-Warehime You can use `byteSlicePtr` to create a copy

misc/e2elive.py
```diff
@@ -82,8 +82,8 @@ def main():
         import boto3
         from botocore.config import Config
         from botocore import UNSIGNED
-
         s3 = boto3.client("s3", config=Config(signature_version=UNSIGNED))
+        s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
```
winder, line 86: Merge conflict with the python formatting, maybe?
```suggestion
        s3 = boto3.client("s3", config=Config(signature_version=UNSIGNED))
```

Eric-Warehime: Yeah I guess I messed up rebasing off of develop.

winder (approved):

PR #1005, opened 2022-05-20
api model: Update model to converge with algod

Eric-Warehime:
## Summary

I used this updated indexer spec to re-generate the java and go sdks.  
The generator uses this new definition to ensure that the HashType enum is not deleted in the Java Sdk despite indexer not using it. The generated go sdk is not changed by this.

See https://github.com/algorand/generator/issues/19 for more info on some of the issues we're solving here.


Eric-Warehime:
@winder @shiqizng @AlgoStephenAkiki Ping for review :octocat:

AlgoStephenAkiki (approved):

PR #1010, opened 2022-05-24
api: return empty lists from fetchApplications and fetchAppLocalStates

erer1243:
Currently, `fetchApplications` and `fetchAppLocalStates` in `handlers.go` declare their result slices using `var`. This causes those slices to be initialized to `nil`. In turn, when there are no applications/app local states to be found, the rest endpoints that use these methods end up returning `null` in the response json instead of an empty list. E.g. querying `/v2/accounts/{account-id}/apps-local-state` for an account not participating in any contracts will return `{ 'apps-local-states': null, 'round': ... }`. This has caused some confusion, as the rest api docs do not acknowledge this possibility. 

This PR changes the behavior of those two methods to create their slices using `make`. This causes the endpoints to return empty lists instead of null.


winder (approved):

winder:
Thanks for the PR!

PR #1012, opened 2022-06-01
New Feature: Adds Data Directory Support

### LINKED ISSUE #997: Use Indexer Data Dir
We should use the data dir added with #969 to automatically:
* create a pidfile (if not provided with pidfile flag)
* load param config file if `param_config.yml` exists (and flag not used)
* other files? token file?
* indexer config file.

Passing in the existing flag for these options would override any file that exists in the data directory.

The daemon code does not have very many tests. We should make sure we improve the coverage with this work.

AlgoStephenAkiki:
Resolves #997

This PR does the following:

- Updates the go-algorand submodule hash to point to rel/beta
- Moves the cpu profiling file, pid file and indexer configuration file
  to be options of only the daemon sub-command
- Changes os.Exit() to be a panic with a special handler.  This is so
  that defer's are handled instead of being ignored.
- Detects auto-loading configuration files in the data directory and
  issues errors if equivalent command line arguments are supplied.
- Updates the README with instructions on how to use the auto-loading
  configuration files and the data directory.


AlgoStephenAkiki:
@shiqizng  Please review to minimize merge conflicts with your branch.
@Eric-Warehime Please review to check that this still works with M1 Mac 
@winder General review

README.md
```diff
@@ -195,6 +195,7 @@ Settings can be provided from the command line, a configuration file, or an envi
 | default-balances-limit        |         | default-balances-limit        | INDEXER_DEFAULT_BALANCES_LIMIT        |
 | max-applications-limit        |         | max-applications-limit        | INDEXER_MAX_APPLICATIONS_LIMIT        |
 | default-applications-limit    |         | default-applications-limit    | INDEXER_DEFAULT_APPLICATIONS_LIMIT    |
+| data-dir                      | i       | data-dir                      | INDEXER_DATA_DIR                      |
```
winder, line 198: I left this comment on Shiqi's PR also. Could we add a viper alias so that this is mapped to `INDEXER_DATA`? That would allow us to skip the manual env variable checking.

README.md
```diff
@@ -232,6 +233,26 @@ If it is not in the current working directory along with the indexer command we
 ~$ ./algorand-indexer daemon -c <full-file-location>/indexer.yml
 ```
 
+## Data Directory
```
winder, line 236: I think it makes sense to guide people towards storing things in the data directory. To that end it might be good to:
1) put this section ahead of the config file section.
2) remove the bullet points about how config files are discovered if none are provided (or mark it as deprecated).

cmd/algorand-indexer/daemon.go
```diff
@@ -199,6 +293,10 @@ func init() {
 
 	daemonCmd.Flags().StringVarP(&indexerDataDir, "data-dir", "i", "", "path to indexer data dir, or $INDEXER_DATA")
 
+	daemonCmd.Flags().StringVarP(&cpuProfile, "cpuprofile", "", "", "file to record cpu profile to")
+	daemonCmd.Flags().StringVarP(&pidFilePath, "pidfile", "", "", "file to write daemon's process id to")
+	daemonCmd.Flags().StringVarP(&configFile, "configfile", "c", "", "file path to configuration file (indexer.yml)")
+
 	viper.RegisterAlias("algod", "algod-data-dir")
```
winder, line 300: Here are the alias', I think something like `viper.RegisterAlias("data-dir", "indexer-data")` should do it.

cmd/algorand-indexer/main.go
```diff
@@ -218,8 +194,11 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Setup our exit handler for maybeFail() and other exit panics
+	defer exitHandler()
+
 	if err := rootCmd.Execute(); err != nil {
 		logger.WithError(err).Error("an error occurred running indexer")
-		os.Exit(1)
+		panic(exit{1})
```
winder, line 202: If we get here any defers inside the execute function would have already been processed, so no need to use the panic trick.

cmd/algorand-indexer/main.go
```diff
@@ -23,12 +22,36 @@ import (
...
+
+// Calling os.Exit() directly will not honor any defer'd statements.
+// Instead, we will create an exit type and handler so that we may panic
+// and handle any exit specific errors
+type exit struct {
+	RC int // The exit code
+}
+
+// exitHandler will handle a panic with type of exit (see above)
+func exitHandler() {
```
winder, line 36: Neat trick, nice catch.

go.mod
```diff
@@ -20,9 +20,9 @@ require (
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.10.0
 	github.com/sirupsen/logrus v1.8.1
-	github.com/spf13/cobra v0.0.5
+	github.com/spf13/cobra v1.3.0
 	github.com/spf13/pflag v1.0.5
-	github.com/spf13/viper v1.7.1
+	github.com/spf13/viper v1.10.0
```
winder, line 25: Anything good in the new versions?

cmd/algorand-indexer/daemon.go
```diff
@@ -49,50 +52,135 @@ var (
...
+		indexerConfigFound := util.FileExists(filepath.Join(indexerDataDir, autoLoadIndexerConfigName))
+		paramConfigFound := util.FileExists(filepath.Join(indexerDataDir, autoLoadParameterConfigName))
+
+		// If we auto-loaded configs but a user supplied them as well, we have an error
+		if indexerConfigFound {
+			if configFile != "" {
+				logger.Errorf(
+					"indexer configuration was found in data directory (%s) as well as supplied via command line.  Only provide one.",
+					filepath.Join(indexerDataDir, autoLoadIndexerConfigName))
+				panic(exit{1})
```
winder, line 98: Instead of a panic, could we return for these?
```suggestion
				return
```

cmd/algorand-indexer/main.go
```diff
@@ -23,12 +22,36 @@ import (
 	"github.com/algorand/indexer/version"
 )
 
+const autoLoadIndexerConfigName = config.FileName + "." + config.FileType
```
winder, line 25: `FileType` is `yaml`, it looks like `yml` is also valid so we should change it to match your documentation:
https://github.com/spf13/viper/blob/da55858fffe6093e995244bfc0aa2d9fba370e30/viper.go#L302

Makefile
```diff
@@ -34,8 +34,8 @@ idb/postgres/internal/schema/setup_postgres_sql.go:	idb/postgres/internal/schema
 	cd idb/postgres/internal/schema && go generate
 
 idb/mocks/IndexerDb.go:	idb/idb.go
-	go install github.com/vektra/mockery/.../
-	cd idb && mockery -name=IndexerDb
+	go install github.com/vektra/mockery/v2@latest
+	cd idb && mockery --name=IndexerDb
```
winder, line 38: Why was this needed?

README.md
```diff
@@ -232,6 +233,26 @@ If it is not in the current working directory along with the indexer command we
...
+
+
+### Auto-Loading Configuration
+
+The Indexer will scan the data directory at startup and load certain configuration files if they are present.  The files are as follows:
+
+- Indexer Configuration File
+  - Normally supplied with the `-c` flag, if the data directory contains a file named `indexer.yml` then the Indexer will assume it is an Indexer configuration file.
+- Parameter Enable/Disable Configuration File
+  - Normally supplied with the `--api-config-file` flag, if the data directory contains a file named `api_config.yml` then the Indexer will assume it is a Parameter Enable/Disable configuration file.
```
winder, line 250: I'm not sure saying "normally" conveys that this is the new recommended way to provide configuration files. Here is a suggestion, but maybe it is too terse. What do you think?
```suggestion
- `indexer.yml` - Indexer Configuration File
- `api_config.yml` - API Parameter Enable/Disable Configuration File
```

winder (changes requested):
Requesting some small changes, and had some comments / questions.

cmd/algorand-indexer/daemon.go
AlgoStephenAkiki: I think we actually have to do `viper.RegisterAlias("data-dir", "data")`

cmd/algorand-indexer/daemon.go
AlgoStephenAkiki: Don't we want to return an error code since this would be an invalid configuration?  It would help with scripts who look at the return code of a program.

go.mod
AlgoStephenAkiki: None that I looked at, this as done as part of the `go mod tidy` cleanup I had to do to make this run.

Makefile
AlgoStephenAkiki: I wasn't able to run the integration tests with the latest version of go.  I had to update the install command to include the `v2@latest` to get it.  The `--name` was done because `-name` was an invalid argument to the latest `mockery` program.

cmd/algorand-indexer/daemon.go
winder: You're right. We could change this function to return an `error` and use more conventional error handling though. I'll let you decide.

AlgoStephenAkiki: I think consistency is important, it is probably easier to just have this panic out that way we aren't dealing with the "should I return an error or panic here" train of thought.

winder (commented):
CI is failing. Otherwise, it looks good.

cmd/algorand-indexer/daemon.go
```diff
@@ -49,50 +52,131 @@ var (
...
+
+		if indexerDataDir == "" {
+			fmt.Fprint(os.Stderr, "indexer data directory was not provided")
+			panic(exit{1})
+		}
+
+		// Create the directory if it doesn't exist
+		if _, err := os.Stat(indexerDataDir); os.IsNotExist(err) {
+			err := os.Mkdir(indexerDataDir, 0755)
+			maybeFail(err, "indexer data directory error, %v", err)
```
winder, line 81:
```suggestion
			maybeFail(err, "failure while creating data directory: %v", err)
```

cmd/algorand-indexer/daemon.go
```diff
@@ -49,50 +52,131 @@ var (
...
+				if err != nil {
+					maybeFail(err, "%v", err)
+				}
+				defer configs.Close()
+				err = viper.ReadConfig(configs)
+				if err != nil {
+					maybeFail(err, "invalid config file (%s): %v", viper.ConfigFileUsed(), err)
+				}
+				fmt.Printf("Using configuration file: %s\n", configFile)
+			}
```
winder, line 107: nit: the linter would have made this suggestion if we were using normal error handling
```suggestion
			}
			// No config file supplied via command line, auto-load it
			configs, err := os.Open(configFile)
			if err != nil {
				maybeFail(err, "%v", err)
			}
			defer configs.Close()
			err = viper.ReadConfig(configs)
			if err != nil {
				maybeFail(err, "invalid config file (%s): %v", viper.ConfigFileUsed(), err)
			}
			fmt.Printf("Using configuration file: %s\n", configFile)
```

cmd/algorand-indexer/daemon.go
```diff
@@ -49,50 +52,131 @@ var (
...
+		if paramConfigFound {
+			if suppliedAPIConfigFile != "" {
+				logger.Errorf(
+					"api parameter configuration was found in data directory (%s) as well as supplied via command line.  Only provide one.",
+					filepath.Join(indexerDataDir, autoLoadParameterConfigName))
+				panic(exit{1})
+			} else {
+				suppliedAPIConfigFile = filepath.Join(indexerDataDir, autoLoadParameterConfigName)
+				fmt.Printf("Auto-loading parameter configuration file: %s", suppliedAPIConfigFile)
+			}
```
winder, line 119: nit: same as above
```suggestion
			if suppliedAPIConfigFile != "" {
				logger.Errorf(
					"api parameter configuration was found in data directory (%s) as well as supplied via command line.  Only provide one.",
					filepath.Join(indexerDataDir, autoLoadParameterConfigName))
				panic(exit{1})
			}
			suppliedAPIConfigFile = filepath.Join(indexerDataDir, autoLoadParameterConfigName)
			fmt.Printf("Auto-loading parameter configuration file: %s", suppliedAPIConfigFile)
```

cmd/algorand-indexer/daemon.go
```diff
@@ -49,50 +52,131 @@ var (
...
+				configs, err := os.Open(configFile)
+				if err != nil {
+					maybeFail(err, "%v", err)
+				}
+				defer configs.Close()
+				err = viper.ReadConfig(configs)
+				if err != nil {
+					maybeFail(err, "invalid config file (%s): %v", viper.ConfigFileUsed(), err)
+				}
+				fmt.Printf("Using configuration file: %s\n", configFile)
```
winder, line 106: It looks like we already print this below, so no need for it here.
```suggestion
```

winder (changes requested):
Noticed a few more small things

README.md
shiqizng: @AlgoStephenAkiki does exporting `INDEXER_DATA_DIR` work for you? after adding the alias, `viper.RegisterAlias("data-dir", "data")`, only `INDEXER_DATA` is working for me.

AlgoStephenAkiki: That's a good point.  Do we want both?  Performing the alias I think removes the DATA_DIR variant.

winder: I don't think we need both, lets update the docs to refer to `INDEXER_DATA`. Nice catch.

winder (approved):

PR #1013, opened 2022-06-02
Local Ledger Deployment

shiqizng:
## Summary
This PR adds a simple migration process for local ledger. Migration completes when latest round in ledger matches the round in postgres db.  Indexer exits with code 1 if SIGTERM or SIGINT is received during the migration. When indexer restarts, ledger migration continues from round `ledger.latest()+1`. 

## Test Plan
Unit tests for the core migration logic in `RunMigration`.  Start a migration on a remote server using a mainnet db snapshot.


idb/idb.go
```diff
@@ -381,6 +381,11 @@ type IndexerDbOptions struct {
 	// This means the total number of active queries that can be running
 	// concurrently can never be more than this
 	MaxConn uint32
+
+	IndexerDatadir string
+	AlgodDataDir   string
+	AlgodToken     string
+	AlgodAddr      string
```
winder, line 388: Technically `AlgodDataDir` resolves to `AlgodToken` / `AlgodAddr`, so we could remove `AlgodDataDir`.

idb/postgres/internal/migrations/local_ledger/migration.go
```diff
@@ -0,0 +1,170 @@
...
+		return fmt.Errorf("RunMigration() err: %w", err)
+	}
+
+	localLedger, err := ledger.OpenLedger(logging.NewLogger(), filepath.Join(path.Dir(opts.IndexerDatadir), "ledger"), false, initState, algodConfig.GetDefaultLocal())
+	if err != nil {
+		return fmt.Errorf("RunMigration() err: %w", err)
+	}
+	defer localLedger.Close()
+	bot.SetNextRound(uint64(localLedger.Latest()) + 1)
+	proc, err := blockprocessor.MakeProcessor(localLedger, nil)
```
winder, line 85: Sorry for the delayed reviews. These changes are now overlapping, I think this code would be slightly simplified once OpenLedger moves into the constructor.

idb/postgres/internal/migrations/local_ledger/migration_test.go
```diff
@@ -0,0 +1,130 @@
...
+		httpmock.NewStringResponder(200, string(json.Encode(test.MakeGenesis()))))
+	// /v2/blocks/0 endpoint
+	blockCert := rpcs.EncodedBlockCert{
+		Block: test.MakeGenesisBlock(),
+	}
+	// /v2/blocks/0
+	httpmock.RegisterResponder("GET", "http://localhost/v2/blocks/0?format=msgpack",
+		httpmock.NewStringResponder(200, string(msgpack.Encode(blockCert))))
+
+	// /v2/blocks/1
```
winder, line 38: It looks like this could be a loop. The main benefit to this is that it becomes obvious that identical blocks are used
```
txn := test.MakePaymentTxn(0, 100, 0, 1, 1,
	0, test.AccountA, test.AccountA, basics.Address{}, basics.Address{})
for i := 1; i < 7; i++ {
	block, err := test.MakeBlockForTxns(test.MakeGenesisBlock().BlockHeader, &txn)
	assert.Nil(t, err)
	blockCert := rpcs.EncodedBlockCert{
		Block: block,
	}
	url := fmt.Sprintf("http://localhost/v2/blocks/%d?format=msgpack", i)
	httpmock.RegisterResponder("GET", url,
		httpmock.NewStringResponder(200, string(msgpack.Encode(blockCert))))
}
```

Maybe a regex would work also? Return the same thing until the blocks are in double digits:
```
	txn := test.MakePaymentTxn(0, 100, 0, 1, 1,
		0, test.AccountA, test.AccountA, basics.Address{}, basics.Address{})
	block, err := test.MakeBlockForTxns(test.MakeGenesisBlock().BlockHeader, &txn)
	assert.Nil(t, err)
	blockCert := rpcs.EncodedBlockCert{
		Block: block,
	}

	// Return the same block for round 1-9
	httpmock.RegisterResponder("GET", `=~^http://localhost/v2/blocks/\d?format=msgpack`,
		httpmock.NewStringResponder(200, string(msgpack.Encode(blockCert))))
```

idb/postgres/internal/migrations/local_ledger/migration_test.go
```diff
@@ -0,0 +1,130 @@
...
+	err = RunMigration(3, &opts)
+	assert.NoError(t, err)
+	genesis := test.MakeGenesis()
+	genesisBlock := test.MakeGenesisBlock()
+	initState, err := util.CreateInitState(&genesis, &genesisBlock)
+	assert.NoError(t, err)
+	l, err := ledger.OpenLedger(logging.NewLogger(), filepath.Join(path.Dir(opts.IndexerDatadir), "ledger"), false, initState, algodConfig.GetDefaultLocal())
+	assert.NoError(t, err)
+	// check 3 rounds written to ledger
+	assert.Equal(t, uint64(3), uint64(l.Latest()))
```
winder, line 108: Nice test. :+1:

idb/postgres/internal/migrations/local_ledger/migration_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// migration continues from last round
+	err = RunMigration(5, &opts)
+	assert.NoError(t, err)
+	l, err = ledger.OpenLedger(logging.NewLogger(), filepath.Join(path.Dir(opts.IndexerDatadir), "ledger"), false, initState, algodConfig.GetDefaultLocal())
+	assert.NoError(t, err)
+	assert.Equal(t, uint64(5), uint64(l.Latest()))
+	l.Close()
+
+	// remove ledger files
```
winder, line 119: Temp dir + recursive delete?

winder (changes requested):
This looks like it should work. I think it's diverged a little from the first PR.

idb/idb.go
shiqizng: I thought both were needed since there are 2 ways for the fetcher to create an algod client, https://github.com/algorand/indexer/blob/df6bcb6ced21a354bbc80daf41642c81226c1d40/idb/postgres/internal/migrations/local_ledger/migration.go#L50-L63

winder: The `ForDataDir` would do something like resolve the address/token and then call `ForNetAndToken`

winder (approved):
LGTM

PR #1024, opened 2022-06-10
[WIP] recovery scenario

shiqizng:
## Summary
This PR implements a recovery mechanism for ledger. If the local disk fails and we lose the data directory, the indexer will reinitialize the ledger. 

## Test Plan

run daemon manually.


processor/blockprocessor/block_processor.go
```diff
@@ -20,6 +22,8 @@ import (
 	"github.com/algorand/indexer/util"
 )
 
+const Prefix = "ledger"
```
winder, line 25: This is failing the build because there's no comment
```suggestion
const prefix = "ledger"
```

processor/blockprocessor/block_processor.go
```diff
@@ -38,12 +42,20 @@ func MakeProcessorWithLedger(l *ledger.Ledger, handler func(block *ledgercore.Va
...
 		return nil, fmt.Errorf("MakeProcessor() err: %w", err)
 	}
-	l, err := ledger.OpenLedger(logging.NewLogger(), filepath.Join(path.Dir(datadir), "ledger"), false, initState, algodConfig.GetDefaultLocal())
+	if dbRound != 0 && !ledgerExists(datadir, Prefix) {
+		msg := fmt.Sprintf("%s\n%s\n%s\n%s\n",
+			"The ledger cache was not found in the data directory and must be initialized. There are several ways to initialize it:",
+			"1.Fetch blocks and re-initialize, this takes a long time and is the most secure",
+			"2.Initialize with a catchpoint, this requires trusting that the relay is providing the correct ledger snapshot.",
+			fmt.Sprintf("3.Copy files ledger.block.sqlite and ledger.tracker.sqlite from an existing node installation from before round %d into the indexer data directory.", dbRound))
+		return nil, fmt.Errorf("MakeProcessor() err: %s", msg)
```
winder, line 56: I don't think it's needed for this PR, maybe not even this release, but we should find a better way to get this information to the user.

processor/blockprocessor/block_processor.go
```diff
@@ -38,12 +42,20 @@ func MakeProcessorWithLedger(l *ledger.Ledger, handler func(block *ledgercore.Va
...
+	}
+	l, err := ledger.OpenLedger(logging.NewLogger(), filepath.Join(path.Dir(datadir), Prefix), false, initState, algodConfig.GetDefaultLocal())
 	if err != nil {
 		return nil, fmt.Errorf("MakeProcessor() err: %w", err)
 	}
```
winder, line 61: There are some edge cases where the ledger is ahead of the required round.
```suggestion
	}
	if l.Latest() > dbRound {
		return nil, fmt.Errorf("MakeProcessor() err: the ledger cache is ahead of the required round and must be re-initialized.")
	}
```

cmd/algorand-indexer/daemon.go
```diff
@@ -133,6 +136,19 @@ var daemonCmd = &cobra.Command{
 				fmt.Fprint(os.Stderr, "missing indexer data directory")
 				os.Exit(1)
 			}
+
+			// sync local ledger
```
winder, line 140: Not required now, but I think we'll need to figure out a way to put this in an Init function thats part of the block processor.

There is also an edge case we could check to fail fast: if the catchpoint is ahead of nextDBRound we should avoid fast catchup, and probably inform the user that they might want to finish catchup with an earlier version of Indexer.

winder (approved):

processor/blockprocessor/block_processor.go
```diff
@@ -38,15 +42,26 @@ func MakeProcessorWithLedger(l *ledger.Ledger, handler func(block *ledgercore.Va
...
 // MakeProcessor creates a block processor
-func MakeProcessor(genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block, datadir string, handler func(block *ledgercore.ValidatedBlock) error) (processor.Processor, error) {
+func MakeProcessor(genesis *bookkeeping.Genesis, genesisBlock *bookkeeping.Block, dbRound uint64, datadir string, handler func(block *ledgercore.ValidatedBlock) error) (processor.Processor, error) {
 	initState, err := util.CreateInitState(genesis, genesisBlock)
 	if err != nil {
 		return nil, fmt.Errorf("MakeProcessor() err: %w", err)
 	}
-	l, err := ledger.OpenLedger(logging.NewLogger(), filepath.Join(path.Dir(datadir), "ledger"), false, initState, algodConfig.GetDefaultLocal())
+	if dbRound != 0 && !ledgerExists(datadir, prefix) {
+		msg := fmt.Sprintf("%s\n%s\n%s\n%s\n",
```
AlgoStephenAkiki, line 51: Is the added friction of causing an error here worth it?  Why not just create the files in sequential migration mode and print a warning to the user saying that there are faster alternatives?

winder (approved):
I think we can merge this in and have a followup handling the catchpoint > next-round case

cmd/algorand-indexer/daemon.go
```diff
@@ -231,6 +234,19 @@ var daemonCmd = &cobra.Command{
 				genesisReader := importer.GetGenesisFile(genesisJSONPath, bot.Algod(), logger)
 				_, err := importer.EnsureInitialImport(db, genesisReader, logger)
 				maybeFail(err, "importer.EnsureInitialImport() error")
+
+				// sync local ledger
+				nextDBRound, err := db.GetNextRoundToAccount()
+				maybeFail(err, "Error getting DB round")
+				if nextDBRound > 0 {
+					if catchpoint != "" {
```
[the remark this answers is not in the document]
shiqizng: we have a check for this in processor, https://github.com/algorand/indexer/blob/d057e1c0ab061f01fc9549c97685327938686cbd/processor/blockprocessor/block_processor.go#L52-L54. maybe also add one in the migration method?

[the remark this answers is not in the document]
winder: This is the opposite check, we need to avoid catching up if the catchpoint if ahead of the desired round (i.e. if you're starting a new node and provide a catchpoint, we shouldn't initialize anything)

PR #1039, opened 2022-06-16
Refactor daemon, add more tests

Eric-Warehime:
## Summary

First in a series which aims to clean up our commands into something more testable and adds coverage.

* Removed the `makeOptions` call in `api_config.go` because it was only being used for a single argument, and I've refactored it to take a config.
* Moved the variables in `daemon.go` to a struct, `daemonConfig`
* Added create functions for `daemonConfig`, and `daemonCmd`
* Moved the bulk of the daemon logic into its own function, `runDaemon`. In the future we can separate this out further into more modular testable pieces, for now I've left it as one giant runner.
* I've begun the process of removing `panic` and `os.Exit` calls from this except in the base cobra command. By only passing errors around it becomes more easily testable, we don't have to worry about defer funcs not running which is the reason we have the panic logic, and we are still able to properly exit non-zero when the error reaches the command's `Run` function.
* `makeOptions` has been refactored to convert between a `daemonConfig` and `api.ExtraOptions`.
* Added `BindFlagSet` which is a copy of `BindFlags` since it makes sense to pass the `FlagSet` around instead of the command, but I'm not yet able to fully convert the existing function without more changes.

## Test Plan

Added various tests to get started on testing--mostly for config variables/params for now. It shows how the new setup can be used to simplify tests and validate specific scenarios.


winder (commented):
This is a great start. RIP globals.

cmd/algorand-indexer/daemon_test.go
```diff
@@ -102,3 +107,85 @@ func TestReadGenesis(t *testing.T) {
...
+}
+
+// Make sure we output and return an error when both an API Config and
+// enable all parameters are provided together.
+func TestConfigWithEnableAllParamsExpectError(t *testing.T) {
+	indexerDataDir := createTempDir(t)
+	defer os.RemoveAll(indexerDataDir)
+	daemonConfig := newDaemonConfig()
+	cmd := newDaemonCmd(daemonConfig)
+	daemonConfig.flags = cmd.Flags()
```
winder, line 126: This seems a little weird, I guess it's simulating some implementation detail in the Execute function? I'm not sure we see a lot of benefit from testing at the Cmd level vs calling runDaemon directly.

Eric-Warehime: Yeah it just needs to be there because `runDaemon` tries to access the flags so we need a non null for `daemonCmd.flags`. I've removed it by setting the variable directly.

cmd/algorand-indexer/daemon.go
```diff
@@ -61,253 +66,299 @@ var (
...
-	daemonCmd.Flags().Uint32VarP(&maxApplicationsLimit, "max-applications-limit", "", 1000, "set the maximum allowed Limit parameter for querying applications")
-	daemonCmd.Flags().Uint32VarP(&defaultApplicationsLimit, "default-applications-limit", "", 100, "set the default Limit parameter for querying applications, if none is provided")
-
-	daemonCmd.Flags().StringVarP(&indexerDataDir, "data-dir", "i", "", "path to indexer data dir, or $INDEXER_DATA")
-
-	daemonCmd.Flags().StringVarP(&cpuProfile, "cpuprofile", "", "", "file to record cpu profile to")
-	daemonCmd.Flags().StringVarP(&pidFilePath, "pidfile", "", "", "file to write daemon's process id to")
-	daemonCmd.Flags().StringVarP(&configFile, "configfile", "c", "", "file path to configuration file (indexer.yml)")
+	daemonConfig := newDaemonConfig()
+	daemonCmd = newDaemonCmd(daemonConfig)
```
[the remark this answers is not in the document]
Eric-Warehime: I want to move everything away from being called in init functions. So I've updated this to just have the root command directly create the config and cmd. Let me know what you think.

config/config.go
```diff
@@ -40,3 +40,21 @@ func BindFlags(cmd *cobra.Command) {
 		}
 	})
 }
+
+func BindFlagSet(flags *pflag.FlagSet) {
```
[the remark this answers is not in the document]
Eric-Warehime: Yeah, I added a commit which does this--it was not too difficult.

cmd/algorand-indexer/daemon.go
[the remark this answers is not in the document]
winder: My only concern is over-correcting and adding too much abstraction. We're using cobra, no need to hide it.

This is the pattern I've adopted lately: https://github.com/algorand/indexer/blob/develop/cmd/block-generator/runner/runner.go
Since the command exists already, the consumer just needs to execute it: https://github.com/algorand/indexer/blob/develop/cmd/block-generator/main.go

Its worked pretty nicely:
https://github.com/algorand/node-ui/blob/master/cmd/tui/main.go#L43
https://github.com/algorand/go-algorand/blob/master/cmd/algokey/keyreg.go#L60

That said, I'm happy to learn a new trick if you have a better idea.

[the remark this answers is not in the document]
Eric-Warehime: Yeah it doesn't matter too much to me--the pattern you're showing just uses the init function to construct the commands whereas I've put it in separate functions. 

I did have a feeling that relying on init functions seemed fragile to me so I investigated...and it can be.

The problem is that we have a command hierarchy--root command which adds subcommands. So when we run the `rootCmd.AddCommand` function we need to be sure that the subcommand we're adding is initialized.
If we use a variable such as this to define a command:  
`cmd/algorand-indexer/z.go`
```
package main

var var1 string
var zCmd := &cobra.Command{
	Use:   "z",
	Short: "foo",
	Long:  "bar",
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Println(var1)
	},
}

func init() {
    importCmd.Flags().StringVarP(&var1, "var1", "", "", "sample flag")
}
```
The above will work perfectly when we call `rootCmd.AddCommand(zCmd)` in `main.go`'s init function because the `zCmd` variable is "global/static" in the package, it will always be initialized before any init functions are called.

However, if we have a top level command variable defined which is initialized in the init function, such as this:
`cmd/algorand-indexer/z.go`
```
var zCmd *cobra.Command

var (
	var1 string
)

func init() {
	zCmd = &cobra.Command{
		Use:   "z",
		Short: "foo",
		Long:  "bar",
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println(var1)
		},
	}
	zCmd.Flags().StringVarP(&var1, "var1", "", "", "sample")
}
```
The above program will compile, but when you run the command you will get a seg fault:
```
./cmd/algorand-indexer/algorand-indexer z
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104dc16bc]

goroutine 1 [running]:
github.com/spf13/cobra.(*Command).AddCommand(0x10686dda0, {0x1400051fd40, 0x1, 0x1})
	/Users/eric/.asdf/installs/golang/1.17.9/packages/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:1149 +0x5c
main.init.3()
	/Users/eric/go/src/github.com/algorand/indexer/cmd/algorand-indexer/main.go:121 +0x404
```
This is because init functions within the same package are run in lexicographical order. It just so happens that all of the commands we currently have come before `main.go`, but `z.go` is run afterwards so the variable is never initialized.

The reason this does not happen in the `block-generator` example above is that init functions in imports are run depth first, and the root command for block-generator is in a different package than the subcommands which are then imported here https://github.com/algorand/indexer/blob/develop/cmd/block-generator/core/commands.go#L7 causing their init functions to run before the block generator's `commands.go` init function.


Sorry for how long this post became 😸 , I just got really interested in this after you brought it up.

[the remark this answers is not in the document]
Eric-Warehime: So I think the best solution is one of:
* separate subcommands into their own packages
* use a top-level variable to initialize commands outside the init functions
* use dedicated constructor functions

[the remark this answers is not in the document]
winder: Sounds good, thanks for the explanation.

cmd/algorand-indexer/daemon.go
```diff
@@ -65,276 +67,338 @@ var (
...
+		if err != nil {
+			logger.WithError(err).Errorf("%v", err)
+			return err
+		}
+		defer configs.Close()
+		err = viper.ReadConfig(configs)
+		if err != nil {
+			logger.WithError(err).Errorf("invalid config file (%s): %v", viper.ConfigFileUsed(), err)
+			return err
+		}
```
winder, line 127: nit: We solved this slightly differently today on `develop` (also, I think this is opening the wrong file)
```suggestion
		// No config file supplied via command line, auto-load it
		configFile = filepath.Join(indexerDataDir, autoLoadIndexerConfigName)
	}
```

cmd/algorand-indexer/daemon.go
```diff
@@ -65,276 +67,338 @@ var (
...
-			configs, err := os.Open(configFile)
-			if err != nil {
-				maybeFail(err, "%v", err)
-			}
-			defer configs.Close()
-			err = viper.ReadConfig(configs)
-			if err != nil {
-				maybeFail(err, "invalid config file (%s): %v", viper.ConfigFileUsed(), err)
-			}
+func loadIndexerParamConfig(cfg *daemonConfig) error {
```
winder, line 146: These helpers don't need the `daemonConfig`, you can pass in the specific value they need.

cmd/algorand-indexer/daemon.go
```diff
@@ -65,276 +67,338 @@ var (
...
+	}
+	return err
+}
 
-		}
+func runDaemon(daemonConfig *daemonConfig) error {
+	var err error
+	config.BindFlagSet(daemonConfig.flags)
+	err = configureLogger()
+	if err != nil {
```
winder, line 199: You can avoid reusing / shadowing the err object by using this pattern more
```suggestion
	if err := configureLogger(); err != nil {
```

cmd/algorand-indexer/daemon.go
```diff
@@ -65,276 +67,338 @@ var (
...
-				maybeFail(err, "%v", err)
-			}
-			defer configs.Close()
-			err = viper.ReadConfig(configs)
+	// Load the Parameter config
+	if err = loadIndexerParamConfig(daemonConfig); err != nil {
+		return err
+	}
+
+	if daemonConfig.pidFilePath != "" {
```
winder, line 219: You can return a closer function from `createIndexerPidFile` to move more of this logic into the helper function.
```
if daemonConfig.pidFilePath != "" {
    closer, err := createIndexerPidFile(daemonConfig.pidFilePath)
    if err != nil {
        return err
    }
    defer closer()
}
```

Just having fun now... I think something extra fancy could work (but now err has to be defined):
```
var closer func()
var err error
if closer, err = createIndexerPidFile(daemonConfig.pidFilePath); err != nil {
    return err
}
defer closer()
```

cmd/algorand-indexer/daemon_test.go
```diff
@@ -102,3 +109,174 @@ func TestReadGenesis(t *testing.T) {
 	assert.Contains(t, err.Error(), "readGenesis() err: EOF")
 
 }
+
+func createTempDir(t *testing.T) string {
+	dir, err := os.MkdirTemp("", "indexer")
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
```
winder, line 117:
```suggestion
	require.NoError(t, err)
```

cmd/algorand-indexer/daemon_test.go
```diff
@@ -102,3 +109,174 @@ func TestReadGenesis(t *testing.T) {
...
+	daemonConfig := &daemonConfig{}
+	daemonConfig.flags = pflag.NewFlagSet("indexer", 0)
+	daemonConfig.indexerDataDir = indexerDataDir
+	daemonConfig.enableAllParameters = true
+	daemonConfig.suppliedAPIConfigFile = "foobar"
+	err := runDaemon(daemonConfig)
+	errorStr := "not allowed to supply an api config file and enable all parameters"
+	if err.Error() != errorStr {
+		t.Fatalf("expected error %s, but got %s", errorStr, err.Error())
+	}
```
winder, line 135: The error validation helpers in testify used to be pretty basic, I think there are some new ones that might help with this sort of check.

Also, theres a panic here if `runDaemon` doesn't return the expected error.
```suggestion
	require.Error(t, err)    
	require.Equals(t, err.Error(), errorStr)
```

cmd/algorand-indexer/daemon_test.go
```diff
@@ -102,3 +109,174 @@ func TestReadGenesis(t *testing.T) {
...
+	daemonConfig := &daemonConfig{}
+	daemonConfig.flags = pflag.NewFlagSet("indexer", 0)
+	daemonConfig.indexerDataDir = indexerDataDir
+	daemonConfig.configFile = tempConfigFile
+	err := runDaemon(daemonConfig)
+	// This error string is probably OS-specific
+	errorStr := "no such file or directory"
+	if !strings.Contains(err.Error(), errorStr) {
+		t.Fatalf("expected error %s, but got %s", errorStr, err.Error())
+	}
```
winder, line 151: You get the idea, same thing for the rest of these.
```suggestion
	require.Error(t, err)
	require.Contains(t, err.Error(), errorStr)
```

cmd/algorand-indexer/daemon_test.go
```diff
@@ -102,3 +109,174 @@ func TestReadGenesis(t *testing.T) {
...
+	defer os.RemoveAll(indexerDataDir)
+	tempConfigFile := indexerDataDir + "/indexer-alt.yml"
+	os.WriteFile(tempConfigFile, []byte(";;;"), fs.ModePerm)
+	daemonConfig := &daemonConfig{}
+	daemonConfig.flags = pflag.NewFlagSet("indexer", 0)
+	daemonConfig.indexerDataDir = indexerDataDir
+	daemonConfig.configFile = tempConfigFile
+	logger.SetOutput(b)
+	// Should assert this is an error even if it's not one we directly control (ours are wrapped)
+	_ = runDaemon(daemonConfig)
```
winder, line 166:
```suggestion
	err := runDaemon(daemonConfig)
	require.Error(t, err)
```

cmd/algorand-indexer/main.go
```diff
@@ -115,6 +115,7 @@ func init() {
 
 	rootCmd.AddCommand(importCmd)
 	importCmd.Hidden = true
+	daemonCfg.setFlags(daemonCmd.Flags())
```
winder, line 118: I think daemonCfg should be initialized in a function which also sets this and defines the flags. We keep this code in close proximity across all our apps, so I think its important to do here as well.

cmd/algorand-indexer/daemon_test.go
```diff
@@ -1,9 +1,16 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"fmt"
+	"github.com/spf13/pflag"
```
winder, line 8: Import order. It isn't super strict, but we usually have 4 groups:
1. builtins
2. external deps
3. algorand external deps
4. project deps

winder (changes requested):
Just a few suggestions. Overall this is really great.

cmd/algorand-indexer/daemon.go
```diff
@@ -65,276 +67,338 @@ var (
...
 	Short: "run indexer daemon",
 	Long:  "run indexer daemon. Serve api on HTTP.",
+	//Args:
 	Run: func(cmd *cobra.Command, args []string) {
-		var err error
-		config.BindFlags(cmd)
-		err = configureLogger()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to configure logger: %v", err)
+		daemonCfg.flags = cmd.Flags()
```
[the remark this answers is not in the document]
shiqizng: I'm thinking replacing `setFlags` with something like, 
```suggestion
		func daemonCmdFlags(cmd *cobra.Command) {
  cmd.Flags().StringVarP(&algodDataDir, "algod", "d", "", "path to algod data dir, or $ALGORAND_DATA")
}
```
then in init()
```suggestion
		daemonCmdFlags(daemonCmd)
```

shiqizng (commented):
This is great! It looks like we should be able to write integration tests for daemon after this change.

cmd/algorand-indexer/daemon.go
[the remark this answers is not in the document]
Eric-Warehime: Ended up moving all of it to a single function as Will suggested.

cmd/algorand-indexer/daemon.go
Eric-Warehime: I tried to match what was merged in from develop. In the most recent update. Let me know if the logic is not correct.

cmd/algorand-indexer/daemon.go
Eric-Warehime: Fixed--this function in particular stores the resolved variable in the config struct so I've left that one as is.

cmd/algorand-indexer/daemon_test.go
Eric-Warehime: Fixed all the tests to use `require`

cmd/algorand-indexer/main.go
Eric-Warehime: Put everything in a single function in the most recent changes.

cmd/algorand-indexer/daemon.go
```diff
@@ -60,265 +64,336 @@ var (
...
 
-		if configFile != "" {
-			configs, err := os.Open(configFile)
-			if err != nil {
-				maybeFail(err, "error with config file (%s): %v", configFile, err)
-			}
-			defer configs.Close()
-			err = viper.ReadConfig(configs)
+	// Detect the various auto-loading configs from data directory
+	if err = loadIndexerConfig(daemonConfig.indexerDataDir, daemonConfig.configFile); err != nil {
```
winder, line 246: @AlgoStephenAkiki if we move this ahead of the `BindFlagSet` call, can we avoid calling that function twice?

winder (approved):
Looks good, just want Stephen to double check the config file loading.

cmd/algorand-indexer/daemon.go
winder: The routine you added looks good.

cmd/algorand-indexer/daemon_test.go
```diff
@@ -75,3 +82,154 @@ func TestImportRetryAndCancel(t *testing.T) {
...
+	autoloadPath := filepath.Join(indexerDataDir, autoLoadIndexerConfigName)
+	os.WriteFile(autoloadPath, []byte{}, fs.ModePerm)
+	daemonConfig := &daemonConfig{}
+	daemonConfig.flags = pflag.NewFlagSet("indexer", 0)
+	daemonConfig.indexerDataDir = indexerDataDir
+	daemonConfig.enableAllParameters = true
+	daemonConfig.suppliedAPIConfigFile = "foobar"
+	err := runDaemon(daemonConfig)
+	errorStr := "not allowed to supply an api config file and enable all parameters"
+	require.EqualError(t, err, errorStr)
```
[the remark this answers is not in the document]
Eric-Warehime: Updated everything to use assert

cmd/algorand-indexer/daemon.go
```diff
@@ -60,265 +64,336 @@ var (
...
+		logger.WithError(err).Errorf("File Does Not Exist Error: %v", err)
+		return err
+	}
+	defer configs.Close()
+	err = viper.ReadConfig(configs)
+	if err != nil {
+		logger.WithError(err).Errorf("invalid config file (%s): %v", viper.ConfigFileUsed(), err)
+		return err
+	}
+	logger.Infof("Using configuration file: %s\n", resolvedConfigPath)
```
AlgoStephenAkiki, line 178: After this log statement we should call BindSetFlags() to make sure we get everything set.

cmd/algorand-indexer/daemon.go
AlgoStephenAkiki: I don't think so since the problem is that we have a circular dependency on the logger inside the `loadIndexerConfig` function (the logger needs to be configured from configuration and the function uses the logger but also can re-load the configuration).

AlgoStephenAkiki (approved):

PR #1041, opened 2022-06-20
Support for custom consensus on private networks

scholtz:
## Summary

Algorand indexer at the moment does not allow using custom consensus on private networks. This pull request adds this feature.

## Test Plan

I tested it manually on custom made private network with custom consensus.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -93,7 +93,7 @@ var daemonCmd = &cobra.Command{
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Unable to load optional consensus protocols file: %s/consensus.json %v\n", indexerDataDir, err)
 			}else{
-				fmt.Printf("consensus loaded\n");
+				fmt.Printf("consensus loaded\n")
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


Eric-Warehime:
Hi, I'm not sure if I can see what this would do since block data in the indexer is imported independent of the consensus process. Can you give me a bit of background on why you might need this change?

scholtz:
hi @Eric-Warehime 

sure, algod loads correctly custom consensus from consensus.json file.

but there is no way to use custom consensus with the indexer.. 

the indexer without this bugfix fails here https://github.com/algorand/indexer/blob/5be3c0e17e455b037a05c941b305a4378b420d55/importer/importer.go#L27 with error **protocol 'custom-protocol' not found**

if you want, you can find the way to load consensus from the algod server, but right now you are using hardcoded consensus values with no option to extend it.. this PR is simple fix to resolve the issue with ability to provide consensus.json file with the indexer

Eric-Warehime (changes requested):
Looks good to me. Just need to fix the fmt issue

cmd/algorand-indexer/daemon.go
```diff
@@ -84,6 +87,15 @@ var daemonCmd = &cobra.Command{
 		// Detect the various auto-loading configs from data directory
 		indexerConfigFound := util.FileExists(filepath.Join(indexerDataDir, autoLoadIndexerConfigName))
 		paramConfigFound := util.FileExists(filepath.Join(indexerDataDir, autoLoadParameterConfigName))
+		consensusConfigFound := util.FileExists(filepath.Join(indexerDataDir, "consensus.json"))
+		if consensusConfigFound {
+			err = goconfig.LoadConfigurableConsensusProtocols(indexerDataDir)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Unable to load optional consensus protocols file: %s/consensus.json %v\n", indexerDataDir, err)
+			}else{
```
Eric-Warehime, line 95:
```suggestion
			} else {
```

Eric-Warehime (changes requested):

PR #1042, opened 2022-06-21
local ledger fixes

shiqizng:
## Summary
This PR includes these changes

- ledger file path. No need to parse the directory from the path
- If ledger and db are already in sync, skip migration
- if catchpoint round > dbRound, skip fast catchup

## Test Plan
manual testing and regression testing

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/rpcs"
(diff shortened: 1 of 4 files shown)


cmd/algorand-indexer/daemon.go
```diff
@@ -240,8 +241,17 @@ var daemonCmd = &cobra.Command{
...
 				if nextDBRound > 0 {
 					if catchpoint != "" {
-						err = localledger.RunMigrationFastCatchup(logging.NewLogger(), catchpoint, &opts)
-						maybeFail(err, "Error running ledger migration in fast catchup mode")
+						round, _, err := ledgercore.ParseCatchpointLabel(catchpoint)
+						if err != nil {
+							maybeFail(err, "catchpoint error")
+						}
+						if uint64(round) >= nextDBRound {
+							logger.Warnf("round for given catchpoint is ahead of db round. skip fast catchup")
```
Eric-Warehime, line 249: Nit: They technically could be equal here since it's not strictly greater than.

migrations/local_ledger/migration.go
```diff
@@ -61,6 +61,10 @@ func RunMigrationSimple(round uint64, opts *idb.IndexerDbOptions) error {
 	if err != nil {
 		return fmt.Errorf("RunMigration() err: %w", err)
 	}
+	// ledger and db states are in sync
+	if proc.NextRoundToProcess()-1 == round {
+		return nil
```
Eric-Warehime, line 66: Nit: You can just return err here since it should be nil if we haven't errored out already.

Eric-Warehime (approved):

winder (approved):

PR #1058, opened 2022-06-23
Ledger init status

shiqizng:
## Summary

This PR adds status logging for ledger init and stops node if it's running when indexer receives SIGTERM or SIGINT. 

## Test Plan
regression testing, run daemon.

### DIFF
--- a/processor/blockprocessor/initialize.go
+++ b/processor/blockprocessor/initialize.go
@@ -34,9 +34,7 @@ func InitializeLedgerSimple(logger *log.Logger, round uint64, opts *idb.IndexerD
 		signal.Notify(cancelCh, syscall.SIGTERM, syscall.SIGINT)
 		go func() {
 			<-cancelCh
-			logger.Errorf("Ledger migration interrupted")
-			// exit process if migration is interrupted so that
-			// migration state doesn't get updated in db
+			logger.Errorf("Ledger initalization aborted")
 			os.Exit(1)
 		}()
 	}
@@ -95,6 +93,20 @@ func fullNodeCatchup(logger *log.Logger, round basics.Round, catchpoint, dataDir
 	}
 	// remove node directory after when exiting fast catchup mode
 	defer os.RemoveAll(filepath.Join(dataDir, genesis.ID()))
+	// stop node if indexer exits during ledger init
+	_, cf := context.WithCancel(context.Background())
+	defer cf()
+	{
+		cancelCh := make(chan os.Signal, 1)
+		signal.Notify(cancelCh, syscall.SIGTERM, syscall.SIGINT)
+		go func() {
+			<-cancelCh
+			logger.Errorf("Ledger initalization aborted")
+			node.Stop()
+			os.Exit(1)
+		}()
+	}
+
 	node.Start()
 	time.Sleep(5 * time.Second)
 	logger.Info("algod node running")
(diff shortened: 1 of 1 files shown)


processor/blockprocessor/initialize.go
```diff
@@ -95,6 +93,20 @@ func fullNodeCatchup(logger *log.Logger, round basics.Round, catchpoint, dataDir
 	}
 	// remove node directory after when exiting fast catchup mode
 	defer os.RemoveAll(filepath.Join(dataDir, genesis.ID()))
+	// stop node if indexer exits during ledger init
```
winder, line 96: There is a context in daemon.go that we should use

AlgoStephenAkiki (approved):

processor/blockprocessor/block_processor.go
```diff
@@ -47,19 +49,19 @@ func MakeProcessorWithLedgerInit(logger *log.Logger, catchpoint string, genesis
...
 
 		}
-		err := InitializeLedgerSimple(logger, nextDBRound-1, &opts)
+		err := InitializeLedgerSimple(ctx, logger, nextDBRound-1, &opts)
 		if err != nil {
 			return &blockProcessor{}, fmt.Errorf("MakeProcessorWithCatchup() slow catchup err: %w", err)
 		}
 	}
-	return MakeProcessor(logger, genesis, nextDBRound, opts.AlgodDataDir, handler)
+	return MakeProcessor(logger, genesis, nextDBRound, opts.IndexerDatadir, handler)
```
shiqizng, line 64: I think it should be the indexer directory?

processor/blockprocessor/initialize.go
```diff
@@ -95,6 +93,20 @@ func fullNodeCatchup(logger *log.Logger, round basics.Round, catchpoint, dataDir
...
+	{
+		cancelCh := make(chan os.Signal, 1)
+		signal.Notify(cancelCh, syscall.SIGTERM, syscall.SIGINT)
+		go func() {
+			<-cancelCh
+			logger.Errorf("Ledger initalization aborted")
+			node.Stop()
+			os.Exit(1)
+		}()
+	}
```
winder, line 108: Lets sync up next week, I can tell you about how the context works

processor/blockprocessor/initialize.go
```diff
@@ -105,18 +117,20 @@ func fullNodeCatchup(logger *log.Logger, round basics.Round, catchpoint, dataDir
 	for status.LastRound < round {
 		time.Sleep(2 * time.Second)
```
winder, line 118: Generally, it lets you do something like this to terminate early:
```suggestion
		select {
			case <- ctx.Done():
				return
			case <- time.After(2 * time.Second):
		}
```

winder: We would also want to use `defer node.Stop()` in case we exit early.

processor/blockprocessor/initialize.go
```diff
@@ -26,20 +24,7 @@ import (
...
-		signal.Notify(cancelCh, syscall.SIGTERM, syscall.SIGINT)
-		go func() {
-			<-cancelCh
-			logger.Errorf("Ledger migration interrupted")
-			// exit process if migration is interrupted so that
-			// migration state doesn't get updated in db
-			os.Exit(1)
-		}()
-	}
+func InitializeLedgerSimple(ctx context.Context, cf context.CancelFunc, logger *log.Logger, round uint64, opts *idb.IndexerDbOptions) error {
```
winder, line 27: Is something missing here? How does this exit when it's all caught up?

processor/blockprocessor/initialize_test.go
```diff
@@ -63,8 +64,9 @@ func TestRunMigration(t *testing.T) {
 		AlgodToken:     "AAAAA",
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
 	// migrate 3 rounds
-	err = InitializeLedgerSimple(logrus.New(), 3, &opts)
+	err = InitializeLedgerSimple(ctx, cancel, logrus.New(), 3, &opts)
```
winder, line 69: Unless you are adding a test that it terminates when you call `cancel()`, these can use `context.Background()`.

processor/blockprocessor/initialize.go
shiqizng: cancel() is called here https://github.com/algorand/indexer/blob/eb2ac877849bf9841e69bd61bd9c5603267bac3a/processor/blockprocessor/initialize.go#L162-L165

processor/blockprocessor/initialize_test.go
shiqizng: this method takes a cancel function which is why I used context.WithCancel.

processor/blockprocessor/initialize.go
winder: Thanks, I missed that.

I think this would cause problems because the context is going to be cancelled when we're finished initializing. You can wrap the context with another `WithCancel`. I had to try it, but this seems to work as expected: https://go.dev/play/p/l1S-LqG0MkR

shiqizng: fix added.

processor/blockprocessor/initialize.go
```diff
@@ -26,21 +24,9 @@ import (
 )
 
 // InitializeLedgerSimple executes the migration core functionality.
-func InitializeLedgerSimple(logger *log.Logger, round uint64, opts *idb.IndexerDbOptions) error {
-	ctx, cf := context.WithCancel(context.Background())
+func InitializeLedgerSimple(ctx context.Context, logger *log.Logger, round uint64, opts *idb.IndexerDbOptions) error {
```
winder, line 27: Is this function supposed to block until initialization is complete? It looks like it starts fetcher and then runs in the background.

winder (approved):

PR #1063, opened 2022-06-24
Block fetcher: catchup without algod

algoganesh:
## Summary
This implementation of fetcher directly fetches blocks from the network instead of fetching from an algod server.

`./algorand-indexer daemon -P <postgres credentials> --catchup <path to genesis.json> -i <indexer datadir>`
starts indexer and fetches blocks directly from the network

## Test Plan
`TestDirectCatchupService` in `fetcher_test.go` implements the logic and fetches first 10 blocks from the `mainnet` network.

### DIFF
--- a/fetcher/catchupservice.go
+++ b/fetcher/catchupservice.go
@@ -3,6 +3,9 @@ package fetcher
 import (
 	"context"
 	"errors"
+	"fmt"
+	"sync"
+	"time"
 
 	"github.com/algorand/go-algorand/agreement"
 	"github.com/algorand/go-algorand/config"
@@ -10,6 +13,8 @@ import (
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/network"
+	"github.com/algorand/go-algorand/rpcs"
+	"github.com/algorand/indexer/util/metrics"
 	"github.com/labstack/gommon/log"
 )
 
@@ -33,6 +38,8 @@ type nodeInfo struct {
 	ctx context.Context
 }
 
+type task func() basics.Round
+
 func (n nodeInfo) IsParticipating() bool { return false }
 
 // MakeCatchupService creats a catchup service and initialzes a gossipnode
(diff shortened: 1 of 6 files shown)


fetcher/catchupservice.go
```diff
@@ -0,0 +1,269 @@
...
+func (n nodeInfo) IsParticipating() bool { return false }
+
+// MakeCatchupService creats a catchup service and initialzes a gossipnode
+func MakeCatchupService(ctx context.Context, genesis bookkeeping.Genesis) (s *CatchupService) {
+	s = &CatchupService{}
+
+	s.cfg = config.AutogenLocal
+	s.genesis = genesis
+	s.log = logging.NewLogger()
+	nodeinfo := makeNodeInfo(ctx)
```
cce, line 52: this can be nil

fetcher/catchupservice.go
```diff
@@ -0,0 +1,269 @@
...
+	s.net, _ = network.NewWebsocketNetwork(s.log, s.cfg, nil, genesis.ID(), genesis.Network, nodeinfo)
```
cce, line 53: making a websocket network and connecting to relays means by default, you are going to ask all those relays to send you transactions, votes, and proposals it sees via the websocket gossip connection. But since all you care about is the /v1/{network}/block/N HTTP endpoint, you can get away with not doing any of the websocket stuff... 

To filter out gossip messages there is a "MessagesOfInterest" map that you can set to be empty so that the relays don't try to send you any gossip messages.

algoganesh: `MessagesOfInterest` map is set during `setup()` through `RegisterMessageInterest` if `relayMessages` boolean is set to true (nominally true for relays, false otherwise). `config.AutogenLocal` leads to `relayMessages` being false, so the default value of map is already empty.

cce: the default set that peers will send you is actually `defaultSendMessageTags` unless you deregister your interest. Otherwise you are going to be sent gossip messages (like votes, transactions, etc) that you'll drop anyway (because you don't have any services calling wn.RegisterHandlers).

PR #1067, opened 2022-06-27
New Feature: Indexer Application Box Ingestion

tzaffi:
# Adding Box Information to Indexer

## Summary

The proposed `app_box` postgres table with all the insertion/update/deletion logic in place.

Additional test recommendations welcome!

## Related PRs 
* go-algorand
  * [Original PR #4182](https://github.com/algorand/go-algorand/pull/4182)
  * Merged into [feature/avm-box](https://github.com/algorand/go-algorand/pull/4149)
* indexer
  * Follow up box search #1117  

### TODO (move this to the integration branch once merged)

- [x] Incorporate the prunings of #1146 when it has been merged and figure out what to do with `idb/postgres/postgres_rand_test.go`
- [x] Manual test of network that creates 125 apps with 1000 boxes and verify directly on IndexerDB that `app_box` is populated as expected
- [x] Ingest [go-algorand/data/basics/userBalance.go::AcccountData](https://github.com/algorand/go-algorand/blob/d601bb1415103eaa7a3ecbe95bfaa23a06e7f6a2/data/basics/userBalance.go#L233)'s `TotalBoxes` and `TotalBoxBytes` into `account.account_data` JSONB
- [x] Validate the ingestion of the summary data in the previous step in all relevant tests
- [x] Unit tests
- [x] Integration
- [ ] End-to-end test
  - [x] is passing against network that includes boxes
  - [ ] has been re-pointed to `rel-nightly` once that job includes boxes (cf. `misc/docker-compose.yml`)
- [ ] Re-point the `third_party/go-algorand` submodule to 
  - [x] `go-algorand` 
  - [ ] a released commit containing boxes
- [ ] Has go-algorand's [feature/avm-box](https://github.com/algorand/go-algorand/pull/4149) been merged?

## An Aside about Minimum Balance Calculation and App-Boxes

CF: The [previous](https://gist.github.com/tzaffi/593648f015ce89294e47ccb64be0282f) (partially wrong) version of this aside.

It would be nice to solve the long-standing issue #808 and provide minimum balance information for accounts, _including_ app-accounts. App-Boxes affect the min balance calculation for an app-account holding boxes, and as this PR introduces their basic information to indexer. 

Here's a brief description of how to add min-balance for apps in indexer:

### The New Minimum Balance Calculation

The new [boxes-included min-balance calculation](https://github.com/jannotti/go-algorand/blob/b8d54d12c22d286e8d12dd2f898cdf8b4332f79e/data/basics/userBalance.go#L519-L525) requires knowing:
* the number of boxes for an app (`totalBoxes`)
* the total storage utilized in those boxes (`totalBoxBytes`) which is comprised of [bytes used for keys and values](https://github.com/jannotti/go-algorand/blob/b8d54d12c22d286e8d12dd2f898cdf8b4332f79e/ledger/internal/applications.go#L234)
The current PR allows calculating this information for a particular app. 

#### Here's some pseudo-SQL:

```SQL
WITH box_stats AS (
    SELECT account_data->>'tbx' AS total_boxes, account_data->>'tbxb' AS total_box_bytes 
    FROM account
    WHERE addr = {$appIndex.Addres()}
) 
SELECT {$BoxFlatMinbalance} * total_boxes + {$BoxByteMinbalance} * total_box_bytes AS app_box_cost
FROM box_stats;
```


Makefile
```diff
@@ -94,3 +101,18 @@ indexer-v-algod-swagger:
 indexer-v-algod: nightly-setup indexer-v-algod-swagger nightly-teardown
 
 .PHONY: test e2e integration fmt lint deploy sign test-package package fakepackage cmd/algorand-indexer/algorand-indexer idb/mocks/IndexerDb.go go-algorand indexer-v-algod
+
+# TEMPORARY RECIPE. DO NOT MERGE.
```
tzaffi, line 105: Revert Makefile before merging

idb/postgres/postgres_rand_test.go
```diff
@@ -33,8 +33,6 @@ func generateAccountData() ledgercore.AccountData {
 		return ledgercore.AccountData{}
 	}
 
-	const numCreatables = 20
```
tzaffi, line 36: compilation warning

processor/blockprocessor/block_processor.go
```diff
@@ -133,9 +133,10 @@ func (proc *blockProcessor) Process(blockCert *rpcs.EncodedBlockCert) error {
 		vb = ledgercore.MakeValidatedBlock(block, delta)
 	}
 
+	// TODO: DO NOT MERGE WITH THIS DEBUGGING CODE IN PLACE
```
tzaffi, line 136: yep, don't merge this way

processor/blockprocessor/block_processor.go
```diff
@@ -194,19 +195,17 @@ func (proc *blockProcessor) NextRoundToProcess() uint64 {
...
-		Accounts:         nil,
-		Resources:        nil,
-		Creators:         make(map[ledger.Creatable]ledger.FoundAddress),
-		DeprecatedBoxes:  deprecatedBoxes,
-		DeprecatedBoxes2: deprecatedBoxes2,
-		TouchedBoxes:     make(map[string]struct{}), // TODO: we can know what the expected size is. Should we provide this as well?
+		Accounts:               nil,
+		Resources:              nil,
+		Creators:               make(map[ledger.Creatable]ledger.FoundAddress),
+		DeprecatedTouchedBoxes: make(map[string]struct{}),
```
tzaffi, line 208: don't merge with `Deprecated...`

Makefile
```diff
@@ -26,10 +26,17 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
...
-go-algorand:
-	git submodule update --init && cd third_party/go-algorand && \
+# go-algorand:
+# 	git submodule update --init && cd third_party/go-algorand && \
+# 		make crypto/libs/`scripts/ostype.sh`/`scripts/archtype.sh`/lib/libsodium.a
+
+
+# TEMPORARY CHANGE - REVERT BEFORE MERGING
+go-algorand: go-avm-box
+	cd third_party/go-algorand && \
```
tzaffi, line 36:
```suggestion
go-algorand:
	git submodule update --init && cd third_party/go-algorand && \
		make crypto/libs/`scripts/ostype.sh`/`scripts/archtype.sh`/lib/libsodium.a

```

Makefile
```diff
@@ -94,3 +101,18 @@ indexer-v-algod-swagger:
...
+
+LOCAL_GO_ALGORAND = "/Users/zeph/github/tzaffi/go-algorand"
+# "/Users/zeph/github/cce/go-algorand"
+zymlocal:
+	mv third_party/go-algorand third_party/zo-algorand
+	ln -s $(LOCAL_GO_ALGORAND) third_party/go-algorand
+
+dezymlocal:
+	# rm third_party/go-algorand
+	mv third_party/zo-algorand third_party/go-algorand
```
tzaffi, line 118:
```suggestion
```

Makefile
```diff
@@ -30,6 +30,9 @@ go-algorand:
 	git submodule update --init && cd third_party/go-algorand && \
 		make crypto/libs/`scripts/ostype.sh`/`scripts/archtype.sh`/lib/libsodium.a
 
+		make crypto/libs/`scripts/ostype.sh`/`scripts/archtype.sh`/lib/libsodium.a
+
+
```
tzaffi, line 35:
```suggestion
```

accounting/eval_preload.go
```diff
@@ -9,7 +9,8 @@ import (
 
 // Add requests for asset and app creators to `assetsReq` and `appsReq` for the given
 // transaction.
-func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD, assetsReq map[basics.AssetIndex]struct{}, appsReq map[basics.AppIndex]struct{}) {
+func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD,
+	assetsReq map[basics.AssetIndex]struct{}, appsReq map[basics.AppIndex]struct{}) {
```
tzaffi, line 13:
```suggestion
func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD, assetsReq map[basics.AssetIndex]struct{}, appsReq map[basics.AppIndex]struct{}) {
```

accounting/eval_preload.go
```diff
@@ -60,7 +61,8 @@ func MakePreloadCreatorsRequest(payset transactions.Payset) (map[basics.AssetInd
 
 // Add requests for account data and account resources to `addressesReq` and
 // `resourcesReq` respectively for the given transaction.
-func addToAccountsResourcesRequest(stxnad *transactions.SignedTxnWithAD, assetCreators map[basics.AssetIndex]ledger.FoundAddress, appCreators map[basics.AppIndex]ledger.FoundAddress, addressesReq map[basics.Address]struct{}, resourcesReq map[basics.Address]map[ledger.Creatable]struct{}) {
+func addToAccountsResourcesRequest(stxnad *transactions.SignedTxnWithAD, assetCreators map[basics.AssetIndex]ledger.FoundAddress, appCreators map[basics.AppIndex]ledger.FoundAddress,
+	addressesReq map[basics.Address]struct{}, resourcesReq map[basics.Address]map[ledger.Creatable]struct{}) {
```
tzaffi, line 65:
```suggestion
func addToAccountsResourcesRequest(stxnad *transactions.SignedTxnWithAD, assetCreators map[basics.AssetIndex]ledger.FoundAddress, appCreators map[basics.AppIndex]ledger.FoundAddress, addressesReq map[basics.Address]struct{}, resourcesReq map[basics.Address]map[ledger.Creatable]struct{}) {
```

idb/postgres/internal/writer/util.go
```diff
@@ -8,7 +8,6 @@ import (
 type copyFromChannelStruct struct {
 	ch   chan []interface{}
 	next []interface{}
-	err  error
```
tzaffi, line 11: compilation warning about unused field

processor/blockprocessor/block_processor.go
```diff
@@ -195,14 +233,19 @@ func prepareEvalResources(l *indexerledger.LedgerForEvaluator, block *bookkeepin
 }
 
 // Preload asset and app creators.
-func prepareCreators(l *indexerledger.LedgerForEvaluator, payset transactions.Payset) (map[basics.AssetIndex]ledger.FoundAddress, map[basics.AppIndex]ledger.FoundAddress, error) {
+func prepareCreators(lfe *indexerledger.LedgerForEvaluator, payset transactions.Payset) (
+	map[basics.AssetIndex]ledger.FoundAddress,
+	map[basics.AppIndex]ledger.FoundAddress,
+	error) {
```
tzaffi, line 239:
```suggestion
func prepareCreators(lfe *indexerledger.LedgerForEvaluator, payset transactions.Payset) (map[basics.AssetIndex]ledger.FoundAddress, map[basics.AppIndex]ledger.FoundAddress, error) {
```

processor/blockprocessor/block_processor.go
```diff
@@ -2,6 +2,7 @@ package blockprocessor
 
 import (
 	"fmt"
+
```
tzaffi, line 5: TODO: Curerently it appears that there are no changes of substance in this file. Confirm, and remove this change before merging.

processor/blockprocessor/block_processor_test.go
```diff
@@ -56,9 +56,9 @@ func TestFailedProcess(t *testing.T) {
 	require.NoError(t, err)
 	defer l.Close()
 	// invalid processor
-	pr, err := block_processor.MakeProcessorWithLedger(nil, nil)
```
tzaffi, line 59: compilation warnings

processor/blockprocessor/initialize_test.go
```diff
@@ -55,7 +55,7 @@ func TestRunMigration(t *testing.T) {
 	httpmock.RegisterResponder("GET", `=~^http://localhost/v2/status/wait-for-block-after/\d+\z`,
 		httpmock.NewStringResponder(200, string(json.Encode(algod.Status{}))))
 
-	dname, err := os.MkdirTemp("", "indexer")
```
tzaffi, line 58: compiler warnings

processor/eval/ledger_for_evaluator.go
```diff
@@ -139,6 +139,16 @@ func (l LedgerForEvaluator) GetAppCreator(indices map[basics.AppIndex]struct{})
 	return res, nil
 }
 
+// LookupKv is part of go-algorand's indexerLedgerForEval interface.
+func (l LedgerForEvaluator) LookupKv(round basics.Round, key string) (*string, error) {
+	// a simple pass-thru to the go-algorand ledger
+	val, err := l.Ledger.LookupKv(round, key)
+
+	fmt.Printf("\nindexer::LedgerForEvaluator.LookupKv(round=%d, key=%s) --> (%v, %v)", round, key, val, err)
```
tzaffi, line 147: TODO: don't merge with this print statement

processor/eval/ledger_for_evaluator_test.go
```diff
@@ -75,7 +75,7 @@ func TestLedgerForEvaluatorAccountDataMissingAccount(t *testing.T) {
 	defer ld.Close()
 
 	var addr basics.Address
-	_, err := rand.Read(addr[:])
+	rand.Read(addr[:])
```
tzaffi, line 78: compiler warnings

idb/postgres/internal/schema/setup_postgres.sql
```diff
@@ -117,3 +117,13 @@ CREATE TABLE IF NOT EXISTS account_app (
 
 -- For looking up existing app local states by account
 CREATE INDEX IF NOT EXISTS account_app_by_addr_partial ON account_app(addr) WHERE NOT deleted;
+
+-- For lookup up app box storage
```
tzaffi, line 121:
```suggestion
-- For looking up app box storage
```

idb/postgres/internal/schema/setup_postgres.sql
```diff
@@ -117,3 +117,13 @@ CREATE TABLE IF NOT EXISTS account_app (
 
 -- For looking up existing app local states by account
 CREATE INDEX IF NOT EXISTS account_app_by_addr_partial ON account_app(addr) WHERE NOT deleted;
+
+-- For looking up app box storage
+CREATE TABLE IF NOT EXISTS app_box (
+  app bigint NOT NULL,
+  name bytea NOT NULL,
+  size smallint NOT NULL,     -- _NEGATIVE_ of box'es size
```
tzaffi, line 125: I'm considering removing this column. If we never care about deleted box sizes, then there is no need for it. If we do, there are alternatives. Here's a copy paste of a recent [slack thread](https://algorand-internal.slack.com/archives/C020LLY8VBM/p1656447698751459?thread_ts=1656435858.591749&cid=C020LLY8VBM):

```
[John Jannotti]
size does not seem needed though, as I don’t expect queries based on size, and the size is derivable from the value itself, which is always full length.

[Zeph Grunschlag]
Regarding size
Suppose we want to be able to answer the following:
does app X have any deleted boxes?
If so, what are their names, and what were their sizes?
and in a sequence of create-replace-delete-create-replace-delete.. using the same name, we only care about the last existing box
If we use the approach that NULL value indicates deleted, then we can no longer query the length of the value for size.
An alternate approach (more space efficient unless there are MANY deleted boxes), is to add another bool column deleted and not set value <-- NULL when deleting.
If we don't care at all to know which boxes have been deleted, then the schema gets simplified and no size column is necessary
Or if we want to know about which boxes were deleted, but don't care about their size while existing, we can use the approach of NULL'ing the value with no size column.
```

processor/blockprocessor/block_processor.go
```diff
@@ -168,6 +207,7 @@ func prepareEvalResources(l *indexerledger.LedgerForEvaluator, block *bookkeepin
 		Accounts:  nil,
 		Resources: nil,
 		Creators:  make(map[ledger.Creatable]ledger.FoundAddress),
+		// DeprecatedTouchedBoxes: make(map[string]struct{}),
```
tzaffi, line 210:
```suggestion
```

idb/postgres/internal/writer/writer.go
```diff
@@ -243,6 +254,12 @@ func writeAssetResource(round basics.Round, resource *ledgercore.AssetResourceRe
 }
 
 func writeAppResource(round basics.Round, resource *ledgercore.AppResourceRecord, batch *pgx.Batch) {
+	// type AppResourceRecord struct {
+	// 	Aidx   basics.AppIndex
+	// 	Addr   basics.Address
+	// 	Params AppParamsDelta
+	// 	State  AppLocalStateDelta
+	// }
```
tzaffi, line 262:
```suggestion
```

idb/postgres/internal/writer/writer.go
```diff
@@ -290,6 +301,33 @@ func writeAccountDeltas(round basics.Round, accountDeltas *ledgercore.AccountDel
...
 	}
+
+}
+
+func writeBoxMods(round basics.Round, kvMods map[string]*string, batch *pgx.Batch) error {
+	// kvMods can in theory support more general storage types than app boxes.
+	// However, here we assume that all the given kvMods represent app boxes.
+
+	// Update `app_box`
+	fmt.Printf("writeAccountDeltas - app_box portion for round=%d", round)
```
tzaffi, line 312: TODO: remove print statement before merging

idb/postgres/internal/writer/writer.go
```diff
@@ -339,6 +377,13 @@ func (w *Writer) AddBlock(block *bookkeeping.Block, modifiedTxns []transactions.
 		}
 		writeAccountDeltas(block.Round(), &delta.Accts, sigTypeDeltas, &batch)
 	}
+	{
+		// TODO: don't think I actually need this code block
```
tzaffi, line 381: ^^^ figure out ^^^

idb/sig_type.go
```diff
@@ -59,5 +59,5 @@ func SignatureType(stxn *transactions.SignedTxn) (SigType, error) {
 		return Lsig, nil
 	}
 
-	return "", fmt.Errorf("unable to determine the signature type")
+	return "", fmt.Errorf("unable to determine the signature type for stxn=%+v", stxn)
```
tzaffi, line 62: @winder - should I remove this very verbose addition? This is what I added following our [recent slack thread](https://algorand-internal.slack.com/archives/C028R5W4DA6/p1656452142169899?thread_ts=1656419400.435169&cid=C028R5W4DA6). But maybe it's actually useful to keep for the rare occasion that this would occur.

processor/blockprocessor/block_processor.go
```diff
@@ -195,14 +234,16 @@ func prepareEvalResources(l *indexerledger.LedgerForEvaluator, block *bookkeepin
 }
 
 // Preload asset and app creators.
-func prepareCreators(l *indexerledger.LedgerForEvaluator, payset transactions.Payset) (map[basics.AssetIndex]ledger.FoundAddress, map[basics.AppIndex]ledger.FoundAddress, error) {
+func prepareCreators(lfe *indexerledger.LedgerForEvaluator, payset transactions.Payset) (map[basics.AssetIndex]ledger.FoundAddress, map[basics.AppIndex]ledger.FoundAddress, error) {
 	assetsReq, appsReq := accounting.MakePreloadCreatorsRequest(payset)
-
-	assets, err := l.GetAssetCreator(assetsReq)
+	assets, err := lfe.GetAssetCreator(assetsReq)
+	if err != nil {
```
tzaffi, line 240: I didn't see a good reason not to percolate `err` up for `GetAssetCreator()`. Is there a good reason this wasn't previously done?

tzaffi, line 240: Percolating `err` up for `GetAssetCreator()` in line with other calls here. Any objections?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "go-algorand"]
 	path = third_party/go-algorand
-	url = https://github.com/algorand/go-algorand
+	url = https://github.com/tzaffi/go-algorand
```
tzaffi, line 3: DO NOT MERGE AS IS!!!!

idb/postgres/internal/writer/writer.go
```diff
@@ -104,6 +107,12 @@ var statements = map[string]string{
 		(addr, app, localstate, deleted, created_at, closed_at)
 		VALUES($1, $2, 'null'::jsonb, TRUE, $3, $3) ON CONFLICT (addr, app) DO UPDATE SET
 		localstate = EXCLUDED.localstate, deleted = TRUE, closed_at = EXCLUDED.closed_at`,
+	upsertAppBoxStmtName: `INSERT INTO app_box AS ab
+		(app, name, value, created_at)
+		VALUES ($1, $2, $3, $4)
+		ON CONFLICT (app, name) DO UPDATE SET
+		value = EXCLUDED.value`,
+	deleteAppBoxStmtName: `DELETE FROM app_box WHERE app = $1 and name = $2`,
```
tzaffi, line 115: This is the only example I can see where we _actually_ delete a row, instead of marking it deleted. @winder - are you ok with this approach?

idb/postgres/internal/writer/writer.go
```diff
@@ -290,6 +299,30 @@ func writeAccountDeltas(round basics.Round, accountDeltas *ledgercore.AccountDel
...
+	// However, here we assume that all the provided kvMods represent app boxes.
+
+	fmt.Printf("writeAccountDeltas - app_box portion for kvMods of len=%d", len(kvMods))
+
+	for key, value := range kvMods {
+		app, name, err := logic.SplitBoxKey(key)
+		if err != nil {
+			return fmt.Errorf("writeBoxMods() err: %w", err)
+		}
+		// TODO: we need solid tests to verify no further encoding of bytes `box_name` and `value` is necessary
```
tzaffi, line 317: yes we do!

jannotti:
A comment on your discussion of min-balance: TotalBoxBytes is calculated as the sum of the lengths of the values _and_ keys.

idb/postgres/internal/writer/writer.go
```diff
@@ -309,12 +342,12 @@ func (w *Writer) AddBlock0(block *bookkeeping.Block) error {
 		_, err := results.Exec()
 		if err != nil {
 			results.Close()
-			return fmt.Errorf("AddBlock() exec err: %w", err)
+			return fmt.Errorf("AddBlock() exec err for block 0: %w", err)
```
jannotti, line 345: To ease grepping, and make it clear what function this is actually in:
```suggestion
			return fmt.Errorf("AddBlock0() exec err: %w", err)
```

processor/blockprocessor/block_processor.go
```diff
@@ -195,14 +233,16 @@ func prepareEvalResources(l *indexerledger.LedgerForEvaluator, block *bookkeepin
...
+		return nil, nil, fmt.Errorf("prepareCreators() err: %w", err)
+	}
+	apps, err := lfe.GetAppCreator(appsReq)
 	if err != nil {
 		return nil, nil, fmt.Errorf("prepareCreators() err: %w", err)
 	}
-	apps, err := l.GetAppCreator(appsReq)
 	if err != nil {
 		return nil, nil, fmt.Errorf("prepareCreators() err: %w", err)
 	}
```
jannotti, line 248: Am I just seeing a weird diff artifact, or is this checking the exact same thing twice on l:243 and then l:246?

processor/blockprocessor/initialize.go
```diff
@@ -175,7 +175,7 @@ func blockHandler(logger *log.Logger, dbRound uint64, proc processor.Processor,
 			case <-ctx.Done():
 				return err
 			case <-time.After(retryDelay):
-				break
+				// noop
```
jannotti, line 178: If this change is right (it looks like it is, but what do I know?), it seems important enough to do in another PR now, rather than wait for boxes.

processor/eval/ledger_for_evaluator.go
jannotti: And, IMO, just `return l.Ledger.LookupKv(round, key)`

idb/postgres/internal/writer/writer.go
jannotti: Hard for me to parse your internal monologue.  Do you mean you're trying to decide if you need the indentation and extra scope? Or whether you need the code at all? It _seems_ like this is the heart of the matter here - the code is needed - so I think you're debating whether to introduce the scope?  If so, I'd say do it, since the same technique was used above.  (Or remove both if this the only place this is done) No point in change style mid-stream.

I assume it was done to isolate the introduction of this `err` variable?

processor/blockprocessor/initialize.go
tzaffi: I made a [suggestion](https://github.com/algorand/indexer/pull/1085#pullrequestreview-1026432861) for the main localledger pr

processor/blockprocessor/block_processor.go
tzaffi: Thanks for noticing. I thought this was an artifact, but a quick check of the actual source confirms that this is truly duplicative!

idb/postgres/internal/writer/writer.go
```diff
@@ -290,6 +299,30 @@ func writeAccountDeltas(round basics.Round, accountDeltas *ledgercore.AccountDel
...
+
+func writeBoxMods(kvMods map[string]*string, batch *pgx.Batch) error {
+	// Insert/Update/Delete ON `app_box`
+	// kvMods can in theory support more general storage types than app boxes.
+	// However, here we assume that all the provided kvMods represent app boxes.
+
+	fmt.Printf("writeAccountDeltas - app_box portion for kvMods of len=%d", len(kvMods))
+
+	for key, value := range kvMods {
+		app, name, err := logic.SplitBoxKey(key)
```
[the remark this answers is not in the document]
tzaffi: Upon further reflection, I'd like to keep this immediate exit as it is. If in the future, a new kind of kvstore element is introduced to `kvMods` the indexer will get stuck ingesting this round. I verified by returning an error intentionally inside the for-loop. When running my simulated migration, I got stuck at round 9 with the following error:

```
err: AddBlock() err on boxes: writeBoxMods() - this is a temporary error!!!!","level":"error","msg":"block 9 import failed","time":"2022-07-01T12:35:32-05:00"
```

In other words, by leaving this as is we're insisting that we know what to do with non-boxes _before_ we enter this method, or modify it to handle correctly. @winder what do you think?  In terms of conventions for this repo, would you rather fail in the future if a new kind of kvstore element is introduced, or just skip silently?

processor/eval/ledger_for_evaluator.go
```diff
@@ -139,6 +139,16 @@ func (l LedgerForEvaluator) GetAppCreator(indices map[basics.AppIndex]struct{})
 	return res, nil
 }
 
+// LookupKv is part of go-algorand's indexerLedgerForEval interface.
```
Eric-Warehime, line 142: This doesn't seem to be true. 
https://github.com/algorand/go-algorand/blob/master/ledger/evalindexer.go#L36

I guess we could add it, but I'm curious as to why you all have chosen not to include Boxes as [Creatables](https://github.com/algorand/go-algorand/blob/77cb33289d0cf02239db75c32cc2b3cfebf65354/data/basics/userBalance.go#L317)? Just from a naive perspective I would think Boxes should fit into that sort of model and so could follow similar lookup paths as Assets and Apps.

tzaffi: You're correct that it's not true on master. However, I've added `LookupKv()` to the interface in the [companion PR](https://github.com/tzaffi/go-algorand/blob/zbox2/ledger/evalindexer.go#L46).

Boxes are associated to apps while Assets and Apps are associated with accounts. Since every app has its own account (basically, the hash of its index), one could in theory associate boxes as well to accounts directly. However, in the future, we may want to allow other sorts of key-value storage in the same `kvstore` ledger table so its more flexible to allow arbitrary keys for lookup, not just apps or accounts (e.g. maybe assets could have specialized boxes too in some future AVM). Creatables can never be re-created once they are deleted (yes, you can re-create an app with the same logic, but it has a new serial appid so it never truly re-creates the original app). At the end of the day, boxes can _outlive_ the app that they belonged to. This is a strange artifact that results in wasted space and min-balance allocation, but it is more evidence that boxes are fundamentally different from Creatables.

idb/postgres/internal/encoding/encoding_test.go
```diff
@@ -599,3 +599,8 @@ func TestLcAccountDataEncoding(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, ad, decodedAd)
 }
+
+// TODO: test encoding of box data
```
tzaffi, line 603: might not make sense to write this test... implement or revert

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
```diff
@@ -46,7 +46,7 @@ func TestLedgerForEvaluatorLatestBlockHdr(t *testing.T) {
 	require.NoError(t, err)
 	defer tx.Rollback(context.Background())
 
-	l, err := ledger_for_evaluator.MakeLedgerForEvaluator(tx, basics.Round(2))
+	l, err := ledger_for_evaluator.MakeDeprecatedLedgerForEvaluator(tx, basics.Round(2))
```
tzaffi, line 49: ^^^

processor/blockprocessor/initialize.go
tzaffi: Suggestion was taken. Resolving.

processor/eval/ledger_for_evaluator.go
Eric-Warehime: Ok thanks for explaining. In terms of how Boxes are going to work this makes sense then. 

Conceptually it seems like the more logical way to register a Box would be to just associate it with an account and let application logic determine which of your account's Apps pull data from which of your account's Boxes, but that is clearly outside the scope of the changes here.

tzaffi: As you suggest, it's logical to associate a box with an account. This reminds me of the issue I raised in the PR description regarding **Min Balance**. If we want to be able to support this feature for accounts and for app-accounts, and we want to avoid a future backfill, we may already in this PR or a follow-up before release, want to add an `app_account` column to `app_box` together with an index. @winder - appreciate your feedback as well.

idb/postgres/postgres.go
```diff
@@ -177,77 +176,77 @@ func (db *IndexerDb) init(opts idb.IndexerDbOptions) (chan struct{}, error) {
 	return db.runAvailableMigrations(opts)
 }
 
-// Preload asset and app creators.
```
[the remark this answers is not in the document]
tzaffi: https://github.com/algorand/indexer/pull/1095 has id covered

idb/postgres/internal/writer/writer.go
```diff
@@ -290,6 +299,31 @@ func writeAccountDeltas(round basics.Round, accountDeltas *ledgercore.AccountDel
...
+
+func writeBoxMods(kvMods map[string]*string, batch *pgx.Batch) error {
+	// INSERT INTO / UPDATE / DELETE FROM `app_box`
+	// WARNING: kvMods can in theory support more general storage types than app boxes.
+	// However, here we assume that all the provided kvMods represent app boxes.
+	// If a non-box is encountered inside kvMods, an error will be returned and
+	// AddBlock() will fail with the migration getting stuck at the corresponding round.
+
+	fmt.Printf("writeAccountDeltas - app_box portion for kvMods of len=%d", len(kvMods))
+
```
tzaffi, line 313:
```suggestion
```

tzaffi (commented):
batched commits

processor/eval/ledger_for_evaluator.go
tzaffi: resolving

idb/postgres/internal/schema/setup_postgres.sql
tzaffi: @shiqizng - I'm re-opening the resolved conversation with @jannotti . I found the argument to rename the original column `box_name` to `name` compelling since the table is `box_app` and so `box_app.name` seems pretty clear. However, Lamprey ought to set the naming conventions IMO. Let me know.

idb/postgres/internal/schema/setup_postgres.sql
```diff
@@ -117,3 +117,11 @@ CREATE TABLE IF NOT EXISTS account_app (
 
 -- For looking up existing app local states by account
 CREATE INDEX IF NOT EXISTS account_app_by_addr_partial ON account_app(addr) WHERE NOT deleted;
+
+-- For looking up app box storage
+CREATE TABLE IF NOT EXISTS app_box (
+  app bigint NOT NULL,
+  name bytea NOT NULL,
```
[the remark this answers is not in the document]
tzaffi: Resolving, since I moved this to another conversation.

idb/postgres/postgres.go
```diff
@@ -177,78 +176,6 @@ func (db *IndexerDb) init(opts idb.IndexerDbOptions) (chan struct{}, error) {
...
-func prepareCreators(l *ledger_for_evaluator.LedgerForEvaluator, payset transactions.Payset) (map[basics.AssetIndex]ledger.FoundAddress, map[basics.AppIndex]ledger.FoundAddress, error) {
```
tzaffi, line 181: These are being deleted by #1095 
**DON'T MERGE AHEAD OF THAT PR AND MAKE SURE THERE ARE NO DIFFS HERE**

idb/postgres/postgres.go
```diff
@@ -2592,7 +2519,7 @@ func (db *IndexerDb) GetAccountState(addressesReq map[basics.Address]struct{}, r
 	}
 	defer tx.Rollback(context.Background())
 
-	l, err := ledger_for_evaluator.MakeLedgerForEvaluator(tx, basics.Round(0))
+	l, err := ledger_for_evaluator.MakeDeprecatedLedgerForEvaluator(tx, basics.Round(0))
```
tzaffi, line 2522: Similar comment. Don't merge!

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
shiqizng: I prefer failing in the future if a new kind of kvstore element is introduced.

idb/postgres/postgres_integration_test.go
```diff
@@ -1549,7 +1550,7 @@ func TestAddBlockCreateDeleteAppSameRound(t *testing.T) {
 	defer l.Close()
 
 	appid := uint64(1)
-	createTxn := test.MakeCreateAppTxn(test.AccountA)
+	createTxn := test.MakeCreateSimpleAppTxn(test.AccountA)
```
tzaffi, line 1553: renaming because a slightly more complex function is introduced rendering the original as "Simple"

processor/eval/ledger_for_evaluator_test.go
```diff
@@ -101,8 +101,7 @@ func TestLedgerForEvaluatorAsset(t *testing.T) {
 	err = pr.Process(&rawBlock)
 	assert.Nil(t, err)
 
-	ld := indxLeder.MakeLedgerForEvaluator(l)
-	require.NoError(t, err)
```
tzaffi, line 105: this `err` and similar throughout this file, is being asserted 3 lines above. Hence, I'm removing this redundant assertion.

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator.go
```diff
@@ -44,24 +44,24 @@ var statements = map[string]string{
 		schema.AccountTotals + `'`,
 }
 
-// LedgerForEvaluator implements the indexerLedgerForEval interface from
+// DeprecatedLedgerForEvaluator implements the indexerLedgerForEval interface from
```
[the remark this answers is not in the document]
tzaffi: Do not merge as is

idb/postgres/internal/ledger_for_evaluator/ledger_for_evaluator_test.go
tzaffi: don't merge as is

processor/eval/ledger_for_evaluator_test.go
```diff
@@ -539,11 +531,219 @@ func TestLedgerForEvaluatorAppCreatorMultiple(t *testing.T) {
...
+
+	rawBlock := rpcs.EncodedBlockCert{Block: block, Certificate: agreement.Certificate{}}
+	err = pr.Process(&rawBlock)
+	require.NoError(t, err)
+
+	ret, err := ld.LookupKv(currentRound, "sanity check")
+	require.NoError(t, err)
+	require.Nil(t, ret) // nothing found isn't considered an error
+
+	// block header handoff: round 1 --> round 3
```
tzaffi, line 567:
```suggestion
	// block header handoff: round 1 --> round 2
```

api/handlers_e2e_test.go
```diff
@@ -236,7 +236,7 @@ func TestAccountExcludeParameters(t *testing.T) {
 
 	const expectedAppIdx = 1 // must be 1 since this is the first txn
 	const expectedAssetIdx = 2
-	createAppTxn := test.MakeCreateAppTxn(test.AccountA)
+	createAppTxn := test.MakeCreateSimpleAppTxn(test.AccountA)
```
tzaffi, line 239: I can understand push back for renaming this. Could have done just as well to keep the old name.

algochoi (commented):
Looking nice

idb/postgres/internal/schema/setup_postgres_sql.go
```diff
@@ -121,4 +121,12 @@ CREATE TABLE IF NOT EXISTS account_app (
 
 -- For looking up existing app local states by account
 CREATE INDEX IF NOT EXISTS account_app_by_addr_partial ON account_app(addr) WHERE NOT deleted;
+
+-- For looking up app box storage
+CREATE TABLE IF NOT EXISTS app_box (
```
shiqizng, line 126: I believe you also need to add a migration step in postgres_migrations.go to create this new table in existing indexers.

processor/eval/ledger_for_evaluator_test.go
```diff
@@ -551,11 +543,220 @@ func TestLedgerForEvaluatorAppCreatorMultiple(t *testing.T) {
 	}
 }
 
+// Test the functionality of `func (l LedgerForEvaluator) LookupKv()`.
+// This is done by handing off a pointer to Struct `processor/eval/ledger_for_evaluator.go::LedgerForEvaluator`
+// in the `CompareAppBoxesAgainstLedger()` assertions function which then asserts using `LookupKv()`
+func TestLedgerForEvaluatorLookupKv(t *testing.T) {
```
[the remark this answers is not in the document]
tzaffi: Yes. While the overall scenarios are identical, this one checks that the local-ledger is getting the expected info about boxes, while the `postgres_intgration_test.go` actually queries indexer's DB after the box information ought to have been ingested to verify that everything works. You can see this because we have:
* `CompareAppBoxesAgainstLedger()` used to validate here
* `CompareAppBoxesAgainstDB()` used to validate in `postgres_integration_test.go`

There might be a way to do everything from the integration test (and in some sense, the integration is "good nuf" already because it subsumes this test), but since we have a lot of examples of testing separate components of indexer separately, and I believe that too many tests is much preferred over too little, I'd like to keep this test.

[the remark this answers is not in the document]
tzaffi: I guess it's also possible to refactor in a way that runs the same logic but receives function pointer for the `CompareXYZ()` function. But that seems like a bit of overkill...

idb/postgres/postgres_integration_test.go
```diff
@@ -2251,3 +2252,219 @@ func TestIndexerDb_GetAccounts(t *testing.T) {
...
+		"disappointing box":         "RUG PULL!!!!",
+		"don't box me in this way":  "non box-conforming",
+		"I will be assimilated":     "THE BORG",
+		"I'm destined for deletion": "I'm still alive!!!",
+		"box #8":                    "eight is beautiful",
+	}
+
+	boxTxns = make([]*transactions.SignedTxnWithAD, 0)
+	expectedAppBoxes[appid] = make(map[string]string)
+	for boxName, valPrefix := range appBoxesToSet {
```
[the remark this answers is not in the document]
tzaffi: There were too many subtle differences between this function and other similar ones. So I'm not convinced of the worthwhileness.

util/test/account_testutil.go
```diff
@@ -288,9 +319,40 @@ func MakeAppCallTxn(appid uint64, sender basics.Address) transactions.SignedTxnW
 	}
 }
 
+// MakeAppCallTxnWithBoxes makes an appl transaction with a NoOp upon completion.
+func MakeAppCallTxnWithBoxes(appid uint64, sender basics.Address, appArgs []string, boxNames []string) transactions.SignedTxnWithAD {
```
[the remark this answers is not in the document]
tzaffi: For the purposes of how this method is used, it was a bit easier to provide strings and have them get translated to bytes. I don't feel especially strong about it though.

idb/postgres/postgres_migrations.go
```diff
@@ -19,6 +19,14 @@ import (
 	"github.com/algorand/indexer/idb/postgres/internal/types"
 )
 
+// AppBoxMigration is used to interpolate the field AppBoxMigration of `cmd/texttosource/main.go` in the SQL template `idb/postgres/internal/schema/setup_postgres.sql`
+const AppBoxMigration = `CREATE TABLE IF NOT EXISTS app_box (
```
tzaffi, line 23: @shiqizng - in order to avoid duplicated `CREATE TABLE` logic, I'm storing the SQL in `AppBoxMigration` and populating it both in the migration step, as well as in the SQL generation script using a a template.

idb/postgres/postgres_random_box_test.go
```diff
@@ -0,0 +1,225 @@
+package postgres
```
tzaffi, line 1: I separated this out from `postgres_random_test.go` because of all the merge issues. But I can unify those scripts again. LMK

idb/postgres/internal/schema/setup_postgres.sql
```diff
@@ -117,3 +117,6 @@ CREATE TABLE IF NOT EXISTS account_app (
 
 -- For looking up existing app local states by account
 CREATE INDEX IF NOT EXISTS account_app_by_addr_partial ON account_app(addr) WHERE NOT deleted;
+
+-- For looking up app box storage
+{{.AppBox}};
```
shiqizng, line 122: this has been used as a source file to generate the string in setup_postgres_sql.go. With your template approach, you're already creating the sql string in AppBoxMigration. This makes me wonder why we need the sql file at all.  Maybe texttosource generator should be refactored so each table is assigned to an exported variable so that `const SetupPostgresSql=fmt.Sprintf("%s\n%s\n%s", Table1,Table2,Table3)`
@winder

idb/postgres/postgres_migrations.go
winder: We deprecate the migrations pretty aggressively, only guaranteeing a 2 week grace period. So duplicate code here is not really a problem.

Let's keep the simple .sql file without template processing, and we'll be able to remove the second definition in the N + 1 release.

[Here is the policy if you are curious for more details.](https://docs.google.com/document/d/1ZQjx_gAO9uhqSy25m00puPHxe6CmQ8y-RVkamOTTDbg/edit#heading=h.djearkgog5ci)

tzaffi: I'll revert to the original approach (copy/pasta the CREATE TABLE). Meanwhile, I'd love to be able to read the policy and have requested access.

tzaffi: This has been reverted @winder

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "go-algorand"]
 	path = third_party/go-algorand
-	url = https://github.com/algorand/go-algorand
+	url = https://github.com/jannotti/go-algorand
```
tzaffi, line 3: ^^^^ don't merge until revert ^^^

misc/Dockerfile
```diff
@@ -23,4 +23,4 @@ RUN pip3 install -r misc/requirements.txt
 
 ENV INDEXER_DATA="${HOME}/indexer/"
 # Run test script
-ENTRYPOINT ["/bin/bash", "-c", "sleep 5 && python3 misc/e2elive.py --connection-string \"$CONNECTION_STRING\" --indexer-bin /opt/go/indexer/cmd/algorand-indexer/algorand-indexer --indexer-port 9890"]
+ENTRYPOINT ["/bin/bash", "-c", "sleep 5 && python3 misc/e2elive.py --connection-string \"$CONNECTION_STRING\" --s3-source-net \"$CI_E2E_FILENAME\" --indexer-bin /opt/go/indexer/cmd/algorand-indexer/algorand-indexer --indexer-port 9890"]
```
tzaffi, line 26: Using same naming convention `CI_E2E_FILENAME ` for the S3 network artifact as in [go-algorand](https://github.com/algorand/go-algorand/blob/cdd4aa1305fd3d99c83c7a629f2202ffe0f3964e/test/scripts/e2e.sh#L184).

misc/docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
       dockerfile: ./misc/Dockerfile    
     environment:
       CONNECTION_STRING: "host=e2e-db port=5432 user=algorand password=algorand dbname=indexer_db sslmode=disable"
+      CI_E2E_FILENAME: "tzaffi-box-e2e-improvements"
```
tzaffi, line 10: ^^^^ TODO: revert `CI_E2E_FILENAME` to `"net_done"` before merging

shiqizng (approved):

idb/postgres/internal/schema/setup_postgres.sql
winder, line 124: Are these names expected to be human readable?

idb/postgres/internal/writer/writer.go
winder: Based on the earlier discussions about whether or not we care to keep historic boxes around, this is fine by me.

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
winder: I agree with Shiqi here. We need to fail whenever we come across something unknown because we only process blocks one time. Migrations to repair the database after the fact can be tricky, so we avoid processing partial blocks as a policy.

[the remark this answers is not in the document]
jannotti: There are things that happen on the block chain that we don't store a permanent record of because we decided not to. App state, for example.  If we ever wanted to add the feature "What was the app state at round R", we would have to re-index.  Similarly, if we add some random new feature that uses the kvstore, we may or may not wish to support indexing it. I don't think the indexer should crash in such a situation, it should keep going. That _forces_  us to do work in indexer even if we have decided we don't care about this new unknown thing.

util/test/box_testutil.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	var delta ledgercore.StateDelta
+	for acct, acctData := range accts {
+		delta.Accts.Upsert(acct, *acctData)
+	}
+	return delta, kvUpdated, boxTotals
+}
+
+// CompareAppBoxesAgainstLedger uses LedgerForEvaluator to assert that provided app boxes can be retrieved as expected
```
winder, line 112: Unused?

util/test/box_testutil.go
```diff
@@ -0,0 +1,146 @@
...
+	acctData.TotalBoxBytes -= prevBoxBytes
+
+	boxTotals[addr] = basics.AccountData{
+		TotalBoxes:    acctData.TotalBoxes,
+		TotalBoxBytes: acctData.TotalBoxBytes,
+	}
+}
+
+// BuildAccountDeltasFromKvsAndMods simulates keeping track of the evolution of the box statistics
+func BuildAccountDeltasFromKvsAndMods(t *testing.T, kvOriginals, kvMods map[string]*string) (
```
winder, line 61: I slightly prefer putting this in `writer_test.go`, or a `utils_test.go` file in the `writer` package. It's possible for files that don't end in `_test.go` to slip into a production binary, so unless they're really used across packages it is best to avoid using real packages.

[This is a long time pet peeve for me.](https://github.com/golang/go/issues/44086)

idb/postgres/internal/writer/writer.go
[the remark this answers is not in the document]
jannotti: I yield to you guys - it's your code.

idb/postgres/internal/writer/writer_test.go
```diff
@@ -1564,3 +1565,169 @@ func TestWriterAddBlock0(t *testing.T) {
...
+	*/
+
+	db, _, shutdownFunc := pgtest.SetupPostgresWithSchema(t)
+	defer shutdownFunc()
+
+	var block bookkeeping.Block
+	block.BlockHeader.Round = basics.Round(1)
+	delta := ledgercore.StateDelta{}
+
+	blockAdder := func(tx pgx.Tx) error {
```
winder, line 1586: nit: passing block/delta into this function would have made the test code a little clearer for me. Or maybe rename it to something like "addNewBlock", my brain read the function as "block address".

idb/postgres/internal/writer/writer_test.go
```diff
@@ -1564,3 +1565,169 @@ func TestWriterAddBlock0(t *testing.T) {
 		assert.Equal(t, expected, accounts)
 	}
 }
+
+// Simulate a scenario where app boxes are created, mutated and deleted in consecutive rounds.
+func TestWriterAppBoxTableInsertMutateDelete(t *testing.T) {
```
winder, line 1570: The only test I see missing here is adding another box after the first set, i.e. inserting box 6 during the second round.

util/test/account_testutil.go
```diff
@@ -211,6 +213,45 @@ func MakeCreateAppTxn(sender basics.Address) transactions.SignedTxnWithAD {
 	}
 }
 
+// MakeComplexCreateAppTxn makes a transaction that creates an arbitrary app
+func MakeComplexCreateAppTxn(sender basics.Address, approval, clear string, assemblerVersion ...uint64) (transactions.SignedTxnWithAD, error) {
```
winder, line 217: The vararg parameter to sneak in an optional parameter is confusing. I would prefer a comment saying "set assemblerVersion to 0 unless you want to override", or if you want to cleaner parameters, a second Make function which only has 3 parameters.

idb/postgres/internal/schema/setup_postgres.sql
tzaffi: No. They can be any bytestring whatsoever. We're not sure what the usage pattern will be, but I personally expect many of the names to be human readable, if not the values.

idb/postgres/internal/encoding/encoding_test.go
```diff
@@ -592,10 +595,55 @@ func TestLcAccountDataEncoding(t *testing.T) {
...
-	expectedString := `{"onl":1,"sel":"DwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","spend":"BgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","stprf":"EwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==","tapl":11,"tapp":10,"tas":13,"tasp":12,"teap":9,"tsch":{"nbs":8,"nui":7},"vote":"DgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","voteFst":16,"voteKD":18,"voteLst":17}`
+	expectedString := `{"onl":1,"sel":"DwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","spend":"BgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","stprf":"EwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==","tapl":11,"tapp":10,"tas":13,"tasp":12,"tbx":20,"tbxb":21,"teap":9,"tsch":{"nbs":8,"nui":7},"vote":"DgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","voteFst":16,"voteKD":18,"voteLst":17}`
 	assert.Equal(t, expectedString, string(buf))
 
 	decodedAd, err := DecodeTrimmedLcAccountData(buf)
 	require.NoError(t, err)
 	assert.Equal(t, ad, decodedAd)
 }
+
+// structFields recurisively gets all fields names in a struct
```
tzaffi, line 606:
```suggestion
// structFields recursively gets all fields names in a struct
```

idb/postgres/internal/encoding/encoding_test.go
```diff
@@ -592,10 +595,55 @@ func TestLcAccountDataEncoding(t *testing.T) {
...
-	expectedString := `{"onl":1,"sel":"DwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","spend":"BgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","stprf":"EwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==","tapl":11,"tapp":10,"tas":13,"tasp":12,"teap":9,"tsch":{"nbs":8,"nui":7},"vote":"DgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","voteFst":16,"voteKD":18,"voteLst":17}`
+	expectedString := `{"onl":1,"sel":"DwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","spend":"BgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","stprf":"EwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==","tapl":11,"tapp":10,"tas":13,"tasp":12,"tbx":20,"tbxb":21,"teap":9,"tsch":{"nbs":8,"nui":7},"vote":"DgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","voteFst":16,"voteKD":18,"voteLst":17}`
 	assert.Equal(t, expectedString, string(buf))
 
 	decodedAd, err := DecodeTrimmedLcAccountData(buf)
 	require.NoError(t, err)
 	assert.Equal(t, ad, decodedAd)
 }
+
+// structFields recursively gets all fields names in a struct
```
tzaffi, line 606:
```suggestion
// structFields recursively gets all field names in a struct
```

util/test/account_testutil.go
tzaffi: I'll make the 4th parameter required and explain in a comment that 0 can be used for falling back to `AssemblerDefaultVersion`

util/test/box_testutil.go
tzaffi: I appreciate your valiant attempt to fight the machine. I'll investigate refactoring along the lines suggested. The most frustration for me during this PR has been working around go's import restrictions for testing.

util/test/box_testutil.go
tzaffi: used in [ledger_for_evaluator_test.go](https://github.com/algorand/indexer/blob/a17bc4b54e068c25204e0634ad9aea191a8d264b/processor/eval/ledger_for_evaluator_test.go#L618)

util/test/box_testutil.go
tzaffi: BTW. There is a significant refactor of testing code in the upcoming followup #1117. It might be worthwhile to punt the discussion of where tests should live exactly till that PR.

tzaffi: In particular, as part of the refactor this file has been deleted:
image

tzaffi: @winder I've analyzed my refactoring for upcoming #1117. That one is not yet ready for review. However, one of the refactorings was doing exactly what you suggested. Moving `BuildAccountDeltasFromKvsAndMods` to [writer_test.go](https://github.com/algorand/indexer/blob/34887da8ca7bc9ed6d1d38166846c3003653c911/idb/postgres/internal/writer/writer_test.go#L1615). Thus I ask that we punt the final refactoring discussion to that PR.

util/test/box_testutil.go
tzaffi: In #1117 it's [moved into that file](https://github.com/algorand/indexer/blob/5500b7f4878ece92a67afa26944acf9abd73bd31/processor/eval/ledger_for_evaluator_test.go#L548)

idb/postgres/internal/writer/writer_test.go
```diff
@@ -1564,3 +1565,174 @@ func TestWriterAddBlock0(t *testing.T) {
...
+	delta.Accts = delta2.Accts
+
+	err = pgutil.TxWithRetry(db, serializable, addNewBlock, nil)
+	require.NoError(t, err)
+
+	validateRow(n1, v1) // untouched
+	validateRow(n2, v2) // new v2
+	validateRow(n3, notPresent)
+	validateRow(n4, v4) // new v4
+	validateRow(n5, notPresent)
```
winder, line 1696:
```suggestion
	validateRow(n5, notPresent)
	validateRow(n6, v6)
```

idb/postgres/internal/writer/writer_test.go
```diff
@@ -1564,3 +1565,174 @@ func TestWriterAddBlock0(t *testing.T) {
...
+	delta.Accts = delta2.Accts
+
+	err = pgutil.TxWithRetry(db, serializable, addNewBlock, nil)
+	require.NoError(t, err)
+
+	validateRow(n1, v1)         // untouched
+	validateRow(n2, v2)         // untouched
+	validateRow(n3, notPresent) // still deleted
+	validateRow(n4, notPresent) // deleted
+	validateRow(n5, v5)         // re-inserted
```
winder, line 1719:
```suggestion
	validateRow(n5, v5)         // re-inserted
	validateRow(n6, v6)         // untouched
```

idb/postgres/internal/writer/writer_test.go
```diff
@@ -1564,3 +1565,174 @@ func TestWriterAddBlock0(t *testing.T) {
...
+	delta.Accts = delta2.Accts
+
+	err = pgutil.TxWithRetry(db, serializable, addNewBlock, nil)
+	require.NoError(t, err)
+
+	validateRow(n1, v1)         // untouched
+	validateRow(n2, v2)         // untouched
+	validateRow(n3, notPresent) // still deleted
+	validateRow(n4, notPresent) // still deleted
+	validateRow(n5, v5)         // untouched
```
winder, line 1735:
```suggestion
	validateRow(n5, v5)         // untouched
	validateRow(n6, v6)         // untouched
```

winder (approved):

PR #1069, opened 2022-06-28
RFC-0001: Rfc 0001 impl

Eric-Warehime:
## Summary

Minimal commit which just adds the interface from https://github.com/algorand/indexer/pull/1061


exporters/exporter.go
```diff
@@ -0,0 +1,27 @@
...
+	Connect(cfg ExporterConfig) error
+
+	// Disconnect will be called during termination of the Indexer process.
+	// There is no guarantee that plugin lifecycle hooks will be invoked in any specific order in relation to one another.
+	// Returns an error if it fails which will be surfaced in the logs, but the process is already terminating.
+	Disconnect() error
+
+	// Receive is called for each block to be processed by the exporter.
+	// Should return an error on failure--retries are configurable.
+	Receive(blockData *rpcs.EncodedBlockCert) error
```
winder, line 23: This isn't quite right. Also, what do you think about wrapping it up in another object so that it's easier to add stuff without changing the function signature?

I'm also suggesting pass by value here, with the individual bits of data as pointers instead. We can make copies of the data object and allow plugins set each piece of data to nil if they want to (i.e. we are currently dropping the Certificate).

```
// ExportData is provided to the Exporter on each round.
type ExportData {
	// Block is the block data written to the blockchain.
	Block       *bookkeeping.Block

	// Delta contains a list of account changes resulting from the block. Processor plugins may have modify this data.
	Delta       *ledgercore.StateDelta

	// Certificate contains voting data that certifies the block. The certificate is non deterministic, a node stops collecting votes once the voting threshold is reached.
	Certificate *agreement.Certificate
}
```
```suggestion
	Receive(data ExportData) error
```

exporters/exporter.go
```diff
@@ -0,0 +1,72 @@
...
+	Config() ExporterConfig
+
+	// Disconnect will be called during termination of the Indexer process.
+	// There is no guarantee that plugin lifecycle hooks will be invoked in any specific order in relation to one another.
+	// Returns an error if it fails which will be surfaced in the logs, but the process is already terminating.
+	Disconnect() error
+
+	// Receive is called for each block to be processed by the exporter.
+	// Should return an error on failure--retries are configurable.
+	Receive(exportData ExportData) error
```
winder, line 63: The main part I'm not sure about is whether we should have this abstraction in the first iteration. It means there will be a required type check in all plugins to convert it into `BlockExportData`.

I think it's ok to leave it like this for now, but lets keep an eye on it.

exporters/exporter.go
```diff
@@ -0,0 +1,72 @@
...
+type ExporterMetadata struct {
+	Name        string
+	Description string
+	Deprecated  bool
+}
+
+// Exporter defines the interface for plugins
+type Exporter interface {
+	// Metadata associated with each Exporter.
+	Metadata() ExporterMetadata
```
winder, line 44: I think you can do something like this, then we can use the same interface for the other plugins
```
type PluginMetadata interface {
	// Metadata associated with each Exporter.
	Metadata() ExporterMetadata
}
```

```suggestion
	PluginMetadata
```

winder (approved):
Looks good!

exporters/exporter.go
Eric-Warehime: Yeah I agree, it seems correct to use the same Metadata construct for all of the plugin types. I'll update this in a subsequent PR to make it more generic. It should probably end up being moved to a separate module in that case as well.

PR #1075, opened 2022-06-28
Generate an error if the catchpoint is not valid for initialization.

winder:
## Summary

If the user provides a catchpoint but it isn't valid for the deployment, we should exit with an error instead of warning.

## Test Plan

New unit test.

### DIFF
--- a/processor/blockprocessor/block_processor.go
+++ b/processor/blockprocessor/block_processor.go
@@ -47,7 +47,7 @@ func MakeProcessorWithLedgerInit(ctx context.Context, logger *log.Logger, catchp
 				return &blockProcessor{}, fmt.Errorf("MakeProcessorWithCatchup() label err: %w", err)
 			}
 			if uint64(round) >= nextDBRound {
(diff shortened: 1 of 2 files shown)


processor/blockprocessor/block_processor.go
```diff
@@ -47,7 +47,7 @@ func MakeProcessorWithLedgerInit(ctx context.Context, logger *log.Logger, catchp
 				return &blockProcessor{}, fmt.Errorf("MakeProcessorWithCatchup() label err: %w", err)
 			}
 			if uint64(round) >= nextDBRound {
-				logger.Warnf("round for given catchpoint is ahead of db round. skip fast catchup")
+				return &blockProcessor{}, fmt.Errorf("catchpoint round is ahead of db round: %d > %d", uint64(round), nextDBRound)
```
shiqizng, line 50: round could equal to nextDBRound

shiqizng (approved):

processor/blockprocessor/block_processor.go
shiqizng: yes

PR #1089, opened 2022-07-01
tests: E2e test update

Eric-Warehime:
## Summary

This change does a few things:
1. Instead of pulling go-algorand binaries via the `update.sh` script grabbing nightlies, we build them via the submodule.
2. Instead of fetching e2e test data from s3, we generate it during the container build based on what is in the submodule.

Reasons for this change are that the e2e test data currently in S3 includes some old stateproofs code using the `cc` compactcert serialization. The new PR for stateproofs changes this in the submodule, but will fail the e2e tests. If I update the e2e test data then the develop branch will break until the PR gets merged--so there is a sort of chicken/egg problem.

This should future proof the e2e tests by ensuring it is always in sync w/ a PR's view of block serialization according to what the submodule contains.

Caveat: This change will make test runs longer--it takes ~5-7 minutes to run `make e2e` now.

## Test Plan

CI


Eric-Warehime:
Because of a lot of the hurdles associated with this PR--the biggest one being the large increase in run time in e2e tests caused by this--I'm going to close this.

PR #1109, opened 2022-07-13
Remove idb dependency from processor.

winder:
## Summary

Remove idb dependency from processor.

## Test Plan

Covered by existing tests.


winder:
Deferring for now, opened #1115

PR #1110, opened 2022-07-13
Misc Local Ledger PR feedback.

winder:
## Summary

Updates from code review.

## Test Plan

N/A, no new functionality.

### DIFF
--- a/api/handlers_e2e_test.go
+++ b/api/handlers_e2e_test.go
@@ -76,7 +76,7 @@ func setupIdb(t *testing.T, genesis bookkeeping.Genesis) (*postgres.IndexerDb, f
 	require.NoError(t, err)
 
 	logger, _ := test2.NewNullLogger()
-	l, err := test.MakeTestLedger(log)
+	l, err := test.MakeTestLedger(logger)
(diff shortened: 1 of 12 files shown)


api/handlers_e2e_test.go
```diff
@@ -78,7 +78,7 @@ func setupIdb(t *testing.T, genesis bookkeeping.Genesis) (*postgres.IndexerDb, f
 	log, _ := test2.NewNullLogger()
 	l, err := test.MakeTestLedger(log)
 	require.NoError(t, err)
-	proc, err := blockprocessor.MakeProcessorWithLedger(l, db.AddBlock)
+	proc, err := blockprocessor.MakeProcessorWithLedger(log, l, db.AddBlock)
```
shiqizng, line 81: in some files, this parameter is named logger or l. can we update them to be the same in all places?

winder: Done, I went with logger. I think thats the name we use most often.

shiqizng, line 81: also rename this to logger?

PR #1114, opened 2022-07-13
Conduit: Postgres exporter initial implementation

Eric-Warehime:
## Summary

Adds an initial module for postgresql in an exporter. It does not touch any of the block_processor or importer modules which will need to be updated to incorporate exporter plugins.

I've created this conduit branch as a feature branch based off localledger since the merge of that to develop will end up blocking development. I've also cherry picked the RFC commit from develop so that is included in this as well.

Where it makes sense I've split out code that I think will be used across multiple plugin types into the plugins package.

## Test Plan

I'll be adding unit tests to this PR. Wanted to get it out asap just to get early feedback.


exporters/exporter.go
```diff
@@ -16,13 +16,13 @@ type ExportData interface {
 // BlockExportData is provided to the Exporter on each round.
 type BlockExportData struct {
```
AlgoStephenAkiki, line 17: Is this going to live in a "common" folder?

Eric-Warehime: It should, I can move it to one.

data/block_export_data.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+)
+
+// RoundProvider is the interface which all data types sent to Exporters should implement
+type RoundProvider interface {
+	Round() uint64
+}
+
+// BlockData is provided to the Exporter on each round.
+type BlockData struct {
```
AlgoStephenAkiki, line 15: It might be useful to standardize an "empty" function here that says if all pointers are nil then the block is empty and the next stage in the pipeline should not proceed.

Eric-Warehime: I added an `Empty` method. Let me know what you think...I don't think the Block should ever be nil--we had discussed just emptying out the set of transactions in a block as a way of still processing every Block, but stripping data that we don't want.

So I've ended up just attempting to return whether the length of the payset is greater than 0.

Eric-Warehime: I think we can create it as we need in the future.

AlgoStephenAkiki: Another that we might want to consider is a "copy from another block" so we can facilitate copying an input block data so we can eventually return it as an output.

Eric-Warehime: I think data can be passed between and modified by each plugin as it proceeds through the processing steps. No need to copy blocks at every stage unless we really need to.

AlgoStephenAkiki (approved):

PR #1117, opened 2022-07-14
New Feature: Lookup Boxes by app-id + Name and Search Boxes by app-id

tzaffi:
## Related PR's
* #1067 
* #1168 

## Summary

Exposing box information for querying indexer. In particular:

* 2 new endpoints:
  * `v2/applications/:application-id/box`
  * `v2/applications/:application-id/boxes`
* returning new fields for `/v2/accounts/:account-id` (account info):
  * `total-box-bytes`
  * `total-boxes`

## TODO / Questions

- [x] Is paging for `.../boxes` working as expected? ANSWER: yes
- [x] Modify the block data on AWS with box info for [make integration](https://github.com/algorand/indexer/blob/75bf63e1b59bb76f628530d758b3140a68c74099/Makefile#L66) ? (this is happening automatically from go-algorand e2e subs)
- [x] Test that ensures we can handle goal/abi encoded box names and values
- [x] Add a `/v2/accounts/:account-id` **fixtures_test** to show that we're exposing the new box fields even for non-app related accounts
- [x] **Question**: Should a test be added to [handlers_test.go](https://github.com/algorand/indexer/blob/99ab66fdadcd77fef4048bb2aa86cae1f06c6e08/api/handlers_test.go#L920) ?This will probably involve implementing [ApplicationBoxes() in mocks/IndexerDb.go](https://github.com/algorand/indexer/blob/571c40918d1143752dd7aade28ec5047ccd90e10/idb/mocks/IndexerDb.go#L85). Tentative answer: NO: The **fixtures test**'s lifecycle is a superset of the lifecycle being tested in the **handlers test**.
- [x] **Question**: Should we handle `MaxBoxResultsPerAppError` a la [MaxAPIResourcesPerAccountError](https://github.com/algorand/indexer/blob/e0b93a7834044e4cd723a9937a4efd4ce56c38a6/idb/idb.go#L284) ? NO - this is unecessary because the built-in limit facilities in indexer should already guard against unexpected response size when querying for boxes. An adjacent question might be: should we take boxes into account as part of the `MaxAPIResourcesPerAccount` controls during the querying of account information? The question here is "no" as well because detailed box information isn't currently included in the `GetAccounts()` query and therefore doesn't impact the limit.
- [x]  **Question**: Should a [Block Generator](https://github.com/algorand/indexer/blob/64e07fda24d2da0c8ccf8c8dbc9018b284298b4a/cmd/block-generator/README.md#L1) test be added for boxes? I believe NO because the end-to-end test covers block processing for blocks containing box transactions. However, I'm open to further discussion.
- [x] The [proposed usage](https://github.com/algorand/indexer/blob/fb71f6d34b498985bf199c43d8421cbf92d0249e/api/test_resources/boxes.json#L760_L800) of `boxes/` search parameter `next` requires a `b64:` rather than providing what was received for `next-token` in the previous response. This violates indexer usage patterns. TODO: adhere to the typical non-modification pattern for `next-token` in indexer. This is now done, but controvertially, `b64:` is being removed altogether. We can adhere to the _opaquness_ property of next while still allowing arbitrary encoding if we simply return the next token with a `b64:` prefix. This is under active discussion. **CONCLUSION: we're returning the b64 prefix in next token, and accepting arbitrary encoding in the same way that box names are handled in the `/box?name=...` endpoint**
- [x] Have gained consensus that `ORDER BY` (i.e. the `order` param) is not exposed in the `boxes/` endpoint, or a consensus in a going forward policy (e.g. issue creation, or change request) **CONCLUSION: hard coded ASC ordering for boxes in `/boxes/` endpoint**
- [X] Should `fixtures_test.go` exist? Or is there sufficient end-to-end testing coverage in `handlers_e2e_test.go` to obviate this new test? Answer: YES, it should exist because it has utility as we can see: _per [this commit](https://github.com/algorand/indexer/pull/1117/commits/b5025ad640fabac0d778b4cac60d558a698ed560) I'm conforming to the same response code and error as with other indexer treatments._ However, I made a commitment to:
  - [x] Explain how to bootstrap a new fixtures test in a **README**
- [x] Reach a consensus on treatment of _box name_ in the `box/` and `boxes/` endpoints. Should this continue to exist as a _required_ query parameter, or should instead become a path parameter? **UPDATE:** a [consensus](https://github.com/algorand/go-algorand/pull/4149#discussion_r968998627) has been reached, and that is to keep the treatment of the endpoints as they currently are in the PR.

## Test Plan

CI


api/indexer.oas2.json
```diff
@@ -2638,23 +2766,6 @@
         }
       }
     },
-    "ErrorResponse": {
```
tzaffi, line 2641: moving `ErrorResponse` model below (more alphabetical)

api/generated/v2/routes.go
```diff
@@ -44,6 +44,12 @@ type ServerInterface interface {
 
 	// (GET /v2/applications/{application-id})
 	LookupApplicationByID(ctx echo.Context, applicationId uint64, params LookupApplicationByIDParams) error
+	// Get box information for a given application.
```
tzaffi, line 47: TODO: improve these description both here in the companian `go-algorand` PR.

idb/postgres/postgres_integration_test.go
```diff
@@ -2255,221 +2254,3 @@ func TestIndexerDb_GetAccounts(t *testing.T) {
 		})
 	}
 }
-
```
tzaffi, line 2258: These have been moved into `idb/postgres/postgres_boxes_test.go`

misc/parity/reports/algod2indexer_dropped.yml
```diff
@@ -6,3 +6,282 @@ definitions:
       - ALGOD: '{"description":"MicroAlgo balance required by the ...'
     required:
     - min-balance
+  ApplicationParams:
```
tzaffi, line 9: TODO: This file needs updating

misc/parity/reports/algod2indexer_full.yml
```diff
@@ -20,3 +20,566 @@ definitions:
     required:
     - - INDEXER: null
       - ALGOD: '"min-balance"'
+  Application:
```
tzaffi, line 23: TODO: This file needs updating

misc/parity/test_indexer_v_algod.py
```diff
@@ -15,7 +15,7 @@
 ASSERTIONS = [DROPPED, FULL]
 
 # When non-empty, keep only:
-PATH_INCLUDES = {"definitions": ["Account"]}
+PATH_INCLUDES = {"paths": [], "definitions": []}
```
tzaffi, line 18: Including paths and not just the models

misc/parity/test_indexer_v_algod.py
```diff
@@ -162,7 +162,7 @@ def test_parity(reports: List[str] = ASSERTIONS, save_new: bool = True):
             old_diff = yaml.safe_load(f)
         new_diff = generate_diff(algod_swgr, indexer_swgr, excludes, diff_type)
 
-        diff_of_diffs = deep_diff(old_diff, new_diff)
+        diff_of_diffs = deep_diff(old_diff, new_diff, arraysets=True)
```
tzaffi, line 165: setting `arraysets=True` results in lists of attributes which only differ by permutations as being regarded as equal.

api/error_messages.go
```diff
@@ -21,15 +21,21 @@ const (
 	errFailedSearchingAsset            = "failed while searching for asset"
 	errFailedSearchingAssetBalances    = "failed while searching for asset balances"
 	errFailedSearchingApplication      = "failed while searching for application"
+	errFailedSearchingBoxes            = "failed while searching for application boxes"
 	errFailedLookingUpHealth           = "failed while getting indexer health"
 	errNoApplicationsFound             = "no application found for application-id"
+	errNoBoxesFound                    = "no applications boxes found"
+	errWrongAppidFound                 = "the wrong application index was found, please contact us, this shouldn't happen"
```
tzaffi, line 28:
```suggestion
	errWrongAppidFound                 = "the wrong application-id was found, please contact us, this shouldn't happen"
```

api/error_messages.go
```diff
@@ -21,15 +21,21 @@ const (
 	errFailedSearchingAsset            = "failed while searching for asset"
 	errFailedSearchingAssetBalances    = "failed while searching for asset balances"
 	errFailedSearchingApplication      = "failed while searching for application"
+	errFailedSearchingBoxes            = "failed while searching for application boxes"
 	errFailedLookingUpHealth           = "failed while getting indexer health"
 	errNoApplicationsFound             = "no application found for application-id"
+	errNoBoxesFound                    = "no applications boxes found"
```
tzaffi, line 27:
```suggestion
	errNoBoxesFound                    = "no application boxes found"
```

api/error_messages.go
```diff
@@ -21,15 +21,21 @@ const (
...
+	errWrongAppidFound                 = "the wrong application index was found, please contact us, this shouldn't happen"
+	errWrongBoxFound                   = "a box with an unexpected name was found, please contact us, this shouldn't happen"
 	ErrNoAccountsFound                 = "no accounts found for address"
 	errNoAssetsFound                   = "no assets found for asset-id"
 	errNoTransactionFound              = "no transaction found for transaction id"
 	errMultipleTransactions            = "multiple transactions found for this txid, please contact us, this shouldn't happen"
 	errMultipleAccounts                = "multiple accounts found for this address, please contact us, this shouldn't happen"
 	errMultipleAssets                  = "multiple assets found for this id, please contact us, this shouldn't happen"
 	errMultipleApplications            = "multiple applications found for this id, please contact us, this shouldn't happen"
+	errMultipleBoxes                   = "multiple applicatoins boxes found for this app id and name, please contact us, this shouldn't happen"
```
tzaffi, line 37:
```suggestion
	errMultipleBoxes                   = "multiple application boxes found for this app id and box name, please contact us, this shouldn't happen"
```

api/error_messages.go
```diff
@@ -21,15 +21,21 @@ const (
...
+	errFailedLookingUpBoxes            = "failed while lookup up application boxes"
```
tzaffi, line 38:
```suggestion
	errFailedLookingUpBoxes            = "failed while looking up application boxes"
```

tzaffi (commented):
typos

misc/parity/reports/algod2indexer_full.yml
```diff
@@ -17,6 +17,16 @@ definitions:
         description:
         - INDEXER: '"Indicates what type of signature is used by this account, must be one of:\n* sig\n* msig\n* lsig\n* or null if unknown"'
         - ALGOD: '"Indicates what type of signature is used by this account, must be one of:\n* sig\n* msig\n* lsig"'
+      total-box-bytes:
```
tzaffi, line 20: Differs from `go-algorand`'s definition. I've [posed a question](https://github.com/algorand/go-algorand/pull/4149/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#r953316978) regarding whether this should be the case.

api/handlers.go
```diff
@@ -557,6 +564,122 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+		BoxName:       boxName,
+	}
+	appid, boxes, round, err := si.fetchApplicationBoxes(ctx.Request().Context(), q)
+
+	if err != nil {
+		msg := fmt.Sprintf("%s: round=?=%d, appid=%d, boxName=%s", errFailedLookingUpBoxes, round, applicationID, encodedBoxName)
+		return indexerError(ctx, fmt.Errorf("%s: %w", msg, err))
+	}
+
+	if len(boxes) == 0 {
```
tzaffi, line 595: query is structured in such a way that:
* no rows back &rarr; no application found
* one row back with null box &rarr; application exists but without the requested box
* one row back with non-null box in the case of successful search

api/handlers.go
```diff
@@ -557,6 +564,122 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+		q.PrevFinalBox = prevBox
+	}
+
+	appid, boxes, round, err := si.fetchApplicationBoxes(ctx.Request().Context(), q)
+
+	if err != nil {
+		msg := fmt.Sprintf("%s: round=?=%d, appid=%d", errFailedSearchingBoxes, round, applicationID)
+		return indexerError(ctx, fmt.Errorf("%s: %w", msg, err))
+	}
+
```
tzaffi, line 652: simlarly to the above, query is structured in such a way that:
* no rows back &rarr; no application found
* more than one row back &rarr; the application exists and the box names for the given "page" are provided within the non-nil results

api/indexer.oas2.json
```diff
@@ -484,6 +484,98 @@
...
+            "name": "application-id",
+            "in": "path",
+            "required": true
+          },
+          {
+            "$ref": "#/parameters/limit"
+          },
+          {
+            "type": "string",
+            "description": "The next page of results starting from a box name in goal-arg form 'encoding:value'. Use the next token provided by the previous results.",
```
tzaffi, line 513:
```suggestion
            "description": "The next page of results with strict lower bound given by box name in goal-arg form 'encoding:value'. Use the next token provided by the previous results.",
```

api/indexer.oas2.json
```diff
@@ -484,6 +484,98 @@
...
+            "name": "next",
+            "in": "query",
+            "x-algorand-format": "GOALarg"
```
tzaffi, line 516: I'm creating this out of "thin air". Should I document in a more transparent fashion somewhere?

tzaffi: @jannotti  ^^^ ?

idb/mocks/IndexerDb.go
```diff
@@ -58,6 +58,29 @@ func (_m *IndexerDb) AppLocalState(ctx context.Context, filter idb.ApplicationQu
 	return r0, r1
 }
 
+// ApplicationBoxes provides a mock function with given fields: ctx, filter
+func (_m *IndexerDb) ApplicationBoxes(ctx context.Context, filter idb.ApplicationBoxQuery) (<-chan idb.ApplicationBoxRow, uint64) {
```
tzaffi, line 62: Not sure how this is used and whether I should use it somehow in testing?

idb/postgres/postgres.go
```diff
@@ -116,6 +117,10 @@ type IndexerDb struct {
 	accountingLock sync.Mutex
 }
 
+// BoxTestComparator is a type used for testing in postres and api
```
tzaffi, line 120: This is a code smell. Does this really need to be here? (I kind of doubt it)

idb/postgres/postgres.go
```diff
@@ -1031,58 +1036,61 @@ func (db *IndexerDb) yieldAccountsThread(req *getAccountsRequest) {
...
-			account.TotalCreatedApps = ad.TotalAppParams
-			account.TotalAssetsOptedIn = ad.TotalAssets
-			account.TotalCreatedAssets = ad.TotalAssetParams
+			account.TotalAppsOptedIn = accountData.TotalAppLocalStates
+			account.TotalCreatedApps = accountData.TotalAppParams
+			account.TotalAssetsOptedIn = accountData.TotalAssets
+			account.TotalCreatedAssets = accountData.TotalAssetParams
+
+			account.TotalBoxes = accountData.TotalBoxes
+			account.TotalBoxBytes = accountData.TotalBoxBytes
```
tzaffi, line 1093: These are the only "real" changes in this function.

api/indexer.oas2.json
jannotti: Is the question where it should be documented? (For that I know nothing with respect to indexer) or for feedback on the interface? My read on that use that we shouldn't need to introduce a token and next mechanism because boxes are inherently ordered. The API can be "give one page of results with all boxes after name `name`". Clients can cycle through them all by sending "" first, and then the last box name from their previous query until an unfilled page is returned. This removes the need to maintain any state at the server.

But then again, if this exact technique has already been used, we could stick with it.

tzaffi: My question is confusing. The paging happening  here is stateless as you suggest and conforms with other indexder endpoints. What is "out of thin air" here is this `"x-algorand-format": "GOALarg"` which I'm proposing to indicate the encoding scheme.

I don't know whether this should be documented in some way, or this self-documentation is good 'nuf.

@winder - any thoughts as well?

jannotti: Oh, yes, I see. How did we document that in `algod`? We have a similar scheme for the box lookup.

tzaffi: I copied the wording from [algod's swagger](https://github.dev/algorand/go-algorand/blob/feature/avm-box/daemon/algod/api/algod.oas2.json#L1377) but I don't know how else we're communicating this. I don't think this is urgent, we can wait till Will comes back from vacay.

tzaffi: That was the wrong link. Here's the updated one: https://github.com/algorand/go-algorand/blob/4688f93ac9bbbb346a871f08602b1fce20ee3d58/daemon/algod/api/algod.oas2.json#L1501

idb/postgres/postgres.go
tzaffi: code smell has been removed

api/indexer.oas2.json
jannotti: Seems like we should use the same language, so this seems right for now.  Worth noting that if someone complains and changes, both should be changed.

api/fixtures_test.go
```diff
@@ -0,0 +1,1139 @@
...
+}
+type responseInfo struct {
+	StatusCode int    `json:"statusCode"`
+	Body       string `json:"body"`
+}
+type prover func(responseInfo) (interface{}, *string)
+
+// ---- BEGIN provers / witness generators ---- //
+
+func accountsProof(resp responseInfo) (wit interface{}, errStr *string) {
```
tzaffi, line 108: I believe these `xyzProof()` functions could be unified with go 1.18 generics. Maybe there's another way?

api/fixtures_test.go
```diff
@@ -0,0 +1,1139 @@
...
+	_, p, e := getProof(path[1:])
+	return p, e
+}
+
+type proofPath struct {
+	parts map[string]proofPath
+	proof prover
+}
+
+var proverRoutes = proofPath{
```
tzaffi, line 234: Currently, only the following 7 endpoints are supported. But it would be easy to add others.
```
v2/accounts
v2/accounts/:account-id
v2/applications
v2/applications/:application-id
v2/applications/:application-id/box
v2/applications/:application-id/boxes
```

api/app_boxes_fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+	"b32":         base32.StdEncoding.EncodeToString([]byte("b32")),
+	"base32":      base32.StdEncoding.EncodeToString([]byte("base32")),
+	"byte base32": base32.StdEncoding.EncodeToString([]byte("byte base32")),
+	"b64":         base64.StdEncoding.EncodeToString([]byte("b64")),
+	"base64":      base64.StdEncoding.EncodeToString([]byte("base64")),
+	"byte base64": base64.StdEncoding.EncodeToString([]byte("byte base64")),
+	"abi":         `(uint64,string,bool[]):[399,"pls pass",[true,false]]`,
+}
+
+func setupLiveBoxes(t *testing.T, proc processor.Processor, l *ledger.Ledger) {
```
tzaffi, line 43: Admittedly, there are D.R.Y. violations in this test (and in the other boxes tests as well). However, in my opinion, violating D.R.Y. for testing isn't always a bad thing. It's easier for tests to evolve separately when they don't need to share common code, and the redundant code doesn't bloat the production binary given go's packaging  method. Also, getting past go's packaging and privacy policy can be daunting.

api/fixtures_test.go
```diff
@@ -0,0 +1,528 @@
...
+	2. Query the test-internal `proverRoutes` struct for a route-compatible `prover`
+	3. Employ the `prover` to parse the indexer's response into a route appropriate `witness` object
+C. Save `xyzLiveFixture` into a non-git-committed fixture `./test_resources/_FIXTURE_NAME.json` (NOTICE THE `_` PREFIX)
+D. Assert that the non-git-commited fixture equals the git-committed version `./test_resources/FIXTURE_NAME.json`
+
+It is the responsibility of the test writer to check that the generated fixture represent the expected results.
+
+----
+TODO: SHOULD this be used to auto-generate unit test cases in the SDK's?
+In particular, one could craft a generic cucumber test that looks something like:
```
tzaffi, line 47: Any thoughts?
@winder @shiqizng @Eric-Warehime @AlgoStephenAkiki @algochoi @michaeldiamant

api/fixtures_test.go
```diff
@@ -0,0 +1,528 @@
...
+	Given an indexer fixture file <file-name>.
+* ("fixture mock setup")
+	When a mock indexer server has been set up using the urls and responses provided by the indexer fixture,
+* ("fixture validation")
+	Then iterate through all of the fixture's test cases: query the mock indexer server, parse the response into an SDK object, and validate that it comports with the witness.
+
+The third step ("fixture validation") is unusually high-level for our cucumber tests, but it would allow for a more streamlined unit tests bootstrapping.
+Implementoing the "fixture validation" step would logically be broken down as follows:
+
+For each `testCase`` in the parsed fixture:
```
tzaffi, line 59:
```suggestion
For each `testCase` in the parsed fixture:
```

api/fixtures_test.go
```diff
@@ -0,0 +1,528 @@
...
+
+		err = mapstructure.Decode(savedCase.Witness, &savedWitness)
+		require.NoError(t, err, msg)
+
+		dequal := reflect.DeepEqual(savedWitness, liveWitness)
+		require.True(t, dequal, msg)
+		require.Equal(t, savedWitness, liveWitness, msg)
+		// require.Equal(t, savedCase.Witness, liveWitness, msg)
+		END OF EXPERIMENTAL */
+
```
tzaffi, line 494:
```suggestion
```

api/handlers_e2e_test.go
```diff
@@ -1428,3 +1435,361 @@ func TestLookupMultiInnerLogs(t *testing.T) {
...
+
+			c, api, rec := setupRequest("/v2/applications/:appidx/box/", "appidx", strconv.Itoa(int(appIdx)))
+			// TODO for prefix in
+			//  []string{"str", "string", "int", "integer", "addr", "address",
+			//		    "b32", "base32", "byte base32",
+			//			"b64", "base64", "byte base64",
+			//			"abi"
+			//			}
+			// ... do the right thing
+			prefixedName := fmt.Sprintf("str:%s", boxName)
```
tzaffi, line 1491:
```suggestion
			prefixedName := fmt.Sprintf("str:%s", boxName)
```

api/handlers_test.go
```diff
@@ -752,47 +750,6 @@ func TestFetchAccountsRewindRoundTooLarge(t *testing.T) {
 	assert.True(t, strings.HasPrefix(err.Error(), errRewindingAccount), err.Error())
 }
 
-// createTxn allows saving msgp-encoded canonical object to a file in order to add more test data
```
tzaffi, line 755: deleting unused helper function

api/test_resources/boxes.json
```diff
@@ -0,0 +1,1341 @@
...
+      },
+      "witness": {
```
tzaffi, line 19: A `witness` is generated by a route's `prover` which typically just parses the response json using generated go types.

api/test_resources/boxes.json
```diff
@@ -0,0 +1,1341 @@
...
+                "num-byte-slice": 0,
+                "num-uint": 0
+              }
+            }
+          },
+          "current-round": 8
+        }
+      },
+      "witnessError": null
+    },
```
tzaffi, line 422: Notice that this app-info response _does not_ contain any information about boxes. To find the summary information for a app's boxes, instead, one will need to query the app's _app-account_ (see below for an example).

@winder @jannotti

api/test_resources/boxes.json
```diff
@@ -0,0 +1,1341 @@
...
+              "pending-rewards": 0,
+              "reward-base": 0,
+              "rewards": 0,
+              "round": 8,
+              "sig-type": "sig",
+              "status": "Offline",
+              "total-apps-opted-in": 0,
+              "total-assets-opted-in": 0,
+              "total-box-bytes": 0,
+              "total-boxes": 0,
```
tzaffi, line 154: Notice that this account which is _not an app account_ still provides box total summaries. Are we okay with this behavior?

@michaeldiamant @jannotti @winder

api/fixtures_test.go
shiqizng: I've used [httpmock](https://github.com/algorand/indexer/blob/develop/processor/blockprocessor/initialize_test.go#L33). I think it's serving similar purpose to the fixtures you're writing.

api/test_resources/boxes.json
```diff
@@ -0,0 +1,1379 @@
...
+            "deleted": false,
+            "pending-rewards": 0,
+            "reward-base": 0,
+            "rewards": 0,
+            "round": 8,
+            "status": "Offline",
+            "total-apps-opted-in": 0,
+            "total-assets-opted-in": 0,
+            "total-box-bytes": 0,
+            "total-boxes": 0,
```
tzaffi, line 490: Even though this is not an app account, we're still are getting the _never_-relevant box totals.

api/fixtures_test.go
tzaffi: Thanks for the tip. I'll research that tool a bit to see if I actually re-invented the wheel and what the pros/cons were.

api/handlers.go
```diff
@@ -934,6 +1057,33 @@ func (si *ServerImplementation) fetchApplications(ctx context.Context, params id
...
+			boxes = append(boxes, result.Box)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return
+	}
+
+	return
```
michaeldiamant, line 1084: @tzaffi Minor:  Does the _suggestion_ match the PR's behavior?

If so, is there a reason to retain the extra branching?  If not, can you help me understand the difference?

```suggestion
	return
```

michaeldiamant: Possibly the intent is to set some error specific values like the surrounding functions.  If so, then disregard the suggestion and I'll look for a commit supplying error specific values.

api/handlers.go
```diff
@@ -557,6 +564,122 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+		return notFound(ctx, fmt.Sprintf("%s: round=%d, appid=%d", errNoApplicationsFound, round, applicationID))
+	}
+
+	if appid != generated.ApplicationId(applicationID) {
+		return indexerError(ctx, fmt.Errorf("%s: round=%d, appid=%d, wrong appid=%d", errWrongAppidFound, round, applicationID, appid))
+	}
+
+	var next *string
+	finalNameBytes := boxes[len(boxes)-1].Name
+	if finalNameBytes != nil {
```
michaeldiamant, line 663: @tzaffi Request for clarification:  In what scenarios can `finalNameBytes == nil`?

From what I understand about upstream codepaths, my intuition is that the value  is always non-nil.

michaeldiamant: For completeness, there's another `nil` check several lines below where I have the same question.

api/indexer.oas2.json
```diff
@@ -484,6 +484,98 @@
...
+        "consumes": [
+          "application/json"
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "tags": [
+          "lookup"
+        ],
+        "operationId": "lookupApplicationBoxByIDandName",
```
michaeldiamant, line 547: @tzaffi Minor:  Is _and_ deliberately lowercased rather than _And_?

I raise the question because it impacts generated method names.

```suggestion
        "operationId": "lookupApplicationBoxByIDAndName",
```

api/indexer.oas2.json
michaeldiamant: @tzaffi Given my current understanding, I prefer to remove `"x-algorand-format": "GOALarg"`.

`x-algorand-format` carries special semantics in _generator_: 
 https://github.com/algorand/generator/search?q=x-algorand-format.  By adding `GOALarg`, it falsely implies _generator_ produces models differently.

api/indexer.oas2.json
```diff
@@ -484,6 +484,98 @@
...
+            "name": "application-id",
+            "in": "path",
+            "required": true
+          },
+          {
+            "$ref": "#/parameters/limit"
+          },
+          {
+            "type": "string",
+            "description": "The next page of results with strict lower bound given by box name in goal-arg form 'encoding:value'. Use the next token provided by the previous results.",
```
michaeldiamant, line 513: @tzaffi Minor:  Let me know if the proposal seems reasonable?

To be consistent with algod API docs, I propose replacing _goal-arg_ with _goal app call arg_.
* Source:  https://github.com/algorand/go-algorand/blob/feature/avm-box/daemon/algod/api/algod.oas2.json
* If a change is made here, there's several additional references in the PR that ought to be changed in tandem.
* If you feel _goal-arg_ is a better fit, then let's briefly discuss.  I want to land on a terminology we apply in indexer + algod consistently.

```suggestion
            "description": "The next page of results with strict lower bound given by box name in goal app call arg form 'encoding:value'. Use the next token provided by the previous results.",
```

api/handlers.go
tzaffi: `finalNameBytes == nil` in the case that the app-id is valid, but the query resulted in no boxes.

Regarding the second `nil` check where I'm skipping nil boxes, a similar effect could be had by having more explicit boundary checks. This approach -while lazy- seemed more robust against query variations. However, I'm happy to dig further with additional test cases and use more explicit checks instead of the current approach. LMK your thoughts.

api/handlers.go
tzaffi: Goot catch.

api/test_resources/boxes.json
tzaffi: (it's not possible -under the current design- to tell the difference between a non-app account and an app-account with no boxes)

api/handlers.go
michaeldiamant: @tzaffi Resolving - Makes sense.  Thanks for the explanation and sorry for the silly question.

Due to my prior experiences, there are circumstances where my mind still operates differently.  It causes me to get tangled like I did here.

api/handlers_e2e_test.go
```diff
@@ -1428,3 +1435,354 @@ func TestLookupMultiInnerLogs(t *testing.T) {
...
+
+		msg := fmt.Sprintf("caseNum=%d, appIdx=%d", caseNum, appIdx)
+
+		expectedBoxes := remainingBoxes[appIdx]
+
+		c, api, rec := setupRequest("/v2/applications/:appidx/boxes", "appidx", strconv.Itoa(int(appIdx)))
+		// TODO: should we add Order to the search params?
+		params := generated.SearchForApplicationBoxesParams{}
+
+		// TODO: also test non-nil Limit, Next
```
michaeldiamant, line 1516: @tzaffi So that I know how to gauge the PR:  Do you intend to address the highlighted TODO in-scope to the PR?

tzaffi: I've been revisiting and resolving these from time to time and hoping that a reviewer would have an obvious answer. I also am waiting for @winder to have a look. I will aim to research the remaining 2 items asap and either resolve or tag someone with an actionable question.

api/test_resources/boxes.json
michaeldiamant: @tzaffi I think _yes_. 

Here are my thoughts - happy to discuss live if helpful.
* My understanding is that algod + indexer do _not_ differentiate between logical kinds of accounts (e.g. app account vs non-app account).
* Perhaps there's a desire to enrich the model because boxes are specific to a kind of account usage.  If so, my initial reaction is to keep as is for now.  I imagine specializing account representation involves _many_ concerns.

api/handlers_e2e_test.go
tzaffi: Oh... didn't realize this was referencing a line of code... Thanks for reminding me of these TODO's. I will look now as we ought not have any remaining TODO's.

cmd/algorand-indexer/daemon.go
```diff
@@ -505,7 +511,7 @@ func blockHandler(proc processor.Processor, retryDelay time.Duration) func(conte
 			case <-ctx.Done():
 				return err
 			case <-time.After(retryDelay):
-				break
+				// NOOP
```
michaeldiamant, line 514: @tzaffi Sanity checking:  Is the removal of the `break` statement by-design?

I'm raising the note because at surface level, the change _seems_ unrelated to Boxes + impacts a seemingly central code path.  I'm wondering if perhaps the change was intended for local-only testing.

api/handlers_e2e_test.go
tzaffi: Answer:
* Order is currently fixed as Ascending. I will make a note of this in the PR description and a TODO that we should gain agreement on this omission, or on a going-forward strategy.
* `non-nil Limit, next`[ have been tested](https://github.com/algorand/indexer/blob/5f1192e9af72214d39b38de2accf50454d760a4d/api/test_resources/boxes.json#L774).

idb/postgres/internal/writer/writer_test.go
```diff
@@ -1565,6 +1565,102 @@ func TestWriterAddBlock0(t *testing.T) {
...
+	acctData.TotalBoxBytes -= prevBoxBytes
+
+	boxTotals[addr] = basics.AccountData{
+		TotalBoxes:    acctData.TotalBoxes,
+		TotalBoxBytes: acctData.TotalBoxBytes,
+	}
+}
+
+// buildAccountDeltasFromKvsAndMods simulates keeping track of the evolution of the box statistics
+func buildAccountDeltasFromKvsAndMods(t *testing.T, kvOriginals, kvMods map[string]*string) (
```
michaeldiamant, line 1614: @tzaffi Tangentially related to the line:  I didn't research the history, but there appears to be a duplicate + unused `BuildAccountDeltasFromKvsAndMods` function defined in `util/test/box_testutil.go`.  

From a cursory review, it looks like `util/test/box_testutil.go` is wholesale unused.  Can it be removed?

api/handlers_e2e_test.go
michaeldiamant: @tzaffi Got it - resolving.

idb/postgres/internal/writer/writer_test.go
tzaffi: Oops. This is a good catch and concerning. I've been through several test refactorings, but I'm surprised this ended up unused. The basic functionality ought to be tested. I will now investigate to see if I duplicated this elsewhere or need to rewire it to an actual test.

tzaffi: After further investigation, all the tests are in place and `util/test/box_testutil.go` is unneeded (probably the result of a poorly resolved merge conflict). I will remove this unnecessary file even as it reduces my line count bragging rights by almost 150.

cmd/algorand-indexer/daemon.go
tzaffi: After catching as similar lint problem in an earlier PR, I cavalierly assumed, that the same problem occurs here. I.e., a `break` is used in a c-style. It's clear that the [behavior is the same](https://go.dev/play/p/kkcq4Qvi330). Also, the comment on line 509:
```
// Delay or terminate before next attempt.
```
`Delay` corresponds to the 2nd part of the switch. However, I think we should get others more familiar with the code to have a look as well.

@AlgoStephenAkiki @winder @shiqizng

api/fixtures_test.go
michaeldiamant: @tzaffi I think I'd benefit from a live walkthrough to better understand the tradeoffs considered here.

I've spent ~20min reviewing the space to build _some_ awareness, though I suspect I don't see the entire space.

api/handlers_e2e_test.go
```diff
@@ -1428,3 +1436,361 @@ func TestLookupMultiInnerLogs(t *testing.T) {
...
+			}
+
+			c, api, rec := setupRequest("/v2/applications/:appidx/box/", "appidx", strconv.Itoa(int(appIdx)))
+			// TODO for prefix in
+			//  []string{"str", "string", "int", "integer", "addr", "address",
+			//		    "b32", "base32", "byte base32",
+			//			"b64", "base64", "byte base64",
+			//			"abi"
+			//			}
+			// ... do the right thing
```
tzaffi:
```suggestion
```

This TODO has been done

api/fixtures_test.go
```diff
@@ -0,0 +1,503 @@
...
+* ("fixture parsing")
+	Given an indexer fixture file <file-name>.
+* ("fixture mock setup")
+	When a mock indexer server has been set up using the urls and responses provided by the indexer fixture,
+* ("fixture validation")
+	Then iterate through all of the fixture's test cases: query the mock indexer server, parse the response into an SDK object, and validate that it comports with the witness.
+
+The third step ("fixture validation") is unusually high-level for our cucumber tests, but it would allow for a more streamlined unit tests bootstrapping.
+Implementoing the "fixture validation" step would logically be broken down as follows:
```
tzaffi, line 57:
```suggestion
Implementing the "fixture validation" step would logically be broken down as follows:
```

misc/parity/reports/algod2indexer_full.yml
tzaffi: Resolving as an issue has been opened in go-algorand: https://github.com/algorand/go-algorand/issues/4478

api/fixtures_test.go
michaeldiamant: Leaving a note after having a verbal with @tzaffi + @shiqizng with my 2c:

* I see overlap between `fixtures_test.go` and `handlers_e2e_test.go`.  I think a primary question for Indexer maintainers to consider is:  What invariants ought to be maintained when modifying + adding endpoints? 
* I'd consider comfort levels with adopting a `fixture_test.go` approach _in place of_ `handlers_e2e_test.go` rather than trying to maintain both approaches.
  * From my perspective, `fixtures_test.go` brings more rigor to API-level testing.  I can see the approach making it easier to catch API regressions in exchange for maintaining a different approach to API-level tests.
  * Ultimately, I think indexer maintainers ought to weigh in on the tradeoff they feel is appropriate.  If helpful, I'm happy to further vocalize my opinion, but I don't think I'm in a decision making role here.
* RE: Usage for SDK test generation - My initial reaction is that the upfront investment is too large relative to the perceived payoff.  So, I'd focus evaluation on usage within indexer without considering downstream SDK testing possibilities.

api/test_resources/boxes.json
```diff
@@ -0,0 +1,1379 @@
...
+            "deleted": false,
+            "pending-rewards": 0,
+            "reward-base": 0,
+            "rewards": 0,
+            "round": 8,
+            "status": "Offline",
+            "total-apps-opted-in": 0,
+            "total-assets-opted-in": 0,
+            "total-box-bytes": 477,
+            "total-boxes": 9,
```
tzaffi, line 566: Here's how you get the box summary info.

api/handlers.go
```diff
@@ -557,6 +559,122 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
 	})
 }
 
+// LookupApplicationBoxByIDAndName returns the value of an application's box
+// (GET /v2/applications/{application-id}/box)
+func (si *ServerImplementation) LookupApplicationBoxByIDAndName(ctx echo.Context, applicationID uint64, params generated.LookupApplicationBoxByIDAndNameParams) error {
```
shiqizng, line 564: please add a check asserting that parameters of type uint64 is <= math.MaxInt64 in all handler methods. this is because indexer db can only store values up to max int64. See existing handlers for examples.

api/handlers.go
```diff
@@ -557,6 +559,122 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+	if appid != generated.ApplicationId(applicationID) {
+		return indexerError(ctx, fmt.Errorf("%s: round=%d, appid=%d, wrong appid=%d, boxName=%s", errWrongAppidFound, round, applicationID, appid, encodedBoxName))
+	}
+
+	if len(boxes) > 1 {
+		return indexerError(ctx, fmt.Errorf("%s: round=%d, appid=%d, boxName=%s", errMultipleBoxes, round, applicationID, encodedBoxName))
+	}
+
+	box := boxes[0]
+	if len(box.Name) == 0 && len(boxName) > 0 {
```
shiqizng, line 603: can box name be empty?

api/handlers.go
```diff
@@ -557,6 +559,122 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+		return notFound(ctx, fmt.Sprintf("%s: round=%d, appid=%d, boxName=%+v", errNoBoxesFound, round, applicationID, encodedBoxName))
+	}
+
+	if string(box.Name) != string(boxName) {
+		return indexerError(ctx, fmt.Errorf("%s: round=%d, appid=%d, boxName=%s", errWrongBoxFound, round, applicationID, encodedBoxName))
```
shiqizng, line 608: should this return not found instead of error 500?

api/handlers.go
tzaffi: If I recall, empty box names can be provided in order to increase the i/o budget of an app-call involving box access, but one cannot actualluy create a box with an empty name.

Is this correct?

@jannotti

api/handlers.go
tzaffi: This indicates a bug in the logic. It should never be the case that we found a box but it has a different name than what we queried for. So I argue for a 500 error.

api/handlers.go
jannotti: Yes. Empty boxes names are not allowed, so empty box refs are "known fake"

api/handlers.go
tzaffi: It turns out that this guard was already in place (probably due to work that you recently put in). By adding 4 new test cases to the fixtures test, we can verify that this is the case.

See the most recent commit.

api/handlers.go
```diff
@@ -569,7 +569,7 @@ func (si *ServerImplementation) LookupApplicationBoxByIDAndName(ctx echo.Context
 	encodedBoxName := params.Name
 	boxNameBytes, err := logic.NewAppCallBytes(encodedBoxName)
 	if err != nil {
-		return badRequest(ctx, err.Error())
+		return badRequest(ctx, fmt.Sprintf("LookupApplicationBoxByIDAndName receieved illegal box name (%s): %s", encodedBoxName, err.Error()))
```
tzaffi, line 572:
```suggestion
		return badRequest(ctx, fmt.Sprintf("LookupApplicationBoxByIDAndName received illegal box name (%s): %s", encodedBoxName, err.Error()))
```

api/fixtures_test.go
Eric-Warehime: Took a brief look, and I agree...the fixtures tests both give me more confidence in the results, and provide more detail on expected vs actual. Some of the existing API handlers tests are incredibly difficult to manually parse/debug because of a combination of b64/msgp encoded data.

I also agree that we should not maintain both sets of tests. 

If the team agrees, let's do the following:
* For the short term I think it should be fine to add these and require that any new API handler tests in the future be added as fixtures. 
* In the medium/longer term let's open an issue for converting the existing tests to the new fixture format.

tzaffi: Not sure how the fixtures test would handle msgp. It's worth further investigation.

api/handlers.go
tzaffi: Upon further investigation, I'm no longer as confident as I was previously that the problem has been addressed. Will keep this thread open and update you as I learn more.

api/indexer.oas2.json
```diff
@@ -485,6 +485,94 @@
...
+          "404": {
+            "$ref": "#/responses/ErrorResponse"
+          },
+          "500": {
+            "$ref": "#/responses/ErrorResponse"
+          }
+        }
+      }
+    },
+    "/v2/applications/{application-id}/box": {
```
winder, line 532: I think this would fit with the existing endpoints a little better as:
`/v2/applications/{application-id}/boxes/{box-name}`

I believe the spec is a little awkward with parameters that appear on the path and as query params, you can look at how we deal with that elsewhere by looking at how the address is handled on `/v2/transactions?account` and `/v2/accounts/{account-id}`

api/indexer.oas2.json
```diff
@@ -1604,6 +1702,43 @@
...
+      "required": [
+        "name", 
+        "value"
+      ],
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "\\[name\\] box name, base64 encoded",
+          "format": "byte",
+          "pattern": "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$",
```
winder, line 1716: The pattern can be left out. I think it became deprecated when we stopped using gorilla mux.

cmd/algorand-indexer/daemon.go
winder: This change is fine, Zeph's suggestion is probably a little more conventional for go programs than having the break.

This code is needed because Indexer takes total control of incoming signals and refuses to return until a graceful shutdown has completed. So when we want a delay between attempts (in case of failure), we need to do so while also checking whether the program needs to terminate. Thus, we check for the `ctx.Done()` signal while waiting for `time.After(retryDelay)` to elapse.

api/fixtures_test.go
```diff
@@ -0,0 +1,490 @@
...
+A. Set up the state of the blockchain
+B. Iterate through a hard-coded `xyzSeedFixture` to generate an `xyzLiveFixture` as follows. For each seed object in `xyzSeedFixture`:
+	1. Query indexer via HTTP for the state of the blockchain using the `seed`'s `Request.Path` + `Request.Params`
+	2. Query the test-internal `proverRoutes` struct for a route-compatible `prover`
+	3. Employ the `prover` to parse the indexer's response into a route appropriate `witness` object
+C. Save `xyzLiveFixture` into a non-git-committed fixture `./test_resources/_FIXTURE_NAME.json` (NOTICE THE `_` PREFIX)
+D. Assert that the non-git-commited fixture equals the git-committed version `./test_resources/FIXTURE_NAME.json`
+
+It is the responsibility of the test writer to check that the generated fixture represent the expected results.
+************************************** */
```
winder, line 44: This seems really complicated. I've spent a bit of time trying to understand how the components go together and am still having trouble.

Could you help me understand the benefits to this system vs the others that we have?

A few particular points that threw me off so far (maybe this is just me):
* `xyz` as a wildcard, I couldn't find `xyzSeedFixture` for a while, eventually I notice `*SeedFixture` functions.
* Using the word `Proof` and a function `getProof` in such close proximity to `StateProofs` and it's own `GetProof` operation.
* generally, there are a lot of moving pieces and I'm having trouble distinguishing between the fixtures and the tests.

api/handlers.go
```diff
@@ -557,6 +559,125 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
 	})
 }
 
+// LookupApplicationBoxByIDAndName returns the value of an application's box
+// (GET /v2/applications/{application-id}/box)
+func (si *ServerImplementation) LookupApplicationBoxByIDAndName(ctx echo.Context, applicationID uint64, params generated.LookupApplicationBoxByIDAndNameParams) error {
+	if err := si.verifyHandler("LookupApplicationBoxByIDAndName", ctx); err != nil {
```
winder, line 565: @AlgoStephenAkiki could you confirm that `verifyHandler` is being used correctly? I don't have any reason to suspect it isn't, just would like you to confirm.

api/handlers.go
```diff
@@ -15,6 +16,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/transactions/logic"
```
winder, line 19: We need to avoid adding new dependencies to go-algorand

api/indexer.oas2.json
tzaffi: It sounds like the proposed change would make boxes work more like other endpoints in `indexer`. However, then it will diverge from the usage in `algod` which I was aiming to keep unified. Another downside is that the SDK work has already been completed for ingesting these new endpoints. However, if the user experience is adversely affected, or it makes developing against indexer's API too burdensome, we should make the change.

winder: Do you know why it was changed? It appears that it used the format I suggested in the original implementation:
https://github.com/algorand/go-algorand/issues/3973

winder: The path indicates that it is a required parameter.

The SDKs would also generate a more intuitive interface. For example:

With the query parameter, the request looks like this:
```
// this is illegal, but allowed.
var badBoxResp = client.lookupApplicationBoxByIDAndName().execute();

// this is legal.
var boxResp = client.lookupApplicationBoxByIDAndName().BoxName("mybox").execute();
```

With a path parameter, it looks like this:
```
var boxResp = client.lookupApplicationBoxByIDAndName("mybox").execute();
```

api/handlers.go
tzaffi: I wanted to enforce the same box encoding behavior as we use in goal and algod, so it seemed prudent to use  the pre-existing `logic.NewAppCallBytes(encodedBoxName)`. I could remove the dependency by re-creating the function in indexer.

api/handlers.go
winder: I believe the other search endpoints do not return a 404, they would just return an empty array.

api/handlers.go
winder: We're going to remove the submodule at some point, which is why I bring it up.

api/handlers.go
winder: Could you return typed errors to make this behavior more explicit? For example the sql package has `sql.ErrNoRows`.

api/indexer.oas2.json
winder: It just occurred to me that if it must be a query parameter, why not have a single `/v2/applications/{application-id}/boxes` endpoint with an optional name filter?

api/handlers.go
michaeldiamant: @winder I'll ultimately follow your guidance here, but want to voice my preference to keep as is.  Does the following rationale work for you?

Rationale:
* I understand we'll move away from the submodule, but the work in-question began prior to the course change.   Since there's multiple changes needed to remove the submodule dependency, the incremental addition here feels marginal.
* Keeping as is makes it easier to understand sources of potential duplication.  If the dependency is dropped in favor of duplication, I think it will be difficult to make a pass later on identifying the duplication.

api/indexer.oas2.json
michaeldiamant: @winder Responding inline below to provide context.  It might be easier to discuss verbally, happy to do so.

> Do you know why it was changed?

* The team went through several iterations prior to landing on the goal app call encoding scheme to safely represent binary data.
* Somewhat subjectively and I can elaborate live:  I preferred a solution using query string parameters over path segments because I think URL encoding query string parameters is less likely to be implemented incorrectly in SDKs (ours or otherwise). 
* In retrospect, I should have raised the API deviation with a broader group sooner.  That's my fault - sorry for that.  I'll obviously be more mindful moving ahead.
* Given that we're already here, my personal preference is to keep as is with the understanding that we're _not_ aiming to propagate the convention elsewhere.  If you feel strongly about retaining existing conventions, I'd prefer to move to a live discussion.

>  why not have a single /v2/applications/{application-id}/boxes endpoint with an optional name filter?

I don't know if it's convincing, but here's a rationale:  The 2 endpoints in-question have different return types.  Keeping distinct endpoints makes the return type explicit + unconditional.

api/handlers.go
winder: I see 3 functions: `logic.NewAppCallBytes`, `logic.SplitBoxKey` and `logic.MakeBoxKey`. They all appear to be small utility functions that would be straightforward to copy.

* The first point doesn't seem fair... it kind of boils down to "you've already got a lot of work ahead, whats a little more?"
* The second makes more sense. Are you imagining that these functions might be pulled out in a similar way to the abi module?

michaeldiamant: @winder Let's focus on the 2nd rationale:  Yes - It feels plausible to me that we'll wish to extract common functionality to license appropriately + minimize the chance of divergence.

winder: Ok, that makes sense. Let's go with your approach and punt on this for now.

api/indexer.oas2.json
tzaffi: removed and resolving

api/handlers.go
tzaffi: Resolving as this has been implemented.

api/handlers.go
tzaffi: From a brief investigation, this sounds plausible. For example here [is the app local states handler](https://github.com/algorand/indexer/blob/cf5eb43112f9b0e966c8ea7637a978073ddd7951/api/handlers.go#L275). When no rows are returned, there isn't an error. Even if the lack of rows occurred because a non-existing account id was provided. On the other hand, [when querying for an account id](https://github.com/algorand/indexer/blob/cf5eb43112f9b0e966c8ea7637a978073ddd7951/api/handlers.go#L234) and no rows are returned, there _is a 404 notFound error_. 

I would argue that the more subtle distinction here between non-existing app (404), vs. existing app with non-existing boxes (200) ought to be emulated by **app local states** handler. I.e., just as a 404 is returned when a non-existing account is looked up, a 404 should be returned when local states of a non-existing account is looked up.

So I'm pushing back but am happy to discuss further.

api/handlers.go
tzaffi: Per [this commit](https://github.com/algorand/indexer/pull/1117/commits/b5025ad640fabac0d778b4cac60d558a698ed560) I'm conforming to the same response code and error as with other indexer treatments.

Resolving

api/handlers.go
tzaffi: I'm using it in two places.

api/handlers.go
tzaffi: Unresolving temporarily, as the commit above shows the convenience of having `fixtures_test.go` lying around.

api/handlers.go
AlgoStephenAkiki: Looks good.  As long as the string matches the function name exactly you should be good.

api/handlers.go
```diff
@@ -557,6 +560,141 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+		return indexerError(ctx, fmt.Errorf("%s: round=%d, appid=%d, boxName=%s", errWrongBoxFound, round, applicationID, encodedBoxName))
+	}
+
+	return ctx.JSON(http.StatusOK, generated.BoxResponse(box))
+}
+
+// SearchForApplicationBoxes returns box names for an app
+// (GET /v2/applications/{application-id}/boxes)
+func (si *ServerImplementation) SearchForApplicationBoxes(ctx echo.Context, applicationID uint64, params generated.SearchForApplicationBoxesParams) error {
+	if err := si.verifyHandler("SearchForApplicationBoxes", ctx); err != nil {
```
AlgoStephenAkiki, line 626: This is good as well.

api/fixtures_test.go
tzaffi: Am resolving per our live discussion. I have committed to better documentation for how to create a fixtures test and put that commitment in the form of a TODO in the PR description.

api/indexer.oas2.json
tzaffi: Resolving as the discussion is now happening elsewhere.

api/handlers.go
tzaffi: Resolving as we orally agreed to continue the distinction that is happening here.

api/README.md
```diff
@@ -20,3 +21,104 @@ Specifically, `uint64` types aren't strictly supported by OpenAPI. So we added a
...
+
+A fixtures test
+
+1. is defined by a go-slice called a _Seed Fixture_  e.g. [var boxSeedFixture](https://github.com/algorand/indexer/blob/b5025ad640fabac0d778b4cac60d558a698ed560/api/app_boxes_fixtures_test.go#L302-L692) which contains request information for making HTTP requests against an Indexer server
+2. iterates through the slice, making each of the defined requests and generating a _Live Fixture_
+3. reads a _Saved Fixture_ from a json file e.g. [boxes.json](./test_resources/boxes.json)
+4. persists the _Live Fixture_ to a json file not in source control
+5. asserts that the _Saved Fixture_ is equal to the _Live Fixture_
+
+In reality, because we always want to save the _Live Fixture_ before making asserstions that could fail the test and preempt saving, steps (3) and (4) happen in the opposite order.
```
tzaffi, line 38:
```suggestion
In reality, because we always want to save the _Live Fixture_ before making assertions that could fail the test and pre-empt saving, steps (3) and (4) happen in the opposite order.
```

api/README.md
```diff
@@ -20,3 +21,104 @@ Specifically, `uint64` types aren't strictly supported by OpenAPI. So we added a
...
+In reality, because we always want to save the _Live Fixture_ before making assertions that could fail the test and pre-empt saving, steps (3) and (4) happen in the opposite order.
+
+## What's the purpose of a Fixtures Test?
+
+A fixtures test should allow one to quickly stand up an end-to-end test to validate that Indexer endpoints are working as expected. After Indexer's state is programmatically set up, it's easy to add new requests and verify that the responses look exactly as expected. Once you're satisfied that the responses are correct, it's easy to _freeze_ the test and guard against future regressions.
+## What does a **Fixtures Test Function** Look Like?
+
+[func TestBoxes](https://github.com/algorand/indexer/blob/b5025ad640fabac0d778b4cac60d558a698ed560/api/app_boxes_fixtures_test.go#L694_L704) shows the basic structure of a fixtures test.
+
+1. `setupLiveServerAndReturnShutdownFunc()` is called to set up the Indexer database
```
tzaffi, line 47:
```suggestion
1. `setupIdbAndReturnShutdownFunc()` is called to set up the Indexer database
```

api/README.md
```diff
@@ -20,3 +21,104 @@ Specifically, `uint64` types aren't strictly supported by OpenAPI. So we added a
...
+3. `setupLiveServerAndReturnShutdownFunc()` is used to bring up an instance of a real Indexer.
+  * this shouldn't need to be modified; however, if running in parallel and making assertions that conflict with other tests, you may need to localize the variable `fixtestListenAddr` and run on a separate port
+  * if running a mock server instead, a different setup function would be needed
+4. `validateLiveVsSaved()` runs steps (1) through (5) defined in the previous section
+  * this is designed to be generic and ought not require much modification going forward
+
+
+## Which Endpoints are Currently _Testable_ in a Fixtures Test?
+
+Endpoints defined in [proverRoutes](https://github.com/algorand/indexer/blob/b955a31b10d8dce7177383895ed8e57206d69f67/api/fixtures_test.go#L213) are testable.
```
tzaffi, line 62:
```suggestion
Endpoints defined in [proverRoutes](https://github.com/algorand/indexer/blob/b955a31b10d8dce7177383895ed8e57206d69f67/api/fixtures_test.go#L232-L263) are testable.
```

api/indexer.oas2.json
```diff
@@ -1613,16 +1711,38 @@
         }
       }
     },
-    "ParticipationUpdates": {
```
tzaffi, line 1616: The `ParticipationUpdates` has been moved further below to starting line 1834 and following the `OnCompletion` section

api/README.md
```diff
@@ -20,3 +21,104 @@ Specifically, `uint64` types aren't strictly supported by OpenAPI. So we added a
...
+## What's the purpose of a Fixtures Test?
+
+A fixtures test should allow one to quickly stand up an end-to-end test to validate that Indexer endpoints are working as expected. After Indexer's state is programmatically set up, it's easy to add new requests and verify that the responses look exactly as expected. Once you're satisfied that the responses are correct, it's easy to _freeze_ the test and guard against future regressions.
+## What does a **Fixtures Test Function** Look Like?
+
+[func TestBoxes](https://github.com/algorand/indexer/blob/b5025ad640fabac0d778b4cac60d558a698ed560/api/app_boxes_fixtures_test.go#L694_L704) shows the basic structure of a fixtures test.
+
+1. `setupIdbAndReturnShutdownFunc()` is called to set up the Indexer database
+  * this isn't expected to require modification
+  2. `setupLiveBoxes()` is used to prepare the local ledger and process blocks in order to bring Indexer into a particular state
```
winder, line 49:
```suggestion
2. `setupLiveBoxes()` is used to prepare the local ledger and process blocks in order to bring Indexer into a particular state
```

api/README.md
```diff
@@ -20,3 +21,104 @@ Specifically, `uint64` types aren't strictly supported by OpenAPI. So we added a
 ## Why do we have indexer.oas2.yml and indexer.oas3.yml?
 
 We chose to maintain V2 and V3 versions of the spec because OpenAPI v3 doesn't seem to be widely supported. Some tools worked better with V3 and others with V2, so having both available has been useful. To reduce developer burdon, the v2 specfile is automatically converted v3 using [converter.swagger.io](http://converter.swagger.io/).
+
+# Fixtures Test
```
winder, line 25: Thanks for adding this!

api/handlers.go
```diff
@@ -557,6 +560,141 @@ func (si *ServerImplementation) LookupApplicationByID(ctx echo.Context, applicat
...
+
+	q := idb.ApplicationBoxQuery{
+		ApplicationID: applicationID,
+		BoxName:       boxName,
+	}
+	appid, boxes, round, err := si.fetchApplicationBoxes(ctx.Request().Context(), q)
+
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return notFound(ctx, fmt.Sprintf("%s: round=%d, appid=%d, boxName=%+v", errNoApplicationsFound, round, applicationID, encodedBoxName))
```
winder, line 591:
```suggestion
			return notFound(ctx, fmt.Sprintf("%s: round=%d, appid=%d, boxName=%s", errNoApplicationsFound, round, applicationID, encodedBoxName))
```

michaeldiamant (approved):

idb/postgres/postgres.go
```diff
@@ -2288,6 +2291,125 @@ func (db *IndexerDb) yieldApplicationsThread(rows pgx.Rows, out chan idb.Applica
...
+		whereArgs = append(whereArgs, queryOpts.BoxName)
+	} else if queryOpts.PrevFinalBox != nil {
+		query += " AND name > $2"
+		whereArgs = append(whereArgs, queryOpts.PrevFinalBox)
+	}
+
+	orderKind := "ASC"
+	if queryOpts.Ascending != nil && !*queryOpts.Ascending {
+		orderKind = "DESC"
+	}
```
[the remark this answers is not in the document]
tzaffi: No. It's not visible. Can we continue the conversation on #1168 ?

PR #1124, opened 2022-07-14
Importer Plugin: Added importer rfc

algoganesh:
## Summary
Importer RFC Proposal and a sample implementation of importer interface using `algod` rest endpoint.


docs/rfc/0003-importer-interface.md
```diff
@@ -0,0 +1,62 @@
...
+
+// Importer defines the interface for importer plugin
+type Importer interface {
+	// Algod returns the importer's algod client
+	Algod() *algod.Client
+
+	// Round returns the recent downloaded block
+	Round() uint64
+
+	// Getblock takes the round number as an input and downloads that specific block, updates the 'lastRound' local variable and returns an encodedBlockCert struct consisting of Block and Certificate
```
Eric-Warehime, line 30: The interface doesn't know about a "`lastRound` local variable". A specific implementation of this may update a local variable, but the interface doesn't necessitate it right?

docs/rfc/0003-importer-interface.md
```diff
@@ -0,0 +1,62 @@
...
+```GO
+package importerplugin
+
+// Importer defines the interface for importer plugin
+type Importer interface {
+	// Algod returns the importer's algod client
+	Algod() *algod.Client
+
+	// Round returns the recent downloaded block
+	Round() uint64
```
Eric-Warehime, line 28: I actually don't know if an `Importer` should be keeping track of which rounds it has imported in the past. It seems like one or both of a `Processor` plugin or `Exporter` which require ordered processing of blocks would be the ones maintaining that state. Then whatever framework is coordinating local ledger catchup, postgres (or other exporter) writes, etc. would just request whatever blocks it needs from the Importer without the Importer caring which rounds it fetched previously.

docs/rfc/0003-importer-interface.md
```diff
@@ -0,0 +1,62 @@
...
+	aclient   *algod.Client
+
+	// lastRound is the last successfully fetched block
+	lastRound uint64
+
+	// log.Logger given by the user through RegisterImporter method
+	log       *log.Logger
+
+	// To determine if the importer should fetch blocks from algod or gossip network or from a file
+	fetchMethod string
```
Eric-Warehime, line 50: It seems like this is meant to switch the import method between one of a few methods. Instead of defining this enum-like string and altering behavior of our implementation based on that, we should just have a separate `Importer` implementation for each fetching method and let the framework instantiate whichever one we want based on user input.

docs/rfc/0003-importer-interface.md
algoganesh: Makes sense, the importer doesn't necessarily need to store any history

docs/rfc/0003-importer-interface.md
algoganesh: Sounds good, I'll update it accordingly.

docs/rfc/0003-importer-interface.md
algoganesh: yeah, there’s no need for interface to facilitate this

algoganesh:
Moved to #1135

PR #1126, opened 2022-07-18
Add the `git submodule update --init` to README

fabrice102:
This step `git submodule update --init` is necessary to build the source code.

## Summary

## Test Plan


winder:
@fabrice102 what behavior were you seeing before running the submodule command? I'm asking because `make` should already be running this command automatically.

fabrice102:
Indeed, some partner had an issue and it looked like the issue was due to that.
But indeed, there is no need to do `git submodule update --init` manually.

PR #1133, opened 2022-07-19
Start block generator at round 1 to match new behavior, wait for indexer process to exit.

winder:
## Summary

Fix the perf testing tool. We no longer fetch round 0 so the generator must start at round 1.


shiqizng (approved):

algobarb (approved):
LGTM. Thank you for going over it with me. Good to merge into this branch.

Issue #1028, opened 2022-06-14
Data Exporter: Performance Testing


winder:
After the changes... run the performance tests.

Eric-Warehime:
![Indexer Import Performance](https://user-images.githubusercontent.com/10753834/191253658-65a8573b-b2e8-413d-acf6-88a41f5efe12.svg)

Eric-Warehime:
Used block-gen to perform 1 hour load test against all scenarios. This is using conduit commit https://github.com/algorand/indexer/commit/756be9bd0f09ba6e1ee6cadd5cd392d17352be0a as HEAD. 

I setup two different tests:
* A "1 NVME" test using an `i4i.4xlarge` instance with both postgresql and indexer running on the single NVME drive.
* A "2 NVME" test using an `i4i.8xlarge` insatnce with postgresql running on one NVME drive and indexer running on the other NVME drive.

In all categories the new conduit branch is running at a higher TPS than the 2.13.0 release. It looks like there was some very high drop off in the payment test TPS numbers between 2.12.4 and 2.13.0--the conduit branch compares more favorably to those, but is still behind.

Eric-Warehime:
[test-results.zip](https://github.com/algorand/indexer/files/9607492/test-results.zip)
These are the log files and test results from both of the conduit test runs.

Eric-Warehime:
After discussing w/ John, he believes that the 2.13.0 tests may have been setup differently than the conduit test runs. He mentioned in his import perf testing of local ledger changes he noticed a huge slow down when the ledger was on a gp2 EBS drive versus an ephemeral NVME. Those are older test results, so there's no way to know exactly the test setup--we need to continue to work towards automating the process to eliminate bad data caused by differences in test setups and build up a more in depth historical data set.

PR #1134, opened 2022-07-20
Conduit: Exporter daemon integration

Eric-Warehime:
## Summary

The goal of this PR is to integrate the new Exporter code with the existing Indexer daemon. In this way we can enable users to create and plug new data destinations into the existing Indexer framework.

I'm keeping this as a draft to work through the issues until it is ready to merge into our integration/feature branch.

A lot of these commits are already in PR in https://github.com/algorand/indexer/pull/1114 as well.

## Test Plan

I'll be adding additional unit tests as I go, and tracking performance testing in https://github.com/algorand/indexer/issues/1028


Eric-Warehime:
Moving to #1147

PR #1119, opened 2022-07-14
Added processor rfc

AlgoStephenAkiki:
Resolves #1098 

RFC Proposal

PR #1135, opened 2022-07-20
Conduit: Algod importer implementation

algoganesh:
## Summary
Defines interface for importer plugins and a sample implementation of importer interface using `algod` rest endpoint.


docs/rfc/0003-importer-interface.md
```diff
@@ -0,0 +1,33 @@
...
+Importer interface allows users to fetch any particular block either from algod rest endpoint (or directly from the network or from a file written to by the exporter plugin).
+
+### Plugin Interface
+Importer plugins that are native to the Indexer (maintained within the Indexer repository) will each implementation the importer interface:
+
+```GO
+package importers
+
+// Importer defines the interface for importer plugins
+type Importer interface {
```
Eric-Warehime, line 23: The other plugin definitions have some useful methods I think we're missing here.

`Metadata() MetaData`  
This is just a convenience at this point--centralizing plugin name info, description, deprecation data, etc. In the future the idea is for the Conduit framework to use this data.

`Init(cfg Config) error`  
Your  algod importer implementation is just getting constructed with some algod client, but we should be constructing that in the Init function. The config should just be an object passed in with all the required params. This is I guess similar to your `RegisterAlgodImporter` function, but it just standardizes these across different implementations.

`Close() error`  
This is for closing network connections, files, flushing buffers, etc. Not sure if this is just a no-op for most importers, but we still probably want it.

docs/rfc/0003-importer-interface.md
```diff
@@ -0,0 +1,33 @@
...
+	// GetBlock given any round number rnd fetches the block at that round
+	// It returns an object of type BlockExportData defined in exporters plugin
+	GetBlock(rnd uint64) (*exporters.BlockExportData, error)
```
shiqizng, line 26: does the processor take block of type exporters.BlockExportData too?

importers/algod/algod_importer.go
```diff
@@ -0,0 +1,81 @@
...
+	return algodImporterMetadata
+}
+
+func (bot *algodImporter) Init(cfg string, logger *logrus.Logger) (err error) {
+	bot.logger = logger
+	if err := bot.unmarhshalConfig(cfg); err != nil {
+		return fmt.Errorf("connect failure in unmarshalConfig: %v", err)
+	}
+	var client *algod.Client
+	if !strings.HasPrefix(bot.cfg.NetAddr, "http") {
```
shiqizng, line 46: nit: maybe use url.Parse() and check the Scheme? the current check doesn't cover a case like "httpsomeaddress".

importers/algod/algod_importer_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestClose(t *testing.T) {
+	bot := New()
+	bot.Close()
+	assert.Equal(t, bot, &algodImporter{})
+}
+
+func TestInit(t *testing.T) {
+	bot := New()
+	s := "netaddr: ''\ntoken: ''"
+	err := bot.Init(s, lg)
```
shiqizng, line 30: does this work? isn't lg a nil pointer?

algoganesh: it works, but ill update test to initialize it.

docs/rfc/0003-importer-interface.md
Eric-Warehime: It should. I think we're still trying to figure out what the best data format is to pass between all of the plugin. 
Some discussion happening in https://github.com/algorand/indexer/pull/1119#discussion_r926010276

importers/importer.go
```diff
@@ -0,0 +1,24 @@
...
+	// Init will initialize each importer with a given config. This config will contain the Unmarhsalled config file specific to this plugin.
+	// It is called during initialization of an importer plugin such as setting up network connections, file buffers etc.
+	Init(cfg plugins.PluginConfig, logger *logrus.Logger) error
+
+	// Close function is used for closing network connections, files, flushing buffers etc.
+	Close() error
+
+	// GetBlock, given any round number-rnd fetches the block at that round
+	// It returns an object of type BlockExportData defined in exporters plugin
+	GetBlock(rnd uint64) (*data.BlockData, error)
```
AlgoStephenAkiki, line 23: Is there a reason we are returning `BlockData` as a pointer?  Is it optional (i.e. if GetBlock returns nothing).

importers/importer.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// Importer defines the interface for importer plugins
+type Importer interface {
+	// Metadata associated with each Importer.
+	Metadata() ImporterMetadata
+
+	// Init will initialize each importer with a given config. This config will contain the Unmarhsalled config file specific to this plugin.
+	// It is called during initialization of an importer plugin such as setting up network connections, file buffers etc.
+	Init(cfg plugins.PluginConfig, logger *logrus.Logger) error
```
AlgoStephenAkiki, line 16: Recommend adding a `context.Context` parameter

importers/algod/algod_importer.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return
+	}
+	bot.aclient = client
+	return
+}
+
+func (bot *algodImporter) Close() (err error) {
+	bot.cancel()
+	bot = nil
```
Eric-Warehime, line 67: I've never seen this before. What is the purpose of doing this?

algoganesh: `bot = nil` is unnecessary, will remove it

importers/algod/algod_importer_test.go
```diff
@@ -0,0 +1,64 @@
...
+	s := "netaddr: ''\ntoken: ''"
+	err := bot.Init(ctx, plugins.PluginConfig(s), logger)
+	assert.NoError(t, err)
+	assert.NotEqual(t, bot, nil)
+	bot.Close()
+}
+
+func TestGetBlock(t *testing.T) {
+	bot := New()
+	s := "netaddr: 'https://node-archival-mainnet.internal.aws.algodev.network'\ntoken: '9XxlZqyx27XDrFvV0JU1EVuxRzXJU96Peo07bK0oqslfBeNZdBHXab53D2eui72ib'"
```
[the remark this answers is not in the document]
algoganesh: I strongly agree with your point.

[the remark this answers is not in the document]
winder: Yes, this should be removed.

importers/algod/algod_importer.go
```diff
@@ -0,0 +1,98 @@
...
+	if err := algodImp.unmarhshalConfig(string(cfg)); err != nil {
+		return fmt.Errorf("connect failure in unmarshalConfig: %v", err)
+	}
+	var client *algod.Client
+	u, err := url.Parse(algodImp.cfg.NetAddr)
+	if err != nil {
+		return err
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		algodImp.cfg.NetAddr = "http://" + algodImp.cfg.NetAddr
```
Eric-Warehime, line 55: Should we log this?

importers/algod/algod_importer.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (algodImp *algodImporter) Config() plugins.PluginConfig {
+	s, _ := yaml.Marshal(algodImp.cfg)
+	return plugins.PluginConfig(s)
+}
+
+func (algodImp *algodImporter) Close() error {
+	var err error
+	algodImp.cancel()
+	return err
```
Eric-Warehime, line 73:
```suggestion
	algodImp.cancel()
	return nil
```

importers/algod/algod_importer.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (algodImp *algodImporter) GetBlock(rnd uint64) (data.BlockData, error) {
+	var blockbytes []byte
+	var err error
+	var blk data.BlockData
+	aclient := algodImp.aclient
+	blockbytes, err = aclient.BlockRaw(rnd).Do(algodImp.ctx)
```
Eric-Warehime, line 81:
```suggestion
	blockbytes, err = algodImp.aclient.BlockRaw(rnd).Do(algodImp.ctx)
```

importers/algod/algod_importer_test.go
```diff
@@ -0,0 +1,120 @@
...
+	testImporter importers.Importer
+)
+
+func init() {
+	logger, _ = test.NewNullLogger()
+	ctx, cancel = context.WithCancel(context.Background())
+	s = ""
+}
+
+func MockAlgodServerReturnsEmptyBlock() *httptest.Server {
```
Eric-Warehime, line 38: We have somewhat similar code in https://github.com/algorand/indexer/blob/develop/fetcher/fetcher_test.go#L103

Also, Will mentioned that block-generator might be doing the same. We should look to refactor these into commonly used conventions for mocking Algod in the future.

Eric-Warehime (approved):
LGTM. Left some minor comments.

importers/algod/algod_importer_test.go
algoganesh: I agree. Also `MockAlgodServerReturnsEmptyBlock()` is intentionally exported for other plugins to utilize in testing.

importers/algod/algod_importer.go
algoganesh: I think its a good idea to log it

importers/algod/algod_importer.go
```diff
@@ -0,0 +1,97 @@
...
+		return fmt.Errorf("connect failure in unmarshalConfig: %v", err)
+	}
+	var client *algod.Client
+	u, err := url.Parse(algodImp.cfg.NetAddr)
+	if err != nil {
+		return err
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		algodImp.cfg.NetAddr = "http://" + algodImp.cfg.NetAddr
+		algodImp.logger.Warnf("Plugin config has missing or invalid URL scheme")
```
Eric-Warehime, line 56: I think it's ok if people use like `127.0.0.1:8080` as their NetAddr, so I would go with an info log instead of warning. I guess the main usefulness of the log is to say that we've defaulted you to http instead of https.

```suggestion
		algodImp.logger.Infof("Algod Importer added http prefix to NetAddr: %s", algodImp.cfg.NetAddr)
```

Eric-Warehime (changes requested):

Eric-Warehime (approved):

PR #1139, opened 2022-07-22
Conduit: Rfc updates

Eric-Warehime:
## Summary

Updates the existing exporter code paths to match the changes discussed in https://github.com/algorand/indexer/pull/1119/

### DIFF
--- a/data/block_export_data.go
+++ b/data/block_export_data.go
@@ -20,7 +20,7 @@ type BlockData struct {
 	BlockHeader bookkeeping.BlockHeader
 
 	// Payset is the set of data the block is carrying--can be modified as it is processed
-	Payset transactions.Payset
+	Payset []transactions.SignedTxnInBlock
 
 	// Delta contains a list of account changes resulting from the block. Processor plugins may have modify this data.
 	Delta *ledgercore.StateDelta
--- a/data/block_export_data.go
+++ b/data/block_export_data.go
@@ -3,6 +3,7 @@ package data
 import (
 	"github.com/algorand/go-algorand/agreement"
 	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger/ledgercore"
 )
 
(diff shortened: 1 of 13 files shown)


data/block_export_data.go
```diff
@@ -14,8 +15,12 @@ type RoundProvider interface {
...
 // BlockData is provided to the Exporter on each round.
 type BlockData struct {
-	// Block is the block data written to the blockchain.
-	Block *bookkeeping.Block
+
+	// BlockHeader is the immutable header from the block
+	BlockHeader bookkeeping.BlockHeader
+
+	// Payset is the set of data the block is carrying--can be modified as it is processed
+	Payset *transactions.Payset
```
AlgoStephenAkiki, line 23: Does this need to be a pointer?  The Payset is a slice so it can have a len of 0 by default.

AlgoStephenAkiki (approved):

algoganesh (approved):

data/block_export_data.go
```diff
@@ -14,8 +15,12 @@ type RoundProvider interface {
...
 // BlockData is provided to the Exporter on each round.
 type BlockData struct {
-	// Block is the block data written to the blockchain.
-	Block *bookkeeping.Block
+
+	// BlockHeader is the immutable header from the block
+	BlockHeader bookkeeping.BlockHeader
+
+	// Payset is the set of data the block is carrying--can be modified as it is processed
+	Payset transactions.Payset
```
winder, line 23: nit: this is an alias for `[]SignedTxnInBlock`, that type might be easier for plugin developers

winder (approved):
LGTM, I'd just consider the one type alias change

util/test/account_testutil.go
```diff
@@ -602,3 +620,11 @@ func MakeGenesisBlock() bookkeeping.Block {
 	}
 	return genesisBlock
 }
+
+// ArbitraryString should be used to generate a pseudo-random string to put in the Note field of a Txn Header.
+// This is necessary to ensure the hash of any two txns used in tests are never the same.
+func ArbitraryString() []byte {
```
winder, line 626: Does this need to be public?

Eric-Warehime: Not within this PR since the only thing using it is in the same module, but I see no reason why it couldn't be used in another test module in the future. 

If you want, I can change it to private for this commit and if we need it elsewhere we can change it to public at that time?

Issue #1098, opened 2022-07-08
Create Processor Interface RFC


winder:
Create RFC describing the processor interface.

The LocalLedger can be used as a starting point. It uses specific Make functions instead of a general Init interface.

[Sample template.](https://github.com/algorand/indexer/pull/1061/files#diff-b9b6142dd28fafd79a4926f2db48dd9b9c64d827d29dfdd9bd35d2ab84cb3051)

PR #1140, opened 2022-07-22
Create RFC for processor

AlgoStephenAkiki:
Resolves https://github.com/algorand/indexer/issues/1098

RFC Proposal


processors/metadata.go
```diff
@@ -0,0 +1,30 @@
+package processors
+
+import "github.com/algorand/indexer/plugins"
+
+// ProcessorMetadata returns fields relevant to identification and description of plugins.
+type ProcessorMetadata struct {
+	ImpName        string
+	ImpDescription string
+	ImpDeprecated  bool
```
Eric-Warehime, line 9: Not sure if "Imp" here stands for "Implementation", but I think it's a bit confusing considering we're going to have "Importers" as well.

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
 	logger  *log.Logger
 }
 
-// MakeProcessorWithLedger creates a block processor with a given ledger
-func MakeProcessorWithLedger(logger *log.Logger, l *ledger.Ledger, handler func(block *ledgercore.ValidatedBlock) error) (processor.Processor, error) {
+// Constructor is the ProcessorConstructor implementation for the "block processor" processor
+type Constructor struct{}
+
+// New initializes a postgresqlExporter
```
Eric-Warehime, line 35: Update doc string

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
...
+		ImpDeprecated:  false,
+	}
+}
+
+func (proc *blockProcessor) Config() plugins.PluginConfig {
+	return ""
+}
+
+func (proc *blockProcessor) Init(ctx context.Context, cfg plugins.PluginConfig) error {
+	processors.RegisterProcessor(implName, &Constructor{})
```
Eric-Warehime, line 55: `Register` should be called in a module's `init` function to load the constructor when the module is loaded in.

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
...
+		ImpDescription: "Local Ledger Block Processor",
+		ImpDeprecated:  false,
+	}
+}
+
+func (proc *blockProcessor) Config() plugins.PluginConfig {
+	return ""
+}
+
+func (proc *blockProcessor) Init(ctx context.Context, cfg plugins.PluginConfig) error {
```
Eric-Warehime, line 54: I think in this function you should load all of the constructor variables from the `cfg` that are necessary to create a ledger. Then call `MakeProcessorWithLedger` to actually construct things.

processors/blockprocessor/block_processor.go
```diff
@@ -39,17 +145,17 @@ func MakeProcessorWithLedger(logger *log.Logger, l *ledger.Ledger, handler func(
 	return &blockProcessor{logger: logger, ledger: l, handler: handler}, nil
 }
 
-// MakeProcessorWithLedgerInit creates a block processor and initializes the ledger.
-func MakeProcessorWithLedgerInit(ctx context.Context, logger *log.Logger, catchpoint string, genesis *bookkeeping.Genesis, nextDBRound uint64, opts idb.IndexerDbOptions, handler func(block *ledgercore.ValidatedBlock) error) (processor.Processor, error) {
+// MakeProcessorWithLedgerInit creates a block processors and initializes the ledger.
```
shiqizng, line 148: nit: typo?

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
...
+	return nil
+}
+
+func (proc *blockProcessor) Close() error {
+	return nil
+}
+
+func (proc *blockProcessor) Process(input data.BlockData) (data.BlockData, error) {
+
+	rval := data.MakeBasicOutputBlockData(input)
```
shiqizng, line 65: rename rval to something like blockData?

processors/blockprocessor/block_processor.go
```diff
@@ -75,14 +181,14 @@ func addGenesisBlock(l *ledger.Ledger, handler func(block *ledgercore.ValidatedB
 	return nil
 }
 
-// Process a raw algod block
-func (proc *blockProcessor) Process(blockCert *rpcs.EncodedBlockCert) error {
+// ProcessBlockCert Process a raw algod block
+func (proc *blockProcessor) ProcessBlockCert(blockCert *rpcs.EncodedBlockCert) error {
```
shiqizng, line 185: ProcessBlockCert() and Proccess() seem to have a lot of overlap. why do we need both?

processors/processor.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/rpcs"
+	"github.com/algorand/indexer/data"
+	"github.com/algorand/indexer/plugins"
+)
+
+// BlockProcessor is the block processors interface
+type BlockProcessor interface {
```
winder, line 12: I'm confused, why are we using a different interface for the block processor?

api/handlers_e2e_test.go
```diff
@@ -120,7 +120,7 @@ func TestApplicationHandlers(t *testing.T) {
 	block, err := test.MakeBlockForTxns(test.MakeGenesisBlock().BlockHeader, &txn, &optInTxnA, &optInTxnB)
 	require.NoError(t, err)
 
-	err = proc.Process(&rpcs.EncodedBlockCert{Block: block})
+	err = proc.ProcessBlockCert(&rpcs.EncodedBlockCert{Block: block})
```
winder, line 123: Why not create the BlockData object instead and stick with the generic interface?

cmd/validator/core/command.go
```diff
@@ -43,7 +43,7 @@ func init() {
 	ValidatorCmd.Flags().IntVarP(&config.RetryDelayMS, "retry-delay", "", 1000, "Time in milliseconds to sleep between retries.")
 	ValidatorCmd.Flags().StringVar(&addr, "addr", "", "If provided validate a single address instead of reading Stdin.")
 	ValidatorCmd.Flags().IntVar(&threads, "threads", 4, "Number of worker threads to initialize.")
-	ValidatorCmd.Flags().IntVar(&processorNum, "processor", 0, "Choose compare algorithm [0 = Struct, 1 = Reflection]")
+	ValidatorCmd.Flags().IntVar(&processorNum, "processors", 0, "Choose compare algorithm [0 = Struct, 1 = Reflection]")
```
winder, line 46: All the renames in this file are wrong.

cmd/validator/core/validator.go
```diff
@@ -70,7 +70,7 @@ type ErrorDetails struct {
 	Diff    []string
 }
 
-// ProcessorID is used to select which processor to use for validation.
+// ProcessorID is used to select which processors to use for validation.
```
winder, line 73: These renames are also wrong.

processors/blockprocessor/block_processor.go
winder: I'm not sure why you need this. This function is a shallow copy, so you could set `input.Payset` directly.

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
...
+
+func (proc *blockProcessor) Close() error {
+	return nil
+}
+
+func (proc *blockProcessor) Process(input data.BlockData) (data.BlockData, error) {
+
+	rval := data.MakeBasicOutputBlockData(input)
+
+	if input.Certificate == nil {
```
winder, line 67: Do we use the certificate? Why is it required?

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
...
+func (proc *blockProcessor) Init(ctx context.Context, cfg plugins.PluginConfig) error {
+	processors.RegisterProcessor(implName, &Constructor{})
+	return nil
+}
+
+func (proc *blockProcessor) Close() error {
+	return nil
+}
+
+func (proc *blockProcessor) Process(input data.BlockData) (data.BlockData, error) {
```
winder, line 63: This looks a lot like `ProcessBlockCert`, why aren't you using the interface?

processors/metadata.go
winder: +1, I think it's fine to go with Name/Description/Deprecated

winder (changes requested):
We need to figure out why `Process` and `ProcessBlockCert` are needed. If we don't have the right interface now is the time to fix it.

processors/blockprocessor/block_processor.go
```diff
@@ -27,8 +29,112 @@ type blockProcessor struct {
...
+	return ""
+}
+
+func (proc *blockProcessor) Init(ctx context.Context, cfg plugins.PluginConfig) error {
+	processors.RegisterProcessor(implName, &Constructor{})
+	return nil
+}
+
+func (proc *blockProcessor) Close() error {
+	return nil
```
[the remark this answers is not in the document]
algoganesh: Can also include cancellation of a local context.Context variable initialized when Init(...) was called

processors/metadata.go
AlgoStephenAkiki: Name wont work since it is also a function for ProcessorMetadata.  Will go with "ImplementationName" etc

processors/processor.go
AlgoStephenAkiki: The BlockProcessor was the original interface made by Shiqi, the new "Processor" interface is the plugin interface.  The `blockProcessor` is the implementation of both BlockProcessor and Processor

api/handlers_e2e_test.go
AlgoStephenAkiki: This was a simple rename of the original interface which had the same name as "Process"

processors/blockprocessor/block_processor.go
AlgoStephenAkiki: Updated

processors/metadata.go
```diff
@@ -0,0 +1,30 @@
+package processors
+
+import "github.com/algorand/indexer/plugins"
+
+// ProcessorMetadata returns fields relevant to identification and description of plugins.
+type ProcessorMetadata struct {
```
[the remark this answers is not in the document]
AlgoStephenAkiki: I think it would be a good idea.  @Eric-Warehime  Thoughts?

processors/blockprocessor/block_processor.go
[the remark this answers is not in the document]
AlgoStephenAkiki: @algoganesh isn't the context passed from the Conduit framework?  Wouldn't that mean that Conduit would need to cancel it?

processors/blockprocessor/block_processor.go
AlgoStephenAkiki: I've consolidated a lot of common functionality.  I've maintained `ProcessBlockCert` to give us backwards functionality with the current Indexer as a stepping stone.

processors/blockprocessor/block_processor.go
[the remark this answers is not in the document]
algoganesh: I was considering situations when the framework would like to abort ongoing actions by a particular plugin by calling that plugin's `Close()` function without having to cancel the original context, since that original context could be mutual to other plugins.

processors/config.go
```diff
@@ -0,0 +1,13 @@
+package processors
+
+type BlockProcessorConfig struct {
+	// Catchpoint to initialize the local ledger to
+	Catchpoint string `yaml:"catchpoint"`
+	GenesisFilePath string `yaml:"genesis-file-path"`
```
winder, line 7: We usually don't have access to the genesis file, it's fetched from algod.

processors/processor.go
winder: I need to review this more closely, but now that you've separated the Process/OnComplete functions I wonder if this interface could be part of the test code instead of the processor. You could return this from `setupIdb`

i.e.
```
func handle(proc processors.Processor, handler func(block *ledgercore.ValidatedBlock)) func(cert *rpcs.EncodedBlockCert) error {
    return func(cert *rpcs.EncodedBlockCert) error {
        blockData := proc.Process(BlockData{ Certificate: cert.cert, Block: cert.Block })
        handler(MakeValidatedBlock(blockData.Block, blockData.StateDelta))
        proc.OnComplete(blockData)
    }
}
```

processors/config.go
```diff
@@ -0,0 +1,14 @@
+package processors
+
+// BlockProcessorConfig configuration for a block processor
+type BlockProcessorConfig struct {
+	// Catchpoint to initialize the local ledger to
+	Catchpoint      string `yaml:"catchpoint"`
+	GenesisFilePath string `yaml:"genesis-file-path"`
+	NextDbRound     uint64 `yaml:"next-db-round"`
```
AlgoStephenAkiki, line 8: Can't be in config

processors/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,386 @@
...
+		blockData, err := (*proc).Process(data.MakeBlockDataFromEncodedBlockCertificate(data.BlockData{}, cert))
+		if err != nil {
+			return err
+		}
+
+		proto, ok := config.Consensus[cert.Block.BlockHeader.CurrentProtocol]
+		if !ok {
+			return fmt.Errorf(
+				"cannot find proto version %s", cert.Block.BlockHeader.CurrentProtocol)
+		}
```
Eric-Warehime, line 362: We call the handler func at the end of this function, which I think is just the [importer.ImportBlock](https://github.com/algorand/indexer/blob/conduit/importer/importer.go#L24) function that also checks the protocol version. 

It's also called in `extractValidatedBlockAndPayset` during `Process`.

So I don't think this is necessary here.

processors/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,386 @@
...
+		var vb ledgercore.ValidatedBlock
+		if protoChanged {
+			block := bookkeeping.Block{
+				BlockHeader: cert.Block.BlockHeader,
+				Payset:      blockData.Payset,
+			}
+			vb = ledgercore.MakeValidatedBlock(block, *blockData.Delta)
+		} else {
+			vb = ledgercore.MakeValidatedBlock(cert.Block, *blockData.Delta)
+		}
```
Eric-Warehime, line 375: Same comment as above. This code is already being executed in `Process`.

data/block_export_data.go
```diff
@@ -13,6 +15,13 @@ type RoundProvider interface {
 	Empty() bool
 }
 
+// InitProvider is the interface that can be used when initializing to get common algod related
+// variables
+type InitProvider interface {
```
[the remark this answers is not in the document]
AlgoStephenAkiki: I think we should extend it to exporters.  Reason being is that adding a variable is simply adding to the interface and doesn't involve changing `N` number of function signatures.  I also think it is cleaner personally.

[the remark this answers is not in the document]
winder: It's similar to the "parameter object" pattern, I'm a fan.

processors/blockprocessor/block_processor.go
AlgoStephenAkiki: The problem is the conditional relies on either using the original certified block or the newly evaluated modified transactions.  The Process function returns the newly evaluated transactions without access to the original so I would have to use a conditional anyway.

I could always expose `extractValidatedBlockAndPayset` but I feel that would take away from the purpose of "only using public interfaces".  However this is a legacy adapter so maybe that is warranted.  Ping me on slack and we can call and discuss.

shiqizng (commented):
the refactored block processor code looks correct. I just left a few small suggestions.

processors/processor_factory.go
```diff
@@ -0,0 +1,45 @@
...
+
+// processorImpls is a k/v store from processor names to their constructor implementations.
+// This layer of indirection allows for different processor integrations to be compiled in or compiled out by `go build --tags ...`
+var processorImpls = make(map[string]ProcessorConstructor)
+
+// RegisterProcessor is used to register ProcessorConstructor implementations. This mechanism allows
+// for loose coupling between the configuration and the implementation. It is extremely similar to the way sql.DB
+// driver's are configured and used.
+func RegisterProcessor(name string, constructor ProcessorConstructor) {
+	processorImpls[name] = constructor
```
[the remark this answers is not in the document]
AlgoStephenAkiki: I could return an error.

shiqizng (approved):
Thanks for the updates.

Eric-Warehime (approved):

processors/blockprocessor/initialize.go
```diff
@@ -13,31 +13,30 @@ import (
...
 
 // InitializeLedger will initialize a ledger to the directory given by the
 // IndexerDbOpts.
 // nextRound - next round to process after initializing.
 // catchpoint - if provided, attempt to use fast catchup.
-func InitializeLedger(ctx context.Context, logger *log.Logger, catchpoint string, nextRound uint64, genesis bookkeeping.Genesis, opts *idb.IndexerDbOptions) error {
-	if nextRound > 0 {
-		if catchpoint != "" {
-			round, _, err := ledgercore.ParseCatchpointLabel(catchpoint)
+func InitializeLedger(ctx context.Context, logger *log.Logger, nextDbRound uint64, genesis bookkeeping.Genesis, opts *processors.BlockProcessorConfig) error {
```
winder, line 24: nit: rename `opts` to `config`?

processors/metadata.go
```diff
@@ -0,0 +1,30 @@
...
+	ImplementationName        string
+	ImplementationDescription string
+	ImplementationDeprecated  bool
```
winder, line 9: This interface doesn't really make sense to me. Why are there getters if the fields are public? I can understand the getters if we want to make them immutable, but not otherwise. I'd consider two changes:
1) Remove `Implementation` prefix, and make the fields private.
2) You'd need a `MakeProcessorMetadata` function now to set the private fields.

processors/processor_factory.go
```diff
@@ -0,0 +1,49 @@
+package processors
+
+import (
+	"context"
+	"fmt"
+	"github.com/algorand/indexer/data"
+
+	"github.com/algorand/indexer/plugins"
+	"github.com/sirupsen/logrus"
```
winder, line 9: nit
```suggestion
	"github.com/sirupsen/logrus"

	"github.com/algorand/indexer/data"
	"github.com/algorand/indexer/plugins"
```

processors/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,367 @@
...
+	resources, err := l.LookupResources(resourcesReq)
+	if err != nil {
+		return nil, nil, fmt.Errorf("prepareAccountsResources() err: %w", err)
+	}
+
+	return accounts, resources, nil
+}
+
+// MakeLegacyProcessorHandlerFunction makes a legacy function that emulates original behavior of block processor
+func MakeLegacyProcessorHandlerFunction(proc *BlockProcessor, handler func(block *ledgercore.ValidatedBlock) error) func(cert *rpcs.EncodedBlockCert) error {
```
winder, line 349: nit: Consider renaming to use `Adapter` instead of `Legacy`. Something like `MakeBlockProcessorHandlerAdapter`

data/block_export_data.go
```diff
@@ -29,6 +38,75 @@ type BlockData struct {
...
+
+// MakeBlockDataFromBlock makes BlockData from boookeeping.Block
+func MakeBlockDataFromBlock(base BlockData, input bookkeeping.Block) BlockData {
+	return BlockData{
+		BlockHeader: input.BlockHeader,
+		Payset:      input.Payset,
+		Delta:       base.Delta,
+		Certificate: base.Certificate,
+	}
+}
```
winder, line 76: These seem mostly unnecessary, why not just create the struct normally?

processors/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,367 @@
...
+			Payset:      payset,
+		}
+		vb = ledgercore.MakeValidatedBlock(block, delta)
+		return vb, payset, nil
+	}
+	vb = ledgercore.MakeValidatedBlock(blockCert.Block, delta)
+	return vb, blockCert.Block.Payset, nil
+}
+
+func (proc *blockProcessor) addValidatedBlockCertToLedger(vb ledgercore.ValidatedBlock, cert agreement.Certificate, roundToWait basics.Round) error {
```
winder, line 174: nit: merge with `OnComplete`, seems like it no longer needs to be a separate function.

processors/blockprocessor/block_processor.go
```diff
@@ -0,0 +1,367 @@
...
+		return fmt.Errorf("could not make ledger: %w", err)
+	}
+	if l == nil {
+		return fmt.Errorf("ledger was created with nil pointer")
+	}
+
+	nextRound := initProvider.NextDBRound()
+	if uint64(l.Latest()) > uint64(nextRound) {
+		return fmt.Errorf("the ledger cache is ahead of the required round (%d > %d) and must be re-initialized", l.Latest(), nextRound)
+	}
```
winder, line 105: nit: you already have the `round` variable.
```suggestion
	if uint64(l.Latest()) > uint64(round) {
		return fmt.Errorf("the ledger cache is ahead of the required round (%d > %d) and must be re-initialized", l.Latest(), round)
	}
```

data/block_export_data.go
AlgoStephenAkiki: I think because it allows people to "merge" previous block data variables with new inputs easily.  I figured if we are building a pipeline that "adding" to the previous block data would be a common operation.

winder: That use case makes sense, how about putting them on the `BlockData` type?

i.e. one of these:
`func (bd *BlockData) UpdateBlock(block bookkeeping.Block)`
`func (bd BlockData) UpdateBlock(block bookkeeping.Block) BlockData`

processors/blockprocessor/block_processor.go
AlgoStephenAkiki: I think it does since it is used in `OnComplete` and in `Init`

winder: `Init` would call `OnComplete` instead

processors/blockprocessor/initialize.go
```diff
@@ -60,7 +59,7 @@ func InitializeLedgerFastCatchup(ctx context.Context, logger *log.Logger, catchp
 
 // InitializeLedgerSimple initializes a ledger with the block processor by
 // sending it one block at a time and letting it update the ledger as usual.
-func InitializeLedgerSimple(ctx context.Context, logger *log.Logger, round uint64, genesis *bookkeeping.Genesis, opts *idb.IndexerDbOptions) error {
+func InitializeLedgerSimple(ctx context.Context, logger *log.Logger, round uint64, genesis *bookkeeping.Genesis, opts *processors.BlockProcessorConfig) error {
```
winder, line 62: It looks like there are two more functions in this file where you can rename `opts` -> `config`. Maybe other files also?

winder (approved):
Try to get the remaining `opts` -> `config` parameter names, otherwise LGTM!

PR #1141, opened 2022-07-22
Don't lookup big foreign assets.

winder:
## Summary

We missed the common code path in the previous hotfix, this is the same fix as before but we avoid adding these assets and apps to begin with.

### DIFF
--- a/accounting/eval_preload.go
+++ b/accounting/eval_preload.go
@@ -1,6 +1,8 @@
 package accounting
 
 import (
+	"math"
+
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger"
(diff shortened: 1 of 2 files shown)


cce (approved):
nice! two birds with one stone.

accounting/eval_preload.go
```diff
@@ -34,9 +36,15 @@ func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD, assetsReq map[ba
 			appsReq[fields.ApplicationID] = struct{}{}
 		}
 		for _, index := range fields.ForeignApps {
+			if index > basics.AppIndex(math.MaxInt64) {
```
shiqizng, line 39: why do we use math.MaxInt64 instead of math.MaxUint64?

winder: This is related to the `BIGINT` -> `NUMERIC(20)` issue, algod/sqlite has a similar problem that this works around.

algorandskiy (approved):

winder:
@algobarb helped me run this on the stuck testnet node, it seems to be working.

PR #1143, opened 2022-07-25
Enhancement: update indexer table schema

shiqizng:
## Summary
This PR updates the schema for db tables to use numeric(20) type for columns that could have values larger than bigint type. 

## Test Plan
Add a unit test for the migration, run indexer daemon to check migration happened, and check db to validate that table schema has been updated.

### DIFF
--- a/idb/postgres/internal/schema/bigint_schema.go
+++ b/idb/postgres/internal/schema/bigint_schema.go
@@ -1,5 +1,6 @@
 package schema
 
+// BigIntSchema is the schema used before bigint to numeric(20) change
 const BigIntSchema = `-- This file is setup_postgres.sql which gets compiled into go source using a go:generate statement in postgres.go
 --
 -- TODO? replace all 'addr bytea' with 'addr_id bigint' and a mapping table? makes addrs an 8 byte int that fits in a register instead of a 32 byte string
--- a/idb/postgres/internal/schema/bigint_schema.go
+++ b/idb/postgres/internal/schema/bigint_schema.go
@@ -1,0 +1,122 @@
+package schema
+
+const BigIntSchema = `-- This file is setup_postgres.sql which gets compiled into go source using a go:generate statement in postgres.go
+--
+-- TODO? replace all 'addr bytea' with 'addr_id bigint' and a mapping table? makes addrs an 8 byte int that fits in a register instead of a 32 byte string
+
+CREATE TABLE IF NOT EXISTS block_header (
+  round bigint PRIMARY KEY,
+  realtime timestamp without time zone NOT NULL,
+  rewardslevel bigint NOT NULL,
+  header jsonb NOT NULL
+);
+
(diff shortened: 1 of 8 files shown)


idb/postgres/internal/schema/setup_postgres.sql
```diff
@@ -3,7 +3,7 @@
 -- TODO? replace all 'addr bytea' with 'addr_id bigint' and a mapping table? makes addrs an 8 byte int that fits in a register instead of a 32 byte string
 
 CREATE TABLE IF NOT EXISTS block_header (
-  round bigint PRIMARY KEY,
+  round numeric(20) PRIMARY KEY,
```
Eric-Warehime, line 6: What is the rationale behind choosing 20 as opposed to another number?

idb/postgres/postgres_migrations_test.go
```diff
@@ -24,11 +24,44 @@ func TestConvertAccountDataIncrementsMigrationNumber(t *testing.T) {
...
+	err = convertAccountData(&db, &migrationState)
 	require.NoError(t, err)
 
 	migrationState, err = db.getMigrationState(context.Background(), nil)
 	require.NoError(t, err)
 
 	assert.Equal(t, types.MigrationState{NextMigration: 6}, migrationState)
 }
+
+func TestConvertBigIntType(t *testing.T) {
```
Eric-Warehime, line 36: Not sure if this is necessary, but should we add a test case that asserts the existing values aren't changed by the type conversion?

i.e. insert round 1,000,000 convert the type and assert that the round is still 1,000,000?

Eric-Warehime (commented):
It looks like I messed up your PR by merging in ps://github.com/algorand/indexer/pull/1144 so you'll have conflicts.

idb/postgres/internal/schema/setup_postgres.sql
shiqizng: this is because uint64 max is 18446744073709551615 and it's a value of 20 digits.

idb/postgres/postgres_migrations_test.go
shiqizng: good point. I think this can be covered by running the block validator.

idb/postgres/internal/schema/setup_postgres.sql
Eric-Warehime: Got it thanks!

Eric-Warehime (approved):

idb/postgres/internal/schema/bigint_schema.go
```diff
@@ -0,0 +1,123 @@
+package schema
```
Eric-Warehime, line 1: Is there a way we can name this file `test` something or otherwise mark the fact that it's not the actual current production schema?

It doesn't seem like we're maintaining old versions of the schema for past updates. What was our process for previous updates regarding schema testing?

shiqizng:
this PR is replaced by https://github.com/algorand/indexer/pull/1166

PR #1144, opened 2022-07-25
test fixes: Submodule updates

Eric-Warehime:
## Summary

A new field in Teal has shown up in the e2e test files which means that we need to update the submodule. However, because the submodule update also pulls in the 320 round changes, there are some test failures showing up.

The root cause/fix for these is that each transaction submitted to a ledger needs to have a different checksum. In order to accomplish this I've put a "random" byte array into the notes field of each transaction created, and have re-created transactions instead of reusing them in our tests.

I've also added a new LedgerForEvaluator method since that was added the upstream interface as well. It is a very simple wrapper method.

## Test Plan

CI

### DIFF
--- a/processor/eval/ledger_for_evaluator.go
+++ b/processor/eval/ledger_for_evaluator.go
@@ -149,6 +149,7 @@ func (l LedgerForEvaluator) LatestTotals() (ledgercore.AccountTotals, error) {
 	return totals, err
 }
 
+// BlockHdrCached is part of go-algorand's indexerLedgerForEval interface.
 func (l LedgerForEvaluator) BlockHdrCached(round basics.Round) (bookkeeping.BlockHeader, error) {
 	return l.Ledger.BlockHdrCached(round)
 }
--- a/processor/eval/ledger_for_evaluator.go
+++ b/processor/eval/ledger_for_evaluator.go
@@ -148,3 +148,7 @@ func (l LedgerForEvaluator) LatestTotals() (ledgercore.AccountTotals, error) {
 	_, totals, err := l.Ledger.LatestTotals()
 	return totals, err
 }
+
+func (l LedgerForEvaluator) BlockHdrCached(round basics.Round) (bookkeeping.BlockHeader, error) {
+	return l.Ledger.BlockHdrCached(round)
+}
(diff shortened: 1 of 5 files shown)


shiqizng (approved):

util/test/account_testutil.go
```diff
@@ -602,3 +620,11 @@ func MakeGenesisBlock() bookkeeping.Block {
 	}
 	return genesisBlock
 }
+
+// ArbitraryString should be used to generate a pseudo-random string to put in the Note field of a Txn Header.
+// This is necessary to ensure the hash of any two txns used in tests are never the same.
+func ArbitraryString() []byte {
+	arb := make([]byte, config.MaxTxnNoteBytes)
+	rand.Read(arb)
+	return arb
```
tzaffi, line 629: 👍

processor/eval/ledger_for_evaluator.go
```diff
@@ -148,3 +148,8 @@ func (l LedgerForEvaluator) LatestTotals() (ledgercore.AccountTotals, error) {
 	_, totals, err := l.Ledger.LatestTotals()
 	return totals, err
 }
+
+// BlockHdrCached is part of go-algorand's indexerLedgerForEval interface.
```
tzaffi, line 152: I don't see where else in the PR this function is used. Was this brought in by the `go-algorand` update? Is there a way to know that this is implemented correctly (seems like it should since this is a pass-through) ?

tzaffi: I should also mention that if indexer is not actually making use of this method, I would argue for implementing via `panic("not implemented")` instead.

tzaffi (commented):
The changes all make sense, but I had a question regarding the `BlockHdrCached` method in `processor/eval/ledger_for_evaluator.go`.

processor/eval/ledger_for_evaluator.go
Eric-Warehime: Yeah it was added in https://github.com/algorand/go-algorand/commit/b8007e4ade943d4bc181a301d1291685eaf60fc4#diff-69a9375fb82e9e6878855f6e962f7b3eeba8bf108dc6c2c07ab21e35cfb8d267R47

And no we don't call it in our code, but it has to be defined as part of the interface, so I just copied what I saw in the above commit.

Eric-Warehime: I'm actually not confident that it's not used. It seems to be called inside `eval` in b8007e4ade943d4bc181a301d1291685eaf60fc4 and as you've pointed out, there wouldn't seem to be a reason to add it to the interface if it wasn't being used.

tzaffi (approved):
LGTM and 🤞 that CI will agree

PR #1147, opened 2022-07-26
Conduit: Exporter daemon integration

Eric-Warehime:
## Summary

The goal of this PR is to integrate the new Exporter code with the existing Indexer daemon. In this way we can enable users to create and plug new data destinations into the existing Indexer framework.
 
Replaces #1134 since there have been a lot of changes which are easier to just branch off of instead of merge into the old PR.

## Test Plan

Added tests to cover new code.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -4,6 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/algorand/indexer/plugins"
+	"gopkg.in/yaml.v3"
+	"io/fs"
+	"io/ioutil"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -23,6 +27,9 @@ import (
 	"github.com/algorand/indexer/api"
 	"github.com/algorand/indexer/api/generated/v2"
 	"github.com/algorand/indexer/config"
+	"github.com/algorand/indexer/exporters"
+	_ "github.com/algorand/indexer/exporters/noop"
+	"github.com/algorand/indexer/exporters/postgresql"
 	"github.com/algorand/indexer/fetcher"
 	"github.com/algorand/indexer/idb"
 	"github.com/algorand/indexer/importer"
@@ -66,6 +73,7 @@ type daemonConfig struct {
 	configFile                string
 	suppliedAPIConfigFile     string
 	genesisJSONPath           string
+	exporterPlugin            string
 }
 
 // DaemonCmd creates the main cobra command, initializes flags, and viper aliases
@@ -118,6 +126,7 @@ func DaemonCmd() *cobra.Command {
 	cfg.flags.StringVarP(&cfg.cpuProfile, "cpuprofile", "", "", "file to record cpu profile to")
 	cfg.flags.StringVarP(&cfg.pidFilePath, "pidfile", "", "", "file to write daemon's process id to")
 	cfg.flags.StringVarP(&cfg.configFile, "configfile", "c", "", "file path to configuration file (indexer.yml)")
+	cfg.flags.StringVarP(&cfg.exporterPlugin, "exporter", "", postgresql.ExporterName, "name of the exporter plugin to use")
 	viper.RegisterAlias("algod", "algod-data-dir")
 	viper.RegisterAlias("algod-net", "algod-address")
 	viper.RegisterAlias("server", "server-address")
@@ -315,9 +324,9 @@ func runDaemon(daemonConfig *daemonConfig) error {
 		daemonConfig.noAlgod = true
 	}
 	opts := idb.IndexerDbOptions{}
-	if daemonConfig.noAlgod && !daemonConfig.allowMigration {
-		opts.ReadOnly = true
-	}
+	// if daemonConfig.noAlgod && !daemonConfig.allowMigration {
+	opts.ReadOnly = true
+	// }
 
 	opts.MaxConn = daemonConfig.maxConn
 	opts.IndexerDatadir = daemonConfig.indexerDataDir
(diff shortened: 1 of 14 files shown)


exporters/noop/noop_exporter_test.go
```diff
@@ -16,18 +16,18 @@ var ne = nc.New()
...
 
 func TestExporterMetadata(t *testing.T) {
 	meta := ne.Metadata()
 	assert.Equal(t, plugins.PluginType(plugins.Exporter), meta.Type())
-	assert.Equal(t, noopExporterMetadata.ExpName, meta.Name())
-	assert.Equal(t, noopExporterMetadata.ExpDescription, meta.Description())
-	assert.Equal(t, noopExporterMetadata.ExpDeprecated, meta.Deprecated())
+	assert.Equal(t, Metadata.ExpName, meta.Name())
+	assert.Equal(t, Metadata.ExpDescription, meta.Description())
+	assert.Equal(t, Metadata.ExpDeprecated, meta.Deprecated())
```
winder, line 30: Seems like we could use the same Metadata object in Export/Importer until we need them to diverge.

processor/processor.go
```diff
@@ -1,13 +1,12 @@
...
-	"github.com/algorand/go-algorand/ledger/ledgercore"
 	"github.com/algorand/go-algorand/rpcs"
 )
 
 // Processor is the block processor interface
 type Processor interface {
 	Process(cert *rpcs.EncodedBlockCert) error
-	SetHandler(handler func(block *ledgercore.ValidatedBlock) error)
 	NextRoundToProcess() uint64
+	AddGenesisBlock() error
```
winder, line 11: Is this going away in Stephen's PR? Seems like a weird method.

winder (commented):
I'm not sure how useful this feedback is since we have a lot of conflicts with #1140.

cmd/algorand-indexer/daemon.go
```diff
@@ -315,61 +324,86 @@ func runDaemon(daemonConfig *daemonConfig) error {
...
 }
 
-func runBlockImporter(ctx context.Context, cfg *daemonConfig, wg *sync.WaitGroup, db idb.IndexerDb, dbAvailable chan struct{}, bot fetcher.Fetcher, opts idb.IndexerDbOptions) {
+// Bridges functionality between setting config via CLI and the new plugin configs
+func writeCLIToPlugin(cfg *daemonConfig) {
+	switch cfg.exporterPlugin {
+	case postgresql.Metadata.Name():
+		pluginCfg := plugins.LoadConfig(logger, cfg.indexerDataDir, postgresql.Metadata)
+		var expCfg postgresql.ExporterConfig
+		err := yaml.Unmarshal([]byte(pluginCfg), &expCfg)
```
shiqizng, line 373: nit: maybe use this pattern
```suggestion
		if err := yaml.Unmarshal([]byte(pluginCfg), &expCfg); err != nil {
			maybeFail(err, "unable to load postgresql config: %v", err)
		}
```

exporters/postgresql/postgresql_exporter.go
```diff
@@ -13,7 +13,8 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const exporterName = "postgresql"
+// ExporterName constant for the postgresql Exporter module
+const ExporterName = "postgresql"
```
shiqizng, line 17: why does this need to be exported?

processor/blockprocessor/block_processor.go
```diff
@@ -3,146 +3,64 @@ package blockprocessor
...
 	if l == nil {
 		return nil, fmt.Errorf("MakeProcessorWithLedger() err: local ledger not initialized")
 	}
-	err := addGenesisBlock(l, handler)
+	var proc processor.Processor
+	if exp, ok := genericHandler.(exporters.Exporter); ok {
+		proc = &ledgerExporterProcessor{logger: logger, ledger: l, exporter: exp}
+	} else if handler, ok := genericHandler.(func(block *ledgercore.ValidatedBlock) error); ok {
+		proc = &blockProcessor{logger: logger, ledger: l, handler: handler}
+	} else if genericHandler == nil {
```
shiqizng, line 32: I think if genericHandler == nil the if statements before this one will fail with an error? I don't believe we are allowed to case nil to a type.

processor/blockprocessor/exporter_processor.go
```diff
@@ -0,0 +1,111 @@
...
+		})
+		if err != nil {
+			return fmt.Errorf("addGenesisBlock() handler err: %w", err)
+		}
+	}
+	return nil
+}
+
+// Process a raw algod block
+func (proc *ledgerExporterProcessor) Process(blockCert *rpcs.EncodedBlockCert) error {
```
shiqizng, line 43: it looks like the same logic is already ran in [block_processor,](https://github.com/algorand/indexer/blob/45d5c4b28523098696f438d98602ad411f84832e/processors/blockprocessor/block_processor.go#L127) why do we need this in the exporter again?

processor/blockprocessor/handler_processor.go
```diff
@@ -0,0 +1,95 @@
...
+		err = proc.handler(&vb)
+		if err != nil {
+			return fmt.Errorf("addGenesisBlock() handler err: %w", err)
+		}
+	}
+	return nil
+}
+
+// Process a raw algod block
+func (proc *blockProcessor) Process(blockCert *rpcs.EncodedBlockCert) error {
```
shiqizng, line 36: I may be missing something. why do I seem to see similar Process logic repeated in multiple places?

cmd/algorand-indexer/daemon.go
```diff
@@ -315,61 +324,86 @@ func runDaemon(daemonConfig *daemonConfig) error {
...
 	}
 
-	fmt.Printf("serving on %s\n", daemonConfig.daemonServerAddr)
-	logger.Infof("serving on %s", daemonConfig.daemonServerAddr)
-
-	options := makeOptions(daemonConfig)
-
-	api.Serve(ctx, daemonConfig.daemonServerAddr, db, bot, logger, options)
+	// Only serve the API for the postgres exporter
+	if daemonConfig.exporterPlugin == postgresql.ExporterName {
```
shiqizng, line 353: is Postgres or postGres considered different from postgres?

cmd/algorand-indexer/daemon.go
```diff
@@ -315,61 +324,86 @@ func runDaemon(daemonConfig *daemonConfig) error {
...
+		expCfg.ConnectionString = postgresAddr
+		data, err := yaml.Marshal(expCfg)
+		if err != nil {
+			maybeFail(err, "failure setting postgresql config data: %v", err)
+		}
+		err = ioutil.WriteFile(plugins.PluginConfigPath(cfg.indexerDataDir, postgresql.Metadata), []byte(data), fs.ModePerm)
+		if err != nil {
+			maybeFail(err, "failure writing updated options to postgresql exporter config file: %v", err)
+		}
+	default:
```
shiqizng, line 388: maybe in default case write to logger that no matching cfg.exporterPlugin was found?

Eric-Warehime:
Cancelling this PR--I've taken a look at all of the feedback and will use it to inform how I tackle the integration between conduit and indexer's daemon. But based on the fact that we have created a new conduit binary on our feature branch, I don't think it makes sense to implement plugin types as individual features in daemon instead of modeling the indexer as a conduit pipeline with an API wrapper.

PR #1152, opened 2022-07-29
Testing: Use tempdir instead of /tmp for e2elive test

winder:
## Summary

When running locally there were test artifacts left in /tmp which caused errors when running the test more than once.

Additionally, increase the lookback for the e2e archive and print more information when a file was not found in S3.

PR #1153, opened 2022-07-29
Enhancement: Format misc/*.py with `black`

tzaffi:
# Summary

This is a NOOP change that only applies `black` to python files under `misc`. I only found one other python file in the repo `api/jsoncanon.py` and that didn't change.

## Note regarding merge with #1152

Here were the changes and their current state:
* [MaxKeys=50](https://github.com/algorand/indexer/blob/8cfb0da21a22ec4aff30d2d2c3893d421714d86c/misc/util.py#L161)
* [logger.warning()](https://github.com/algorand/indexer/blob/8cfb0da21a22ec4aff30d2d2c3893d421714d86c/misc/util.py#L175)
* [tempdir instead of /tmp](https://github.com/algorand/indexer/blob/8cfb0da21a22ec4aff30d2d2c3893d421714d86c/misc/e2elive.py#L92)

## Test Plan

CI

### DIFF
--- a/misc/blockarchiver.py
+++ b/misc/blockarchiver.py
@@ -27,17 +27,18 @@ logger = logging.getLogger(__name__)
 # b=algod.block_info(algod.status()['lastRound'])
 # print(json.dumps(b, indent=2))
 def token_addr_from_algod(algorand_data):
-    addr = open(os.path.join(algorand_data, 'algod.net'), 'rt').read().strip()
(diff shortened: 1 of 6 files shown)


tzaffi:
put in draft mode as I expect to pull in #1152 which will fix the CI failure

winder (approved):

PR #1166, opened 2022-08-05
Enhancement: add max int64 checks

shiqizng:
## Summary

This PR adds max int64 checks to round, application ID and asset ID in API requests.
 
## Test Plan
Add unit tests for the handler and filter methods.

### DIFF
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"math"
 	"sort"
 	"strconv"
 	"strings"
@@ -588,6 +589,10 @@ func (si *ServerImplementation) transactionParamsToTransactionFilter(params gene
 	filter.ApplicationID = uintOrDefault(params.ApplicationId)
 	filter.Limit = min(uintOrDefaultValue(params.Limit, si.opts.DefaultTransactionsLimit), si.opts.MaxTransactionsLimit)
 	filter.Round = params.Round
+	// Integer > math.Int64
+	if filter.AssetID > math.MaxInt64 || filter.ApplicationID > math.MaxInt64 || (filter.Round != nil && uint64(*filter.Round) > math.MaxInt64) {
+		errorArr = append(errorArr, errValueExceedingInt64)
+	}
 
 	// String
 	filter.AddressRole, errorArr = decodeAddressRole(params.AddressRole, params.ExcludeCloseTo, errorArr)
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"math"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


api/handlers.go
```diff
@@ -159,6 +166,9 @@ func (si *ServerImplementation) verifyHandler(operationID string, ctx echo.Conte
 // LookupAccountByID queries indexer for a given account.
 // (GET /v2/accounts/{account-id})
 func (si *ServerImplementation) LookupAccountByID(ctx echo.Context, accountID string, params generated.LookupAccountByIDParams) error {
+	if params.Round != nil && uint64(*params.Round) > math.MaxInt64 {
+		return badRequest(ctx, errValueExceedingInt64)
+	}
 	if err := si.verifyHandler("LookupAccountByID", ctx); err != nil {
 		return badRequest(ctx, err.Error())
 	}
```
winder, line 174: Please make sure all of these happen after `verifyHandler`. If the endpoint is disabled we should make sure that is the returned error

winder: What do you think about returning `404` errors instead of bad request?

api/converter_utils.go
```diff
@@ -588,6 +589,10 @@ func (si *ServerImplementation) transactionParamsToTransactionFilter(params gene
 	filter.ApplicationID = uintOrDefault(params.ApplicationId)
 	filter.Limit = min(uintOrDefaultValue(params.Limit, si.opts.DefaultTransactionsLimit), si.opts.MaxTransactionsLimit)
 	filter.Round = params.Round
+	// Integer > math.Int64
+	if filter.AssetID > math.MaxInt64 || filter.ApplicationID > math.MaxInt64 || (filter.Round != nil && uint64(*filter.Round) > math.MaxInt64) {
```
Eric-Warehime, line 593: I see that there are a lot more fields in a filter which are `uint64` https://github.com/algorand/indexer/blob/develop/idb/idb.go#L196 

Why are these the only ones we need to do checks for as opposed to checking all of the supplied filters?

api/handlers.go
shiqizng: I think 404 makes sense. I'll update.

api/converter_utils.go
shiqizng: I moved these checks to validateTransactionFilter and added checks for all uint64 fields except for AssetAmount which is type numeric(20) in the DB table.

api/handlers.go
```diff
@@ -51,6 +52,32 @@ type ServerImplementation struct {
...
+	}
+
+	// algos > math.MaxInt64
+	if (filter.AlgosLT != nil && *filter.AlgosLT > math.MaxInt64) ||
+		(filter.AlgosGT != nil && *filter.AlgosGT > math.MaxInt64) {
+		errorArr = append(errorArr, errValueExceedingInt64)
+	}
+
+	// effectiveAmount > math.MaxInt64
+	if (filter.EffectiveAmountLT != nil && *filter.EffectiveAmountLT > math.MaxInt64) ||
```
shiqizng, line 76: Should I update these from *uint64 to *int64 to match the table schema? I think we remove these checks if we update the var types.

idb/postgres/postgres.go
```diff
@@ -582,12 +582,12 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 		partNumber++
 	}
 	if tf.AssetAmountGT != nil {
-		whereParts = append(whereParts, fmt.Sprintf("(t.txn -> 'txn' -> 'aamt')::bigint > $%d", partNumber))
+		whereParts = append(whereParts, fmt.Sprintf("(t.txn -> 'txn' -> 'aamt')::numeric(20) > $%d", partNumber))
```
winder, line 585: nice catch, did this cause a failure before?

winder (approved):

idb/postgres/postgres.go
shiqizng: it was causing a failure for a test case I just added that has a txn with asset amount maxUint64.

Eric-Warehime (approved):

PR #1167, opened 2022-08-05
1165 create initial conduit binary

### LINKED ISSUE #1165: Create Initial Conduit Binary
Create an initial Conduit Binary along with appropriate RFC documentation.

AlgoStephenAkiki:
Resolves #1165 

Creates an initial framework for conduit binary production.  Introduces
a data directory flag as well as init and shutdown (basic)
functionality.

Instructions:

1. Type `make` in project root.  Conduit binary should now be located in `cmd/conduit`
2. Run from shell: `mkdir -p /tmp/conduit_data_dir && touch /tmp/conduit_data_dir/conduit.yml`
3. Insert the following into `/tmp/conduit_data_dir/conduit.yml`:
```
LogLevel: info
Importer:
  Name: "algod"
  Config:
    netaddr: "http://127.0.0.1:8080"
    token: "e36c01fc77e490f23e61899c0c22c6390d0fff1443af2c95d056dc5ce4e61302"
Processors:
  - Name: "noop"
    Config:
      catchpoint: "7560000#3OUX3TLXZNOK6YJXGETKRRV2MHMILF5CCIVZUOJCT6SLY5H2WWTQ"
Exporter:
  Name: "noop"
  Config:
    connectionstring: ""
```
4. start `algod` instance
5. Run `cmd/conduit/conduit -d /tmp/conduit_data_dir/`


conduit/errors.go
```diff
@@ -0,0 +1,11 @@
+package conduit
+
+import "fmt"
+
+// CriticalError an error that causes the entire conduit pipeline to
+// stop
+type CriticalError struct{}
```
Eric-Warehime, line 7: This looks unused as of now.

exporters/noop/noop_exporter.go
```diff
@@ -17,6 +17,11 @@ type noopExporter struct {
 	cfg   plugins.PluginConfig
 }
 
+// package-wide init function
+func init() {
+	exporters.RegisterExporter("noop", &Constructor{})
+}
+
```
Eric-Warehime, line 24: This is already being called at the bottom of the file--line 66/71.

conduit/init_provider.go
```diff
@@ -0,0 +1,26 @@
+package conduit
+
+import (
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+)
+
+// AlgodInitProvider algod based init provider
+type AlgodInitProvider struct {
```
Eric-Warehime, line 9: I don't understand which part of this is specific to Algod

conduit/pipeline.go
```diff
@@ -0,0 +1,242 @@
...
+		return fmt.Errorf("conduit configuration was nil")
+	}
+
+	if len(cfg.Importer) == 0 {
+		return fmt.Errorf("importer configuration was empty")
+	}
+
+	if len(cfg.Processors) == 0 {
+		return fmt.Errorf("processor configuration was empty")
+	}
```
Eric-Warehime, line 42: I think 0 processors could be a valid pipeline that we might want to support.

conduit/pipeline.go
```diff
@@ -0,0 +1,242 @@
...
+	Exporter   map[string]interface{}   `yaml:"Exporter"`
+}
+
+// Valid validates pipeline config
+func (cfg *PipelineConfig) Valid() error {
+	if cfg.ConduitConfig == nil {
+		return fmt.Errorf("conduit configuration was nil")
+	}
+
+	if len(cfg.Importer) == 0 {
```
Eric-Warehime, line 36:
```suggestion
	if len(cfg.Importer) != 1 {
```

conduit/pipeline.go
```diff
@@ -0,0 +1,242 @@
...
+		return fmt.Errorf("importer configuration was empty")
+	}
+
+	if len(cfg.Processors) == 0 {
+		return fmt.Errorf("processor configuration was empty")
+	}
+
+	if len(cfg.Exporter) == 0 {
```
Eric-Warehime, line 44:
```suggestion
	if len(cfg.Exporter) != 1 {
```

data/block_export_data.go
```diff
@@ -20,6 +20,7 @@ type RoundProvider interface {
 type InitProvider interface {
 	Genesis() *bookkeeping.Genesis
 	NextDBRound() basics.Round
+	AdvanceDBRound()
```
Eric-Warehime, line 23: Where is this getting called?

Eric-Warehime (commented):
Nice, I think it's a great start! Left some comments/questions--hope they help.

conduit/errors.go
AlgoStephenAkiki: Yeah, this is just here for future development.

conduit/init_provider.go
AlgoStephenAkiki: This would essentially have a client handle to algod since Will told me that most plugin developers won't have a location for the genesis file.  So here we would simply grab the genesis file from an algod instance.

data/block_export_data.go
AlgoStephenAkiki: This would be called at the end of the pipeline so that the init provider would know when to increase the database round we are working on.  Right now the NextDBRound only gets the current round (maybe we should prefix it with "Get"), so we would need something to advance it.

conduit/init_provider.go
Eric-Warehime: In that case I think we should just integrate this into the Importer plugin. Wherever we're fetching blocks from is going to have to have the genesis info so I think it makes sense to require it to be able to provide that as well as a block for a given round.

conduit/pipeline.go
AlgoStephenAkiki: Isn't this the opposite logic though?  We are trying to see if there are no importers and then throw an error.

conduit/pipeline.go
AlgoStephenAkiki: See comment above on importers

conduit/pipeline.go
Eric-Warehime: It's just a matter of what we want to consider a valid pipeline. I think requiring a single importer and Exporter makes sense, and having 0 or more Processors is also valid. So we want to check that they have exactly 1 of Importer/Exporter, and we don't need to validate length of processors because any number is valid.

Eric-Warehime: Discussed out of band--the existing logic is what we want.

conduit/logging.go
```diff
@@ -0,0 +1,20 @@
...
+	Formatter *log.JSONFormatter
+	Type      string
+	Name      string
+}
+
+// Format allows this to be used as a logrus formatter
+func (f PluginLogFormatter) Format(entry *log.Entry) ([]byte, error) {
+	// Underscores force these to be in the front in order type -> name
+	entry.Data["__type"] = f.Type
+	entry.Data["_name"] = f.Name
```
Eric-Warehime, line 18: Why use 1 underscore in one key and 2 in the other?

Eric-Warehime (commented):
LGTM for a first pass. Let's continue the discussion around moving Genesis data to the Importer interface. Will also be good to get Will's opinion on this when he has time.

conduit/logging.go
AlgoStephenAkiki: It's to force the order when outputting the json to the screen to be type then name.  Like this:

`{"__type":"Conduit","_name":"main","level":"info","msg":"Log level set to: warn","time":"2022-08-09T08:48:44-04:00"}`

If one is looking at the logs manually it is a little easier to look at.

cmd/conduit/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/algorand/indexer/conduit"
+	"github.com/algorand/indexer/config"
+)
+
+import (
+	_ "github.com/algorand/indexer/exporters/all"
+	_ "github.com/algorand/indexer/importers/all"
+	_ "github.com/algorand/indexer/processors/all"
+)
```
Eric-Warehime, line 18: Nit: I think these can all be in the main import block. If you're going to separate them at least put a comment saying why they're important and not to remove them.

conduit/init_provider.go
```diff
@@ -0,0 +1,27 @@
...
+// PipelineInitProvider algod based init provider
+type PipelineInitProvider struct {
+	currentRound basics.Round
+	genesis      *bookkeeping.Genesis
+}
+
+// AdvanceDBRound advances the database round
+func (a *PipelineInitProvider) AdvanceDBRound() {
+	a.currentRound = a.currentRound + 1
+}
```
Eric-Warehime, line 17: I'm kind of skeptical that this is a good way to manage round state in the pipeline, but I'm interested to see what you have in mind in the next PR.

importers/algod/algod_importer_test.go
```diff
@@ -37,18 +38,26 @@ func init() {
...
+			var blockbytes []byte
+			w.WriteHeader(http.StatusOK)
+			response := struct {
+				Block bookkeeping.Block `codec:"block"`
+			}{
+				Block: blk.Block,
+			}
+			enc := codec.NewEncoderBytes(&blockbytes, protocol.CodecHandle)
+			enc.Encode(response)
+			w.Write(blockbytes)
```
Eric-Warehime, line 59: Nit: I think we could simplify this with something like the following, but I haven't tested it.
```suggestion
			rnd, _ := strconv.Atoi(path.Base(r.URL.Path))
			blk := rpcs.EncodedBlockCert{Block: bookkeeping.Block{BlockHeader: bookkeeping.BlockHeader{Round: basics.Round(rnd)}}}
			blockbytes := protocol.EncodeJSON(*blk)
			w.WriteHeader(http.StatusOK)
			w.Write(blockbytes)
```

importers/algod/algod_importer.go
```diff
@@ -36,31 +38,66 @@ func New() importers.Importer {
...
 	client, err = algod.MakeClient(algodImp.cfg.NetAddr, algodImp.cfg.Token)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	algodImp.aclient = client
-	return err
+
+	// For testing check against nil
+	if !netAddrIsEmpty {
```
Eric-Warehime, line 83: Are we sure we want to return a `nil` error when the algod address is empty?

Eric-Warehime (approved):
I'd like to merge just to keep development moving forward. Left some comments for some things.

Also, we need to add `conduit/pipeline_test.go`, `conduit/config_test.go`, and probably `processors/noop/noop_processor_test.go` (even though I know it does nothing).

importers/algod/algod_importer.go
AlgoStephenAkiki: How would we indicate that we don't want to retrieve a genesis file for testing purposes?

Eric-Warehime: Use the mock algod server for the init test in the same way that we do for the get block test.

It's not a huge issue since an empty address in your algod client will fail pretty quickly when trying to retrieve a block, but we can detect here that it's not going to work if it has no address so it would maybe be easier to debug if we caught it here.

cmd/conduit/.gitignore
```diff
@@ -0,0 +1 @@
+conduit
```
winder, line 1: could you move this to the root .gitignore?

cmd/conduit/main.go
```diff
@@ -0,0 +1,112 @@
...
+func conduitCmd() *cobra.Command {
+	cfg := &conduit.Config{}
+	conduitCmd := &cobra.Command{
+		Use:   "conduit",
+		Short: "run the conduit framework",
+		Long:  "run the conduit framework",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// Tie cobra variables into local go-lang variables
```
winder, line 54: Are we sure we want to use environment variables? How does that work if there is more than one level in the YAML file?

cmd/conduit/main.go
```diff
@@ -0,0 +1,112 @@
...
+				return err
+			}
+
+			pCfg, err := conduit.MakePipelineConfig(logger, cfg)
+
+			if err != nil {
+				return err
+			}
+
+			logLevel, _ := log.ParseLevel(pCfg.PipelineLogLevel)
```
winder, line 68: I see where you're going with the `Valid()` pattern and ignoring the error here, but it looks really out of place to ignore errors. I'm not sure we should have the extra indirection instead of verifying everything at the top.

cmd/conduit/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+// conduitCmd creates the main cobra command, initializes flags, and viper aliases
+func conduitCmd() *cobra.Command {
+	cfg := &conduit.Config{}
+	conduitCmd := &cobra.Command{
+		Use:   "conduit",
+		Short: "run the conduit framework",
+		Long:  "run the conduit framework",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
winder, line 52: Could you have a `Run` function which accepts the config object? We'll hopefully get to use it later in `daemon.go`.

conduit/pipeline.go
```diff
@@ -0,0 +1,286 @@
...
+		return fmt.Errorf("exporter configuration was empty")
+	}
+
+	return nil
+}
+
+// MakePipelineConfig creates a pipeline configuration
+func MakePipelineConfig(logger *log.Logger, cfg *Config) (*PipelineConfig, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("empty conduit config")
```
winder, line 58: Could you use the pattern we use elsewhere with the error string starting with the function name? That has been very helpful for debugging.

conduit/pipeline.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	logger.Infof("Auto-loading Conduit Configuration: %s", autoloadParamConfigPath)
+
+	configStr, err := ioutil.ReadFile(autoloadParamConfigPath)
+	if err != nil {
+		return nil, err
+	}
+
+	err = yaml.Unmarshal(configStr, &pCfg)
```
winder, line 83: If we are using viper, we should use it here also.

data/block_export_data.go
winder: I definitely don't think we should have a public method for this. If I was a plugin developer it might be tempting to call it in order to skip ahead.

conduit/pipeline.go
```diff
@@ -0,0 +1,286 @@
...
+
+	importer   *importers.Importer
+	processors []*processors.Processor
+	exporter   *exporters.Exporter
+}
+
+func (p *pipelineImpl) Start() error {
+	p.logger.Infof("Starting Pipeline Initialization")
+
+	// TODO need to change config to map[string]interface{}
```
winder, line 119: done?

exporters/exporter_factory.go
```diff
@@ -41,3 +41,13 @@ func ExporterByName(name string, dataDir string, logger *logrus.Logger) (Exporte
 	}
 	return val, nil
 }
+
+// ExporterBuilderByName returns a Processor constructor for the name provided
+func ExporterBuilderByName(name string) (ExporterConstructor, error) {
```
winder, line 46: May as well pass in the config and call the constructor as well. That way the caller doesn't need to care about how things are actually made.

conduit/pipeline.go
```diff
@@ -0,0 +1,286 @@
...
+type pipelineImpl struct {
+	ctx    context.Context
+	cfg    *PipelineConfig
+	logger *log.Logger
+
+	initProvider *data.InitProvider
+
+	importer   *importers.Importer
+	processors []*processors.Processor
+	exporter   *exporters.Exporter
```
winder, line 113: I would consider making all of these slices and generate errors when creating the things below if it isn't supported yet.

importers/algod/algod_importer.go
winder: +1

conduit/pipeline.go
```diff
@@ -0,0 +1,286 @@
...
+		ctx:          context.Background(),
+		cfg:          cfg,
+		logger:       logger,
+		initProvider: nil,
+		importer:     nil,
+		processors:   []*processors.Processor{},
+		exporter:     nil,
+	}
+
+	importerName, ok := cfg.Importer["Name"]
```
winder, line 237: Could you make this a property so that we don't need to check that it exists and cast it to a string?

conduit/logging.go
winder: Looks like you can provide a sorting function for this kind of thing without messing around with the names: https://pkg.go.dev/github.com/sirupsen/logrus#TextFormatter.SortingFunc

AlgoStephenAkiki: I think that is only for `TextFormatter` not `JSONFormatter` though.  Do we care if we lose the JSON data format?  I figured the structured logging would be useful for log aggregating tools.

conduit/pipeline.go
AlgoStephenAkiki: Should read: "Need to change interfaces to accept config of map[string]interface{}

conduit/pipeline.go
AlgoStephenAkiki: Could you explain more on that?  So would it become:

```
importer   []*importers.Importer
processors []*processors.Processor
exporter   []*exporters.Exporter
```
?

exporters/exporter_factory.go
AlgoStephenAkiki: There is a function above this that does that.  I wanted a function to pull out a constructor so that the pipeline can perform the same function all all objects in a defined sequence.  I figured it would be easier to reason about initialization that way.

conduit/pipeline.go
AlgoStephenAkiki: Would viper read the file in?

cmd/conduit/main.go
AlgoStephenAkiki: I think we need to modify it anyway since the function uses the `INDEXER` prefix which isn't what we want.

conduit/pipeline.go
winder: Yes, you give it a path and it loads in the config.

Looks like it has utilities for extracting subtrees too:
https://github.com/spf13/viper#extracting-a-sub-tree

conduit/pipeline.go
winder: Right. This way you can use the same pattern for initializing all of them, only the error handling would change. I don't think this was my most important bit of feedback, just something to consider.

Makefile
```diff
@@ -22,7 +22,7 @@ COVERPKG := $(shell go list ./...  | grep -v '/cmd/' | egrep -v '(testing|test|m
 # Used for e2e test
 export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 
-conduit: go-algorand
+conduit: cmd/algorand-indexer/algorand-indexer go-algorand
```
winder, line 25: This isn't a real dependency is it? If you want `make` to build `algorand-indexer` move this target down. The default target is the first one.

Alternatively, have the first target be "all".

cmd/conduit/main.go
AlgoStephenAkiki: For this level, we would only be setting the data directory via environmental variables.  All other (nested) configurations would be handled by the supplied config file `conduit.yml` and wouldn't be affected by environmental variables.

exporters/exporter_factory.go
winder: This suggestion applies to all of them, Use the `ByName` variant across each. There are probably a few good opportunities for generics in this code.

importers/algod/algod_importer.go
winder: Is this check still needed?

AlgoStephenAkiki: This isn't there on the latest commit.

Makefile
```diff
@@ -22,7 +22,12 @@ COVERPKG := $(shell go list ./...  | grep -v '/cmd/' | egrep -v '(testing|test|m
 # Used for e2e test
 export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 
-# This is the default target, build the indexer:
+# This is the default target, build everything:
+all: conduit cmd/algorand-indexer/algorand-indexer go-algorand idb/postgres/internal/schema/setup_postgres_sql.go idb/mocks/IndexerDb.go
+
+conduit:
```
winder, line 28: This is still a legit dependency that we should make sure is current.
```suggestion
conduit: go-algorand
```

cmd/conduit/main.go
```diff
@@ -0,0 +1,132 @@
...
+		Type: "Conduit",
+		Name: "main",
+	}
+
+	logger.SetFormatter(&formatter)
+	logger.SetOutput(os.Stdout)
+}
+
+// runConduitCmdWithConfig run the main logic with a supplied conduit config
+func runConduitCmdWithConfig(cfg *conduit.Config) error {
```
winder, line 42: having viper/pflag stuff in the run function might make it more difficult to use this in daemon later on. I think this can just accept a single `dataDir string` parameter.

winder: Should we manually check `CONDUIT_DATA_DIR` since it's the only special case?

conduit/logging.go
```diff
@@ -0,0 +1,20 @@
...
+
+// PluginLogFormatter formats the log message with special conduit tags
+type PluginLogFormatter struct {
+	Formatter *log.JSONFormatter
+	Type      string
+	Name      string
+}
+
+// Format allows this to be used as a logrus formatter
+func (f PluginLogFormatter) Format(entry *log.Entry) ([]byte, error) {
```
winder, line 15: I finally realized that you're implementing the `Format` interface. Neat trick :+1:

conduit/pipeline.go
```diff
@@ -0,0 +1,299 @@
...
+	pipelineLogLevel string `yaml:"LogLevel"`
+	PipelineLogLevel log.Level
+	// Store a local copy to access parent variables
+	Importer   nameConfigPair   `yaml:"Importer"`
+	Processors []nameConfigPair `yaml:"Processors"`
+	Exporter   nameConfigPair   `yaml:"Exporter"`
+}
+
+// Valid validates pipeline config
+func (cfg *PipelineConfig) Valid() error {
```
winder, line 45: This can be private

winder: It might also be possible to set defaults and have some of it be automatic:
https://github.com/spf13/viper#establishing-defaults

conduit/pipeline.go
```diff
@@ -0,0 +1,299 @@
...
+			PluginLogFormatter{
+				Formatter: &log.JSONFormatter{
+					DisableHTMLEscape: true,
+				},
+				Type: "Processor",
+				Name: (*processor).Metadata().Name(),
+			},
+		)
+
+		err := (*processor).Init(p.ctx, *p.initProvider, "")
```
winder, line 184: Is this missing the configuration and logger?

conduit/pipeline.go
```diff
@@ -0,0 +1,299 @@
...
+	genesis, err := (*p.importer).Init(p.ctx, plugins.PluginConfig(jsonEncode), importerLogger)
+
+	currentRound := basics.Round(0)
+
+	var initProvider data.InitProvider = &PipelineInitProvider{
+		currentRound: &currentRound,
+		genesis:      genesis,
+	}
+
+	p.initProvider = &initProvider
```
winder, line 163:
```suggestion
	p.initProvider = &PipelineInitProvider{
		currentRound: &currentRound,
		genesis:      genesis,
	}
```

Also, consider renaming to `pluginProvider` if we're using the same one for all plugin types.

conduit/pipeline.go
```diff
@@ -0,0 +1,299 @@
...
+			Formatter: &log.JSONFormatter{
+				DisableHTMLEscape: true,
+			},
+			Type: "Importer",
+			Name: (*p.importer).Metadata().Name(),
+		},
+	)
+
+	// TODO modify/fix ?
+	jsonEncode := string(json.Encode(p.cfg.Importer.Config))
```
winder, line 153: You could put a `String()` function on the `p.cfg.Importer.Config` object to streamline this a bit.

conduit/pipeline.go
```diff
@@ -0,0 +1,299 @@
...
+		PluginLogFormatter{
+			Formatter: &log.JSONFormatter{
+				DisableHTMLEscape: true,
+			},
+			Type: "Exporter",
+			Name: (*p.exporter).Metadata().Name(),
+		},
+	)
+
+	err = (*p.exporter).Init("", p.logger)
```
winder, line 204: This one is also missing the config.

conduit/pipeline.go
```diff
@@ -0,0 +1,299 @@
...
+	logger *log.Logger
+
+	initProvider *data.InitProvider
+
+	importer   *importers.Importer
+	processors []*processors.Processor
+	exporter   *exporters.Exporter
+}
+
+func (p *pipelineImpl) Start() error {
```
winder, line 136: consider renaming to `initialize`, since I'm sure there is going to be a lot more to do for `Start` in the next iteration.

winder (approved):
I think this is in a good spot, we can followup on any remaining tasks in later PRs.

PR #1169, opened 2022-08-10
cicd: add darwin arm64 support to release script

algolucky:
## Summary

this adds support for darwin machines running on arm64 architecture (m1) to the `release.py` script. this makes it so that darwin-arm64 builds can be enabled at a later time.

### DIFF
--- a/misc/release.py
+++ b/misc/release.py
@@ -38,6 +38,7 @@ osArchArch = [
     ("linux", "arm", "armhf"),
     ("linux", "arm64", "arm64"),
     ("darwin", "amd64", None),
+    ("darwin", "arm64", None)
 ]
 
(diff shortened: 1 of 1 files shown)


winder:
This looks like cross compiling, I'm not sure that works anymore ever since we started including go-algorand.

algolucky:
> This looks like cross compiling

I'm not sure I'm following. from my testing, if I execute the `package` make target on darwin-arm64, `release.py` skips compilation and doesn't create a release artifact.

algobarb (approved):
This looks to work as expected on the test pipeline run for darwin arm64 by adding this.

algojack (approved):

PR #1178, opened 2022-08-14
Add waits/retries to algod importer

Eric-Warehime:
## Summary

Indexer's fetcher uses the GetBlock API while catching up since it knows the blocks should exist, and adds a WaitForBlock call during follow. 

I've added a wait call for each block fetch so that fetching blocks via the Algod Importer should function properly for both older blocks and new ones.

I've also refactored the mock algod server used in these tests and the fetcher tests to use a common package.

Also added files to the codecov ignore list since they are mocks/tests.

## Test Plan

Added/updated unit tests.

### DIFF
--- a/fetcher/fetcher_test.go
+++ b/fetcher/fetcher_test.go
@@ -69,7 +69,7 @@ func TestFetcherImplCatchupLoopBlockError(t *testing.T) {
 		// returning an empty block `passingCalls` times before throwing 500s
 		func(path string, w http.ResponseWriter) bool {
 			if strings.Contains(path, "v2/blocks/") {
-				if passingCalls <= 0 {
+				if passingCalls == 0 {
 					w.WriteHeader(http.StatusInternalServerError)
 				} else {
 					var block bookkeeping.Block
--- a/fetcher/fetcher_test.go
+++ b/fetcher/fetcher_test.go
@@ -3,8 +3,9 @@ package fetcher
 import (
 	"context"
 	"fmt"
+	"github.com/algorand/indexer/util/test"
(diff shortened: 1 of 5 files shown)


fetcher/fetcher_test.go
```diff
@@ -36,86 +27,69 @@ func (handler *BlockHandler) handlerFunc(ctx context.Context, cert *rpcs.Encoded
...
 func TestFetcherImplCatchupLoopBlockError(t *testing.T) {
-	mockAlgodHandler := &AlgodHandler{}
-	aclient := mockAClient(t, mockAlgodHandler)
 	passingCalls := 5
+	aclient, err := test.MockAClient(test.NewAlgodHandler(
+		// Our mock algod client will process /v2/blocks/{round} calls
+		// returning an empty block `passingCalls` times before throwing 500s
+		func(path string, w http.ResponseWriter) bool {
+			if strings.Contains(path, "v2/blocks/") {
+				if passingCalls <= 0 {
```
shiqizng, line 72: nit: passingCalls==0?

importers/algod/algod_importer_test.go
```diff
@@ -2,22 +2,13 @@ package algodimporter
 
 import (
 	"context"
-	"net/http"
-	"net/http/httptest"
-	"path"
-	"strconv"
-	"strings"
+	"github.com/algorand/indexer/util/test"
```
AlgoStephenAkiki, line 5: nit: ordering

AlgoStephenAkiki (approved):

shiqizng (approved):

PR #1179, opened 2022-08-15
Bug Fix: Fixes api-config bug

### LINKED ISSUE #1094: Running `algorand-indexer api-config` command emits unwanted messages to the log
This line: https://github.com/algorand/indexer/blob/79e84763bc9519b8a54c1a24d71c00f9344af5ca/cmd/algorand-indexer/api_config.go#L38 emits an info messages to the log (which is by default `os.Stdout`).

According to https://github.com/algorand/indexer/blob/develop/docs/DisablingParametersGuide.md#how-do-i-supply-my-own-configuration, it is suggested to use the `algorand-indexer api-config` command to generate the `api_config.yml` file by redirecting the command's output:
```sh
~$ mkdir ~/indexer-data
~$ algorand-indexer api-config > ~/indexer-data/api_config.yml
~$ algorand-indexer daemon --data-dir ~/indexer-data
```
The result is an `api_config.yml` file that includes info log messages:
```yaml
{"level":"info","msg":"Enable all parameters flag is set to: false","time":"2022-07-05T21:50:14Z"}
/v2/accounts:
    optional:
        - currency-greater-than: disabled
        - currency-less-than: disabled
/v2/accounts/{account-id}/transactions:
    optional:
        - note-prefix: disabled
        - tx-type: disabled
        - sig-type: disabled
        - asset-id: disabled
        - before-time: disabled
        - after-time: disabled
        - rekey-to: disabled
/v2/assets:
    optional:
        - name: disabled
        - unit: disabled
/v2/assets/{asset-id}/balances:
    optional:
        - currency-greater-than: disabled
        - currency-less-than: disabled
/v2/assets/{asset-id}/transactions:
    optional:
        - note-prefix: disabled
        - tx-type: disabled
        - sig-type: disabled
        - before-time: disabled
        - after-time: disabled
        - currency-greater-than: disabled
        - currency-less-than: disabled
        - address-role: disabled
        - exclude-close-to: disabled
        - rekey-to: disabled
    required:
        - asset-id: disabled
/v2/transactions:
    optional:
        - note-prefix: disabled
        - tx-type: disabled
        - sig-type: disabled
        - asset-id: disabled
        - before-time: disabled
        - after-time: disabled
        - currency-greater-than: disabled
        - currency-less-than: disabled
        - address-role: disabled
        - exclude-close-to: disabled
        - rekey-to: disabled
        - application-id: disabled
```

AlgoStephenAkiki:
Resolves #1094 

Current branch functionality removes source of inadvertent log message but
causes a panic when a configuration file is not supplied.  Supply
the default Postgres configuration to give appropriate behavior.

## Summary

## Test Plan

### DIFF
--- a/cmd/algorand-indexer/api_config.go
+++ b/cmd/algorand-indexer/api_config.go
@@ -42,6 +42,8 @@ var apiConfigCmd = &cobra.Command{
 				fmt.Fprintf(os.Stderr, "failed to created disabled map config from file: %v", err)
 				panic(exit{1})
 			}
+		} else {
+			potentialDisabledMapConfig = api.GetDefaultDisabledMapConfigForPostgres()
 		}
 
 		var displayDisabledMapConfig *api.DisplayDisabledMap


winder:
Fixed by #1180

PR #1183, opened 2022-08-15
conduit: pipeline run loop implementation

Eric-Warehime:
## Summary

I've gone through the initial implementation of the conduit pipeline run loop, and found/fixed a few bugs along the way.

Bug fixes/changes:
* Configs were being loaded, but empty string were being passed into plugins during init--now we parse and load plugins with proper configs
* The Processor interface was the only plugin interface not taking a logger, so I've added one
* Additionally, I've made sure that all conduit loggers are properly being passed to Inits
* Fixed import order in as many places as I saw
* Rearranged ledger variable assignment in block_processor to avoid a null pointer exception
* algod importer was not properly decoding config data--switched to using the protocol package to fix this issue
* Fixed an issue where importing the genesis file in the ledger initializes the 0 block, but the same is not true for pgsql
  * We now no-op the 0 round in the ledger instead of failing on it.
* Added a run loop for the pipeline--infinitely runs, exiting on any error

I know of a few more bugs that I wasn't able to address here that I will create tickets for.

## Test Plan

This PR runs using local algod importer, (noop/block_processor) processors, and (noop/postgresql) exporters.

### DIFF
--- a/conduit/pipeline.go
+++ b/conduit/pipeline.go
@@ -131,12 +131,32 @@ type pipelineImpl struct {
 	importer   *importers.Importer
 	processors []*processors.Processor
 	exporter   *exporters.Exporter
+	round      basics.Round
 }
 
 func (p *pipelineImpl) Start() error {
 	p.logger.Infof("Starting Pipeline Initialization")
 
 	// TODO Need to change interfaces to accept config of map[string]interface{}
+
+	exporterLogger := log.New()
+	exporterLogger.SetFormatter(
+		PluginLogFormatter{
+			Formatter: &log.JSONFormatter{
+				DisableHTMLEscape: true,
+			},
+			Type: "Exporter",
+			Name: (*p.exporter).Metadata().Name(),
+		},
+	)
+
+	jsonEncode := string(json.Encode(p.cfg.Exporter.Config))
+	err := (*p.exporter).Init(plugins.PluginConfig(jsonEncode), exporterLogger)
+	exporterName := (*p.exporter).Metadata().Name()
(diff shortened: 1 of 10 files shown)


conduit/pipeline.go
```diff
@@ -131,13 +131,33 @@ type pipelineImpl struct {
...
+	exporterLogger := log.New()
+	exporterLogger.SetFormatter(
+		PluginLogFormatter{
+			Formatter: &log.JSONFormatter{
+				DisableHTMLEscape: true,
+			},
+			Type: "Exporter",
+			Name: (*p.exporter).Metadata().Name(),
+		},
+	)
```
winder, line 151: We should probably have a helper to construct these.

conduit/pipeline.go
```diff
@@ -131,13 +131,33 @@ type pipelineImpl struct {
...
+
+	jsonEncode := string(json.Encode(p.cfg.Exporter.Config))
+	err := (*p.exporter).Init(plugins.PluginConfig(jsonEncode), exporterLogger)
```
winder, line 154: Could you put a comment somewhere that this is imported first so that we can fetch the round?
Would we ever want multiple exporters, which round do we choose in that scenario?

Eric-Warehime: I'll update w/ comment.

I think it's hard to enumerate all of possible combinations of multiple exporters. I'd prefer to make it really easy to run a pipeline so that people can just spin up multiple pipelines if they need to run multiple exporters.

winder (approved):
LGTM

PR #1189, opened 2022-08-19
Conduit: Network importer implementation

algoganesh:
## Summary
This PR introduces a new importer plugin, `network-importer`, that allows users to fetch block directly from the network.
Config file provided to the plugin must contain paths to `genesis` and `config` files.

### DIFF
--- a/importers/network/network_importer.go
+++ b/importers/network/network_importer.go
@@ -1,0 +1,124 @@
+package networkimporter
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/algorand/go-algorand/catchup"
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/network"
+
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/indexer/data"
+	"github.com/algorand/indexer/importers"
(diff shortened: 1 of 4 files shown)


importers/network/network_importer.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return blk, err
+	}
+	blk.BlockHeader = block.BlockHeader
+	blk.Payset = block.Payset
+	blk.Certificate = cert
+	return blk, nil
+}
+
+func (networkImp *networkImporter) unmarhshalConfig(cfg string) error {
```
winder, line 122:
```suggestion
func (networkImp *networkImporter) unmarshalConfig(cfg string) error {
```

importers/network/network_importer_test.go
```diff
@@ -0,0 +1,118 @@
...
+	assert.Equal(t, metadata.Name(), networkImporterMetadata.Name())
+	assert.Equal(t, metadata.Description(), networkImporterMetadata.Description())
+	assert.Equal(t, metadata.Deprecated(), networkImporterMetadata.Deprecated())
+}
+
+func TestCloseSuccess(t *testing.T) {
+	testImporter = New()
+	_, err := testImporter.Init(ctx, plugins.PluginConfig(
+		"genesis-path: /Users/ganesh/go-algorand/installer/genesis/mainnet/genesis.json"+
+			"\nconfig-path: /Users/ganesh/go-algorand/installer/"), logger)
```
winder, line 44: This might be a little more readable if it created the ImporterConfig object and serialized it to a string.

Also, I wonder how we could do this without depending on a genesis.json file.

PR #1197, opened 2022-08-19
Enhancement: update e2e test policy

shiqizng:
## Summary

This PR pins submodule to rel/nightly builds.  

## Test Plan

Make sure ci runs as expected.


.circleci/config.yml
```diff
@@ -45,6 +45,8 @@ jobs:
     parameters:
       go_version:
         type: string
+    environment:
+      CI_E2E_FILENAME: "net_done"
```
Eric-Warehime, line 49: I don't know if this is what we want.

> PRs should use an artifact that is pinned according to the current submodule.
The pinned submodule can be overridden to allow pinning to a specific feature branch.

Correct me if I'm wrong...

The most recent changes to the go-algorand CI mean that we are only currently uploading e2e test files for builds of the `rel-nightly` branch unless someone manually runs the CI w/ the parameter supplied.
See https://github.com/algorand/go-algorand/blob/master/.circleci/config.yml#L635 and https://github.com/algorand/go-algorand/blob/master/test/scripts/e2e.sh#L166

This means that `net_done.tar.bz2`, which is now the default name when `CI_E2E_FILENAME` is not found https://github.com/algorand/go-algorand/blob/master/test/scripts/e2e.sh#L180 should not get updated almost ever.

We need to modify the go-algorand CI to upload files with their commit hash associated w/ them so that we can look at our submodule and pull the e2e test files that were generated from the same commit that our submodule is running.

Eric-Warehime (changes requested):

.circleci/config.yml
shiqizng: this makes sense. I'll make the changes.

shiqizng: in go-algorand, do we want to create an artifact for every merge to master or create another artifact with commit hash as part of rel/nightly build?

shiqizng: or can we tag the rel/nightly artifact with commit hash?

shiqizng: or can we tag the rel/night artifact by adding the commit hash in the metadata?

winder: I think you can select a specific artifact by providing a longer "prefix" in order to make it unique / searchable. For example: `faffce2f/rel-nightly`

misc/e2elive.py
```diff
@@ -84,7 +84,15 @@ def main():
         from botocore import UNSIGNED
         s3 = boto3.client("s3", config=Config(signature_version=UNSIGNED))
         tarpath = os.path.join(tempdir, tarname)
-        prefix = "indexer/e2e4"
+        if os.getenv("E2E_FILE_VERSION")=="submodule":
```
winder, line 87: There are some things to consider here:
1) I believe this approach, where we look up the commit hash, fails if we set the submodule to something that wasn't the latest commit. i.e. if there were 10 commits today, only one of them would be used for the nightly test. Also, would it ever work if we pin the submoule to rel/beta or master?
2) We could use the prefix approach, but then we would need to remember to update the E2E_FILE_VERSION each time the submodule is updated.

shiqizng: right. this approach would only works with rel/nightly artifacts. I don't think we need to update E2E_FILE_VERSION. it's a flag for determining whether to use rel-nightly artifiact or use the one with the same commit hash as the submodule.

Makefile
```diff
@@ -68,7 +68,7 @@ integration: cmd/algorand-indexer/algorand-indexer
 	test/postgres_integration_test.sh
 
 e2e: cmd/algorand-indexer/algorand-indexer
-	cd misc && docker-compose build --build-arg GO_IMAGE=${GO_IMAGE} && docker-compose up --exit-code-from e2e
+	cd misc && docker-compose build --build-arg GO_IMAGE=${GO_IMAGE} && VERSION=${CI_E2E_FILE_VERSION} docker-compose up --exit-code-from e2e
```
tzaffi, line 71: I wonder if `VERSION` or `CI_E2E_FILE_VERSION` can just be added to the `e2e` service in `misc/docker-compose.yml`. The more we can unify the locations of similar env vars, the easier it will be to comprehend the meaning. But I haven't tried to make sure it works.

misc/buildtestdata.sh
```diff
@@ -49,4 +49,4 @@ mkdir -p "${E2EDATA}"
 RSTAMP=$(TZ=UTC python -c 'import time; print("{:08x}".format(0xffffffff - int(time.time() - time.mktime((2020,1,1,0,0,0,-1,-1,-1)))))')
 
 echo "COPY AND PASTE THIS TO UPLOAD:"
```
tzaffi, line 51:
```suggestion
```
Looks like the "COPY AND PASTE" suggestion is no longer applicable since the command is running directly in the script.

misc/e2elive.py
tzaffi: Since this is running in a python script, you could use python-native git libraries. I already have an example of this in my parity job [here](https://github.com/algorand/indexer/blob/b197dc35b836d8cf443dab7c6510bd1be6f089d0/misc/parity/test_indexer_v_algod.py#L36-L40)

Makefile
```diff
@@ -93,4 +93,7 @@ indexer-v-algod-swagger:
 
 indexer-v-algod: nightly-setup indexer-v-algod-swagger nightly-teardown
 
+update-submodule:
```
tzaffi, line 96: Is a reversal of this command needed as well to bring back the git-submodule to its previous state so that other follow-up tests on the same box work as expected? If this is for the very last job, it's not as much of an issue. But to be extra careful you could add a companion that reverts the update. E.g. `make nightly-setup` with its companion `make nightly-teardown`.

tzaffi: also consider a slightly more informative name such as `rel-nightly-update-submodule` or similar

shiqizng: this target is not for nightly run. it's meant to be ran manually when the submodule needs to be updated. I think naming it `rel-nightly-update-submodule` would be confusing.

Eric-Warehime (approved):

winder (approved):

.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ workflows:
 jobs:
   test:
```
winder, line 42: Don't we still need the environment variable for this one?

```
    environment:
      CI_E2E_FILENAME: "494948f4/rel-nightly"
```

shiqizng: actually, we still need to upload an artifact to the path  "<commit-hash>/rel-nightly" and in e2elive it will determine the correct commit hash to use based on the submodule.

winder (approved):
LGTM

winder (approved):

.circleci/config.yml
```diff
@@ -41,10 +41,12 @@ workflows:
...
     machine:
-      image: << pipeline.parameters.ubuntu_image >> 
+      image: << pipeline.parameters.ubuntu_image >>
     parameters:
       go_version:
         type: string
+    environment:
+      CI_E2E_FILENAME: "fafa8862/rel-nightly"
```
tzaffi, line 49: It might be helpful to add a note to the [misc README](https://github.com/algorand/indexer/blob/develop/misc/README.md) explaining when this needs to be updated, especially if this env var is expected to be updated on a regular basis.

shiqizng: added a note in the readme.

PR #1199, opened 2022-08-21
Testing: Add a sample state proof with high index for testing

algoidan:
## Summary
Use this sample of a state proof taken from beta-net. In this example, one of the reveals given by the proof has a high index (== this participant has a relatively low stake - but it might get chosen to reveal the signature ). This sample might cause a problem if state-proof decoding has an error.


Eric-Warehime (changes requested):
Thanks for adding these! I think the existing test cases are still valid. Instead of replacing them with the new ones can we add these as additional test cases?

Eric-Warehime (approved):

PR #1208, opened 2022-08-29
conduit: Convert daemon to conduit pipeline

Eric-Warehime:
## Summary

Converts `algorand-indexer`'s data pipeline to a conduit pipeline.

## Test Plan

Indexer e2e tests.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -36,7 +36,7 @@ type ServerImplementation struct {
 
 	db idb.IndexerDb
 
-	fetcher error
+	dataError func() error
 
 	timeout time.Duration
 
@@ -139,8 +139,10 @@ func (si *ServerImplementation) MakeHealthCheck(ctx echo.Context) error {
 		errors = append(errors, fmt.Sprintf("database error: %s", health.Error))
 	}
 
-	if si.fetcher != nil && si.fetcher.Error() != "" {
-		errors = append(errors, fmt.Sprintf("fetcher error: %s", si.fetcher.Error()))
+	if si.dataError != nil {
+		if err := si.dataError(); err != nil {
+			errors = append(errors, fmt.Sprintf("data error: %s", err))
+		}
 	}
 
 	return ctx.JSON(http.StatusOK, common.HealthCheckResponse{
--- a/conduit/pipeline.go
+++ b/conduit/pipeline.go
@@ -3,13 +3,12 @@ package conduit
 import (
 	"context"
 	"fmt"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
 	"os"
 	"path/filepath"
 	"runtime/pprof"
 	"sync"
-
-	log "github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
 
 	"github.com/algorand/go-algorand-sdk/encoding/json"
 	"github.com/algorand/go-algorand/data/basics"
@@ -122,19 +121,22 @@ func MakePipelineConfig(logger *log.Logger, cfg *Config) (*PipelineConfig, error
 // sequence of events, taking in importers, processors and
 // exporters and generating the result
 type Pipeline interface {
-	Start() error
+	Init() error
+	Start()
 	Stop()
 	Error() error
+	Wait()
 }
 
 type pipelineImpl struct {
 	ctx      context.Context
 	cf       context.CancelFunc
-	running  sync.WaitGroup
+	wg       sync.WaitGroup
 	cfg      *PipelineConfig
 	logger   *log.Logger
-	err      error
 	profFile *os.File
+	err      error
+	mu       sync.RWMutex
 
 	initProvider *data.InitProvider
 
@@ -145,10 +147,19 @@ type pipelineImpl struct {
 }
 
 func (p *pipelineImpl) Error() error {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
 	return p.err
 }
 
-func (p *pipelineImpl) Start() error {
+func (p *pipelineImpl) setError(err error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.err = err
+}
+
+// Init prepares the pipeline for processing block data
+func (p *pipelineImpl) Init() error {
 	p.logger.Infof("Starting Pipeline Initialization")
 
 	if p.cfg.CPUProfile != "" {
@@ -226,13 +237,12 @@ func (p *pipelineImpl) Start() error {
 		p.logger.Infof("Initialized Processor: %s", processorName)
 	}
 
-	go p.runPipeline()
 	return err
 }
 
 func (p *pipelineImpl) Stop() {
 	p.cf()
-	p.running.Wait()
+	p.wg.Wait()
 
 	if p.profFile != nil {
 		if err := p.profFile.Close(); err != nil {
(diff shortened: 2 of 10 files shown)


api/handlers.go
```diff
@@ -139,8 +139,10 @@ func (si *ServerImplementation) MakeHealthCheck(ctx echo.Context) error {
 		errors = append(errors, fmt.Sprintf("database error: %s", health.Error))
 	}
 
-	if si.fetcher != nil && si.fetcher.Error() != "" {
-		errors = append(errors, fmt.Sprintf("fetcher error: %s", si.fetcher.Error()))
+	if si.dataError != nil {
+		if err := si.dataError(); err != nil {
```
shiqizng, line 143:
```suggestion
		if si.dataError != nil && si.dataError() != nil {
		}
```

conduit/pipeline.go
```diff
@@ -118,13 +112,17 @@ func MakePipelineConfig(logger *log.Logger, cfg *Config) (*PipelineConfig, error
...
+	Error() error
 }
 
 type pipelineImpl struct {
-	ctx    context.Context
-	cfg    *PipelineConfig
-	logger *log.Logger
+	ctx     context.Context
+	cf      context.CancelFunc
+	running sync.WaitGroup
```
shiqizng, line 122: nit: consider naming this wg so it's clear that it's a go waitgroup.

conduit/pipeline.go
```diff
@@ -208,49 +214,55 @@ func (p *pipelineImpl) Stop() error {
...
-				// TODO Retries?
 				p.logger.Infof("Pipeline round: %v", p.round)
 				// fetch block
 				blkData, err := (*p.importer).GetBlock(uint64(p.round))
 				if err != nil {
-					p.logger.Errorf("%v\n", err)
-					return err
+					p.logger.Errorf("%v", err)
+					p.err = err
+					goto pipelineRun
```
shiqizng, line 237: this will retry forever?

conduit/pipeline.go
Eric-Warehime: I've copied the name from https://github.com/algorand/go-algorand/blob/master/catchup/catchpointService.go#L75 since I liked it.

To me, `running is a waitgroup counting the running goroutine(1)` makes sense. Open to changing it if we want to stick with `wg` for waitgroup vars though.

conduit/pipeline.go
Eric-Warehime: I've tried to mimic the behaviour that we have in fetcher. https://github.com/algorand/indexer/blob/develop/fetcher/fetcher.go#L219

This will run until the context is expires, which is probably whenever the cancel func is called. Errors can be raised by individual plugins, at which point we will set `p.err` so that it can be referenced by the `p.Error()` function, and the pipeline will start running again from the Importer on the same round.

conduit/pipeline.go
```diff
@@ -208,49 +214,55 @@ func (p *pipelineImpl) Stop() error {
...
 					if err != nil {
-						p.logger.Errorf("%v\n", err)
-						return err
+						p.logger.Errorf("%v", err)
+						p.err = err
+						goto pipelineRun
 					}
 				}
 				// Increment Round
+				p.err = nil
```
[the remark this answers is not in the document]
Eric-Warehime: If an error has occurred in one of the plugins, `p.err` will be set to that. The pipeline will continue trying to process this round, and if it is successful we set the error to nil and move onto processing the next round. It's similar to https://github.com/algorand/indexer/blob/develop/fetcher/fetcher.go#L182 for reference.

api/handlers.go
Eric-Warehime: The reason I've separated it out is that we need to add `err` to the set of known errors. We could just call `si.dataError()` again inside the block to reference the error, but because the pipeline will keep retrying it's possible that the first call would show an error but the second call might return `nil`.

This way we have a single call which either returns an error that is added to the health check or it's `nil`.

conduit/pipeline_test.go
```diff
@@ -327,47 +339,29 @@ func TestPipelineCpuPidFiles(t *testing.T) {
...
-	wg.Add(1)
-
-	go func() {
-		err := pImpl.Start()
-		assert.Nil(t, err)
-		wg.Done()
-	}()
-
-	// Give some time for the goroutine to start
-	time.Sleep(1 * time.Second)
```
winder, line 341: Why isn't the delay needed anymore?

winder (changes requested):
Looks really good. The only thing that really should change is how the blocking is done in `runConduitCmdWithConfig`.

conduit/pipeline.go
winder: I guess this was done in order to match the current Indexer behavior?

Does Indexer also retry on processor and writer errors?

Thinking out loud here, do you think that the error handling policy should be configurable per plugin?

conduit/pipeline.go
```diff
@@ -228,10 +226,27 @@ func (p *pipelineImpl) Start() error {
...
 	}
 
-	return p.RunPipeline()
+	go p.runPipeline()
+	return err
 }
 
-func (p *pipelineImpl) Stop() error {
+func (p *pipelineImpl) Stop() {
+	p.cf()
```
AlgoStephenAkiki, line 234: How is this cancel function set?

conduit/pipeline_test.go
AlgoStephenAkiki: +1...just curious

conduit/pipeline.go
Eric-Warehime: In `MakePipeline` which takes a context, we create a cancel context and assign it to the pipelineImpl. 

The one thing which I haven't included is an easy way to reset that context. So if you call `Start` and then `Stop` things wouldn't really work if you called `Start` again.

conduit/pipeline_test.go
Eric-Warehime: `Pipeline.Start` synchronously creates the PID file and CPU Profile if necessary, launches the goroutine which runs block imports, and then returns. So since we're not creating these files in a goroutine anymore they will always be created by the time we've returned.

conduit/pipeline.go
Eric-Warehime: > I guess this was done in order to match the current Indexer behavior?

Yes, I've tried to match the behaviour of the current Indexer here.

> Does Indexer also retry on processor and writer errors?

The Indexer's data pipeline is controlled by fetcher which runs two things concurrently:
* Fetching and enqueuing blocks
* Running the block handler on the queued blocks

The fetcher will stop running when:
* The cancel function is called or the context is `Done`
* It receives block bytes from algod which it cannot decode into a Block object
* The handler returns an error

However, the [handler we're using](https://github.com/algorand/indexer/blob/develop/cmd/algorand-indexer/daemon.go#L492) will just continue retrying when it errors. The effect is that fetcher only exits on malformed block data or context canceling. 


> Thinking out loud here, do you think that the error handling policy should be configurable per plugin?

I think it's a good idea to allow plugins to either return an error that can be retried or one that causes the pipeline to terminate. At the moment though, if we want to be backwards compatible w/ Indexer, all errors in the existing plugins would be non-fatal (except for block decoding).

winder: Great, thanks for the clarifications. This approach sounds good to me.

conduit/pipeline.go
winder: I agree with Shiqi here. Though I've never seen it written down, calling this `wg` seems to be a go idiom similar to `ctx` and `cf` above.

conduit/pipeline.go
```diff
@@ -228,10 +226,27 @@ func (p *pipelineImpl) Start() error {
 		p.logger.Infof("Initialized Processor: %s", processorName)
 	}
 
-	return p.RunPipeline()
+	go p.runPipeline()
+	return err
```
winder, line 230: I missed this in my first pass. I think it's better to let the caller decide when to start a go routine, that way it's more natural to control a graceful shutdown with sync.WaitGroup.

It is nice that the pidFile/profFile tests are deterministic this way.

Maybe it would be the best of both solutions to split Init/Start by renaming:
* `Start` -> `Initialize`
* `runPipeline` -> `Start`

Eric-Warehime: Moved all of the goroutine specific code into `Start` and added a synchronous `Init` function. Also added a blocking `Wait` function to the pipeline so that we don't have an empty loop places where we have no other blocking operations to wait on.

Slightly modified the error variable used in the handlers, but like you've said I don't think there's an idiomatic way to achieve that other than just protecting reads/writes with a mutex. 

Let me know what you think...I tried your suggestion above of moving wait group management into the conduit daemon, but it ended up splitting up the wait group and context handling/blocking in ways that I didn't like.

conduit/pipeline.go
AlgoStephenAkiki: I think we just say that once you call stop you are closing conduit.

AlgoStephenAkiki (approved):

winder (approved):

shiqizng (approved):

PR #1213, opened 2022-09-06
file exporter

shiqizng:
## Summary

write block data to a file. 

## Test Plan
added unit tests for the exporter.

### DIFF
--- a/exporters/filewriter/file_exporter.go
+++ b/exporters/filewriter/file_exporter.go
@@ -1,0 +1,122 @@
+package filewriter
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/indexer/data"
+	"github.com/algorand/indexer/exporters"
+	"github.com/algorand/indexer/plugins"
+	"github.com/sirupsen/logrus"
+)
+
+const exporterName = "filewriter"
+
+type fileExporter struct {
+	round  uint64
+	fh     *os.File
+	cfg    ExporterConfig
+	logger *logrus.Logger
+}
+
+var fileExporterMetadata = exporters.ExporterMetadata{
+	ExpName:        exporterName,
+	ExpDescription: "Exporter for writing data to a file.",
+	ExpDeprecated:  false,
+}
+
+// Constructor is the ExporterConstructor implementation for the filewriter exporter
+type Constructor struct{}
+
+// New initializes a fileExporter
+func (c *Constructor) New() exporters.Exporter {
+	return &fileExporter{
+		round: 0,
+	}
+}
+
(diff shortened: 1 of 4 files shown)


exporters/filewriter/file_exporter.go
```diff
@@ -0,0 +1,135 @@
...
+	return &fileExporter{
+		round: 0,
+	}
+}
+
+func (exp *fileExporter) Metadata() exporters.ExporterMetadata {
+	return fileExporterMetadata
+}
+
+func (exp *fileExporter) Init(cfg plugins.PluginConfig, logger *logrus.Logger) error {
```
shiqizng, line 49: can PluginConfig be a file path to a configs file? For the file exporter, Close() should update the round in the configs file so the value doesn't need to be manually updated.

Eric-Warehime: > can PluginConfig be a file path to a configs file?

I'm not sure what you mean. The `PluginConfig` is just a wrapper around a string. Each plugin then deserializes that into expected values via defined configs such as the one in `file_exporter_config.go`.

> For the file exporter, Close() should update the round in the configs file so this value doesn't need to be manually updated.

Yes, this sounds correct--that sort of thing is exactly why we have the `Close` function.

exporters/filewriter/file_exporter.go
```diff
@@ -0,0 +1,160 @@
...
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/indexer/data"
+	"github.com/algorand/indexer/exporters"
+	"github.com/algorand/indexer/plugins"
+	"github.com/sirupsen/logrus"
```
Eric-Warehime, line 17:
```suggestion
	"github.com/sirupsen/logrus"
	"gopkg.in/yaml.v3"

	"github.com/algorand/go-algorand/crypto"
	"github.com/algorand/go-algorand/data/bookkeeping"

	"github.com/algorand/indexer/data"
	"github.com/algorand/indexer/exporters"
	"github.com/algorand/indexer/plugins"
```

Eric-Warehime (approved):

exporters/filewriter/file_exporter_config.go
```diff
@@ -0,0 +1,12 @@
...
+// ExporterConfig specific to the file exporter
+type ExporterConfig struct {
+	// round to start at
+	Round uint64 `yaml:"round"`
+	// full file path to existing block file.
+	// Create if file doesn't exist
+	BlockFilepath string `yaml:"path"`
+	// full file path to a configuration file.
+	ConfigFilePath string `yaml:"configs"`
+}
```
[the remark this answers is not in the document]
shiqizng: I will make the change to write blocks to separate files. ConfigFilePath is needed to update round value in the configs when Close() is called. This way after restart, the exporter should pick up from the correct next round.

winder (approved):
Thanks for the changes, this looks good.

Eric-Warehime (approved):

PR #1220, opened 2022-09-13
Initial Filter Processor

### LINKED ISSUE #1100: Implement Filter Processor
Implement a new Processor for filtering data.

At a minimum it should allow the following options:
* Filter by address (sender/receiver/etc)
* Filter by app-id
* Filter by asset-id
* Filter ApplyData (remove it, at least app state)

AlgoStephenAkiki:
Resolves #1100

Adds dynamic filter processor and initial tests

## Summary

## Test Plan


processors/filterprocessor/filter_processor.go
```diff
@@ -0,0 +1,305 @@
...
+type fieldOperation string
+
+const someFieldOperation fieldOperation = "some"
+const allFieldOperation fieldOperation = "all"
+
+type fieldSearcher struct {
+	Exp          *expression
+	Tag          string
+	MethodToCall string
+}
```
Eric-Warehime, line 85: I feel like moving `fieldSearcher` and `Expression` and all of their related methods to a separate module or package would make it easier to see what the expected usage of these is. As is, some of the methods are exported and some or not so it's a bit tough to figure out how this library of functions might be used by another package effectively.

It will also probably make testing them easier.

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestCheckTagExistsAndHasCorrectFunction(t *testing.T) {
+	// check that something that doesn't exist throws an error
+	err := checkTagExistsAndHasCorrectFunction("const", "SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.LoreumIpsum.SDF")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "does not exist in transactions")
+
+	err = checkTagExistsAndHasCorrectFunction("const", "LoreumIpsum")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "does not exist in transactions")
```
Eric-Warehime, line 26:
```suggestion
	assert.ErrorContains(t, err, "does not exist in transactions")
```

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Fee does not have a "String" Function so we cant use const with it.
+	err = checkTagExistsAndHasCorrectFunction("const", "SignedTxnWithAD.SignedTxn.Txn.Header.Fee")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "does not contain the needed method")
```
Eric-Warehime, line 31:
```suggestion
	assert.ErrorContains(t, err, "does not contain the needed method")
```

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/algorand/indexer/data"
+	"github.com/algorand/indexer/plugins"
+	"github.com/algorand/indexer/processors"
+)
+
+func TestCheckTagExistsAndHasCorrectFunction(t *testing.T) {
+	// check that something that doesn't exist throws an error
+	err := checkTagExistsAndHasCorrectFunction("const", "SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.LoreumIpsum.SDF")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "does not exist in transactions")
```
Eric-Warehime, line 22:
```suggestion
	assert.ErrorContains(t, err, "does not exist in transactions")
```

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+							Receiver: sampleAddr3,
+						},
+					},
+				},
+			},
+		},
+	)
+
+	output, err := fp.Process(bd)
+	assert.Nil(t, err)
```
Eric-Warehime, line 98:
```suggestion
	assert.NoError(t, err)
```

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+    - tag: SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.Receiver
+      expression-type: regex 
+      expression: "` + sampleAddr1.String() + `"
+    - tag: SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.Receiver
+      expression-type: const
+      expression: "` + sampleAddr2.String() + `"
+`
+
+	fpBuilder, err := processors.ProcessorBuilderByName(implementationName)
+	assert.Nil(t, err)
```
Eric-Warehime, line 53:
```suggestion
	assert.NoError(t, err)
```

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	fp := fpBuilder.New()
+	err = fp.Init(context.Background(), &conduit.PipelineInitProvider{}, plugins.PluginConfig(sampleCfgStr), logrus.New())
+	assert.Nil(t, err)
```
Eric-Warehime, line 57:
```suggestion
	assert.NoError(t, err)
```

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,101 @@
...
+	assert.Contains(t, err.Error(), "does not exist in transactions")
+
+	// Fee does not have a "String" Function so we cant use const with it.
+	err = checkTagExistsAndHasCorrectFunction("const", "SignedTxnWithAD.SignedTxn.Txn.Header.Fee")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "does not contain the needed method")
+}
+
+// TestFilterProcessor_Init tests initialization of the filter processor
+func TestFilterProcessor_Init(t *testing.T) {
```
Eric-Warehime, line 35: I think we need more tests here. I don't see `all` being tested, there's not error tests for example supplying a bad tag, or a field which doesn't have the correct operations defined such as a MicroAlgos field.

Eric-Warehime (changes requested):
Really cool stuff! I left some small comments on various things. I think the big thing for me is adding more tests, and splitting the library functions out and getting individual tests for each of those interfaces/utility functions.

processors/filterprocessor/filter_processor.go
```diff
@@ -0,0 +1,305 @@
...
+	"context"
+	"fmt"
+	"gopkg.in/yaml.v3"
+	"reflect"
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/algorand/go-algorand/data/transactions"
```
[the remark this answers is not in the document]
AlgoStephenAkiki: Where should I put the order?

processors/config.go
```diff
@@ -10,3 +10,18 @@ type BlockProcessorConfig struct {
 	AlgodToken     string `yaml:"algod-token"`
 	AlgodAddr      string `yaml:"algod-addr"`
 }
+
+// FilterProcessorSubConfig is the configuration needed for each additional filter
+type FilterProcessorSubConfig struct {
+	// The tag of the struct to analyze
+	FilterTag string `yaml:"tag"`
+	// The type of expression to search for "const" or "regex"
+	ExpressionType string `yaml:"expression-type"`
```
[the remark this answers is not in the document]
winder: +1

Lets keep the plugins as self contained as possible.

processors/config.go
```diff
@@ -10,3 +10,18 @@ type BlockProcessorConfig struct {
...
+	// The expression to search
+	Expression string `yaml:"expression"`
+}
+
+// FilterProcessorConfig configuration for the filter processor
+type FilterProcessorConfig struct {
+	Filters []map[string][]FilterProcessorSubConfig `yaml:"filters"`
+}
```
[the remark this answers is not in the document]
winder: +1, let's keep the plugins self-contained.

processors/config.go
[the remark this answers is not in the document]
AlgoStephenAkiki: I've created the enum but I'm not sure if the library will honor them.  I think we still have to do our own checking.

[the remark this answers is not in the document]
winder: Same with the `map[string]`, there could be a type like `map[FilterClause]`.

processors/filterprocessor/filter_processor.go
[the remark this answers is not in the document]
Eric-Warehime: Not sure if this is defined anywhere in a best practices document, but what I was initially told was
1. golang packages
2. 3rd party packages (like logrus or the yaml library)
3. "2nd party" packages (so like algorand/msgp, algorand/go-algorand)
4. 1st party packages (anything from the same repo)

processors/filterprocessor/filter_processor.go
AlgoStephenAkiki: Done

processors/filterprocessor/filter_processor.go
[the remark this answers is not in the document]
AlgoStephenAkiki: Done

processors/filterprocessor/expression/expression.go
```diff
@@ -0,0 +1,65 @@
...
+// TypeToFunctionMap maps the expression-type with the needed function for the expression.
+// For instance the const or regex expression-type might need the String() function
+// Can't make this consts because there are no constant maps in go...
+var TypeToFunctionMap = map[Type]string{
+	ConstFilter: "String",
+	RegexFilter: "String",
+}
+
+// Interface the expression interface
+type Interface interface {
```
winder, line 27: The constructor is `MakeExpression`, should the name of this interface be Expression?

processors/filterprocessor/expression/expression.go
```diff
@@ -0,0 +1,65 @@
+package expression
+
+import (
+	"fmt"
+	"regexp"
+)
+
+// Type is the type of the filter (i.e. const, regex, etc)
+type Type string
```
Eric-Warehime, line 9: I think naming things "type" or any other language keyword is a bit confusing. Maybe `ExpressionFilterType` or `FilterType` or something like that?

AlgoStephenAkiki: Done

processors/filterprocessor/config.go
```diff
@@ -0,0 +1,18 @@
+package filterprocessor
+
+import "github.com/algorand/indexer/processors/filterprocessor/expression"
+
+// SubConfig is the configuration needed for each additional filter
+type SubConfig struct {
+	// The tag of the struct to analyze
+	FilterTag string `yaml:"tag"`
+	// The type of expression to search for "const" or "regex"
```
shiqizng, line 9: by const, do you mean exact match?

processors/filterprocessor/fields/filter.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/indexer/data"
+)
+
+// Operation an operation like "some" or "all" for boolean logic
+type Operation string
+
+const someFieldOperation Operation = "some"
+const allFieldOperation Operation = "all"
```
shiqizng, line 14: nit: should we name these Ops Or, And?

processors/filterprocessor/fields/searcher.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Search returns true if block contains the expression
+func (f Searcher) Search(input transactions.SignedTxnInBlock) bool {
+
+	e := reflect.ValueOf(&input).Elem()
+
+	var field string
+
+	for _, field = range strings.Split(f.Tag, ".") {
```
shiqizng, line 26:
```suggestion
	for _, field := range strings.Split(f.Tag, ".") {
```

processors/filterprocessor/filter_processor_test.go
AlgoStephenAkiki: I added more unit tests.  The test that detects either a bad tag or a field that doesnt have the correct function is located in searcher_test.go

processors/filterprocessor/config.go
AlgoStephenAkiki: Yeah.  Do you think I should change it to "exact"?

processors/filterprocessor/fields/searcher.go
AlgoStephenAkiki: Done

processors/filterprocessor/fields/searcher.go
```diff
@@ -0,0 +1,78 @@
...
+	Tag          string
+	MethodToCall string
+}
+
+// Search returns true if block contains the expression
+func (f Searcher) Search(input transactions.SignedTxnInBlock) bool {
+
+	e := reflect.ValueOf(&input).Elem()
+
+	var field string
```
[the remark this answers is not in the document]
AlgoStephenAkiki: Done

processors/filterprocessor/fields/filter.go
AlgoStephenAkiki: We can.  Would it be more clear to devs?

shiqizng: I think it would since and/or are commonly used operations. `matchall` and `matchany` would also be more clear.

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,272 @@
...
+    - tag: SignedTxnWithAD.SignedTxn.Txn.Header.Sender
+      expression-type: exact
+      expression: "` + sampleAddr2.String() + `"
+  - any: 
+    - tag: SignedTxnWithAD.SignedTxn.Txn.AssetTransferTxnFields.AssetCloseTo
+      expression-type: exact
+      expression: "` + sampleAddr2.String() + `"
+    - tag: SignedTxnWithAD.SignedTxn.Txn.AssetTransferTxnFields.AssetReceiver
+      expression-type: regex
+      expression: "` + sampleAddr2.String() + `"
```
shiqizng, line 47: what happens if all and any have overlapping tags?

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	sampleCfgStr := `---
+filters:
+  - any:
+    - tag: SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.Receiver
+      expression-type: regex 
+      expression: "` + sampleAddr1.String() + `"
+    - tag: SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.Receiver
+      expression-type: exact
+      expression: "` + sampleAddr2.String() + `"
```
shiqizng, line 219: what happens if I duplicate one of the tags? Will I get duplicate blocks in result?

AlgoStephenAkiki: Depends on the operator.  If it is "any" then if any of the duplicate tags expressions are true then it passes.  If it is all, then all the tag expressions must be true.  A use case for the latter would be to have a regex that specifies that an address begins and ends with some characters (although you could do that in one regex if you wanted).

processors/filterprocessor/filter_processor_test.go
AlgoStephenAkiki: The filters are evaluated in a top down order.  Meaning that the first filter that gets encountered gets executed.

winder: That's a feature isn't it? i.e. if you want to match two different accounts:
```
- any:
    - tag:  SignedTxnWithAD.SignedTxn.Txn.Sender
      expression-type: const
      expression: "AAAAAAA..."
    - tag:  SignedTxnWithAD.SignedTxn.Txn.Sender
      expression-type: const
      expression: "BBBBBBB..."
```

processors/filterprocessor/filter_processor.go
```diff
@@ -0,0 +1,134 @@
...
+				return fmt.Errorf("filter processor Init(): filter key was not a valid value: %s", key)
+			}
+
+			var searcherList []*fields.Searcher
+
+			for _, subConfig := range subConfigs {
+
+				exp, err := expression.MakeExpression(subConfig.ExpressionType, subConfig.Expression)
+				if err != nil {
+					return fmt.Errorf("filter processor Init(): could not make expression with string %s for filter tag %s - %w", subConfig.Expression, subConfig.FilterTag, err)
```
Eric-Warehime, line 86: Looks like this and the error cases of MakeExpression aren't covered.

Eric-Warehime (changes requested):
There's a few places that I think we should cover--noted them in the comments. I'm working on fixing the codecov report, but I've used the data uploaded from the most recent commit if you're curious https://app.codecov.io/github/algorand/indexer/commit/0fdcf5035e651fd933fe4321914fea7815afb204

Sorry for being paranoid about tests here, but I think because some of this code has potential to cause panics it's better to be extra careful with testing here.

processors/filterprocessor/config.go
winder, line 9: Looks like `const` this was renamed. Also, since you have comments anyway, could you add the variable names so that they matches the godoc format?

processors/filterprocessor/filter_processor_test.go
```diff
@@ -0,0 +1,272 @@
...
+	sampleAddr2 := basics.Address{2}
+	sampleAddr3 := basics.Address{3}
+
+	sampleCfgStr := `---
+filters:
+  - any: 
+    - tag: SignedTxnWithAD.SignedTxn.AuthAddr
+      expression-type: exact
+      expression: "` + sampleAddr1.String() + `"
+    - tag: SignedTxnWithAD.SignedTxn.Txn.AssetTransferTxnFields.AssetSender
```
winder, line 31: I think it's important to use the msgpack path here. Have you looked into how that would be implemented?

For example, this would be `txn.asnd`.

processors/filterprocessor/filter_processor.go
AlgoStephenAkiki: I've added a filter processor table test that addresses this but I didn't add regex expression specific tests since it would only fail on mal-formed regex.

processors/filterprocessor/filter_processor_test.go
AlgoStephenAkiki: I can take a look for sure.  I think its just different logic in the reflection package.

Eric-Warehime (approved):

PR #1221, opened 2022-09-13
Conduit: e2e tests

Eric-Warehime:
## Summary

Converts the E2E tests to a python library...for a few reasons.
1. It was already a small python library, so I've expanded that and updated it a bit.
2. It makes it easier to separate the code that is running for Indexer vs. the code running for conduit in these tests.
3. This should allow us to pull out the logic which sets up the conduit pipeline (currently in e2econduit.py) so that we anyone can declare a set of fixtures to be run as an independent scenario.

As a sidenote I've decided (for the moment) to pull the conduit tests outside of docker and have individual fixture manage docker containers (if required) in order to instantiate their prerequisite components (like postgresql).

TODO: I'll use a followup PR to generalize the importer/processor/exporter logic so that it can be defined for any set of modules.

## Test Plan

Added make conduit-e2e to the CI so these should run for every PR.


misc/e2e_tests/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "indexer_e2e_tests"
+description = "End to end tests for Algorand Indexer/Conduit"
+version = "0.0.1"
+requires-python = ">=3.8"
+dependencies = [
+    "boto3 >=1.14.9,<2",
+    "markdown2 >=2.3.9,<3",
```
algolucky, line 12: AFAIK, this is only used by release.py

misc/requirements.txt
```diff
@@ -1,8 +0,0 @@
-# packages needed for test and release
```
algolucky, line 1: there are other scripts in this directory that use this file for documenting dependencies (e.g. release.py). accounting for this removal is straight-forward on our end, but then I feel like where losing something here and moving version documentation elsewhere.

Eric-Warehime: Do you think it would be better to leave that in `misc/` and just move the E2E tests to their own top level directory? That way the existing utilities in `misc/` are unaffected by changes made to the e2e tests.

misc/e2e_tests/docker/indexer/docker-compose.yml
```diff
@@ -3,8 +3,8 @@ version: '3'
 services:    
   e2e:    
     build:    
-      context: ../.
-      dockerfile: ./misc/Dockerfile    
+      context: ../../../../.
```
tzaffi, line 6: This is now getting very very deeply nested and given the importance of E2E tests, suggests a new placement in my mind's eye. Maybe christen a new top-level directory `e2e_tests` or nestle `docker` right under `./test` ?

misc/e2e_tests/docker/indexer/Dockerfile
```diff
@@ -13,14 +13,17 @@ RUN mkdir -p /opt/go/indexer
...
+RUN rm -f /opt/go/indexer/cmd/algorand-indexer/algorand-indexer
 WORKDIR /opt/go/indexer/third_party/go-algorand
 RUN make install
 WORKDIR /opt/go/indexer
 RUN make
 ENV PATH="${HOME}/go/bin/:${PATH}"
-RUN pip3 install -r misc/requirements.txt
+COPY ./misc/e2e_tests ./e2e_tests
+WORKDIR ./e2e_tests
+RUN pip3 install ./
```
tzaffi, line 24: nice!

misc/e2e_tests/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+    "msgpack >=1,<2",
+    "py-algorand-sdk >=1.3.0,<2",
+    "pytest==6.2.5",
+    "GitPython==3.1.26",
+    "PyYAML==6.0",
+    "setuptools >=65.3.0",
```
tzaffi, line 18: Should we lock the dependencies down to specific versions? Seems a bit tedious, but given the various supply chain exploits happening these days, it might be best to follow this best practice.

misc/e2e_tests/src/e2e_conduit/fixtures/processors/block_evaluator.py
```diff
@@ -0,0 +1,25 @@
+from e2e_conduit.fixtures.plugin_fixture import PluginFixture
+
+
+class BlockEvaluator(PluginFixture):
```
tzaffi, line 4: Why doesn't `BlockEvaluator` need to implement `resolve_config_output()` ?

tzaffi (commented):
Great improvements.

I left some minor suggestions and a couple of questions.

misc/requirements.txt
algolucky: that sounds fine with me.

misc/e2e_tests/src/e2e_conduit/fixtures/exporters/postgresql.py
```diff
@@ -0,0 +1,59 @@
...
+        self.name = "postgresql"
+        self.user = "algorand"
+        self.password = "algorand"
+        self.db_name = "e2e_db"
+        # Should we have a random port here so that we can run multiple of these in parallel?
+        self.port = "45432"
+        self.container_name = ""
+        super(PostgresqlExporter, self).__init__()
+
+    def setup(self, max_conn=0):
```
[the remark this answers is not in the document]
Eric-Warehime: It's the default value in the config, and yeah we have code that translates it to the max https://github.com/algorand/indexer/blob/conduit/idb/postgres/postgres.go#L54

shiqizng (approved):

e2e_tests/src/e2e_conduit/fixtures/plugin_fixture.py
```diff
@@ -0,0 +1,17 @@
+class PluginFixture:
+    def __init__(self):
+        self.config_input = {}
+        self.config_output = {}
+
+    @property
+    def name(self):
```
tzaffi, line 7: I believe that `NotImplemented` is meant for usage when a special dunder method is not implemented, while `NotImplementedError` is meant for abstract methods that should be implemented by concrete subclasses. Some links:
* https://docs.python.org/3/library/constants.html#NotImplemented
* https://realpython.com/python-interface/#using-abstract-method-declaration

e2e_tests/src/e2e_conduit/fixtures/processors/block_evaluator.py
```diff
@@ -0,0 +1,36 @@
...
+        self.catchpoint = catchpoint
+        super().__init__()
+
+    @property
+    def name(self):
+        return "block_evaluator"
+
+    def setup(self, accumulated_config):
+        try:
+            self.algod_data_dir = accumulated_config["algod_data_dir"]
```
tzaffi, line 19: nit: usually object attributes are defined in the constructor. So consider adding `algod_data_dir`, `algod_token` etc. to `__init__()` as `None`'s or another sensible initial value.

e2e_tests/src/e2e_conduit/scenarios/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+        processors: list[PluginFixture],
+        exporter: PluginFixture,
+    ):
+        self.name = name
+        self.importer = importer
+        self.processors = processors
+        self.exporter = exporter
+        self.accumulated_config = {}
+        self.conduit_dir = ""
+
```
tzaffi, line 20:
```suggestion
from dataclasses import dataclass

@dataclass
class Scenario:
    """Data class for conduit E2E test pipelines"""
    
    name,
    importer: PluginFixture,
    processors: list[PluginFixture],
    exporter: PluginFixture,
    accumulated_config: dict = {}
    conduit_dir: str = ""

```
Take the above with a big grain of salt. But I couldn't help the temptation to suggest using [Python's dataclasses](https://realpython.com/python-data-classes/), especially because of your comment.

e2e_tests/src/e2e_conduit/scenarios/__init__.py
```diff
@@ -0,0 +1,22 @@
+from e2e_conduit.fixtures.plugin_fixture import PluginFixture
+
+
+class Scenario:
```
tzaffi, line 4: Should this class go into its own file `Scenario.py`? It's really a stylistic choice. I find it easier to know which classes are available by scanning a module directory's contents and often overlook `__init__.py`, but in some org's `__init__.py` is known to be important to always look at.

tzaffi (commented):
Made some suggestions, none of which affect functionality.

e2e_tests/src/e2e_conduit/scenarios/__init__.py
Eric-Warehime: Nice, I haven't used these before. I initially reached for NamedTuples, but they're immutable so they don't quite fit here. I'll see if dataclasses work

misc/e2e_tests/src/e2e_conduit/e2econduit.py
```diff
@@ -0,0 +1,199 @@
...
+    atexit.register(indexerdp.kill)
+    indexerout = subslurp(indexerdp.stdout)
+
+    logger.info(f"Waiting for conduit to reach round {algod_importer.lastblock}")
+
+    try:
+        indexerout.run(algod_importer.lastblock)
+    except RuntimeError as exc:
+        logger.error(f"{exc}")
+        logger.error(f"conduit hit an error during execution: {indexerout.error_log}")
```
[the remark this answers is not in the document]
Eric-Warehime: I haven't included it--the conduit logs will all be timestamped and if it's running in circleCI it will have times, etc.

misc/e2e_tests/src/e2e_conduit/fixtures/exporters/postgresql.py
```diff
@@ -0,0 +1,59 @@
...
+logger = logging.getLogger(__name__)
+
+
+class PostgresqlExporter(PluginFixture):
+    def __init__(self):
+        self.name = "postgresql"
+        self.user = "algorand"
+        self.password = "algorand"
+        self.db_name = "e2e_db"
+        # Should we have a random port here so that we can run multiple of these in parallel?
```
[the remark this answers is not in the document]
Eric-Warehime: Left this alone for now. It will need to be done if we want tests to run in parallel.

misc/e2e_tests/src/e2e_conduit/fixtures/exporters/postgresql.py
```diff
@@ -0,0 +1,59 @@
...
+        self.container_name = ""
+        super(PostgresqlExporter, self).__init__()
+
+    def setup(self, max_conn=0):
+        self.container_name = "".join(
+            random.choice(string.ascii_lowercase) for i in range(10)
+        )
+        self.max_conn = max_conn
+        try:
+            xrun(
```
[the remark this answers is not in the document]
Eric-Warehime: Looks interesting. I've not added it at the moment--just focusing on getting all of the other bits figured out here. Will see if it makes sense to swap in during a followup PR.

misc/e2e_tests/src/e2e_conduit/fixtures/processors/block_evaluator.py
Eric-Warehime: In the most recent updates you can kind of see the vision hopefully. The idea is to accumulate config values that get passed along to subsequent plugins since there are some implicit dependencies. I've just used a dict for that which gets updated by the `config_output` whereas the `config_input` is what actually gets written to the conduit config file.

Let me know if you have suggestions/comments on how to better tackle that problem in python.

e2e_tests/docker/indexer/docker-compose.yml
```diff
@@ -1,10 +1,10 @@
...
-services:
-  e2e:
-    build:
-      context: ../.
-      dockerfile: ./misc/Dockerfile
+version: '3'    
+    
+services:    
+  e2e:    
+    build:
```
tzaffi, line 5:
```suggestion
version: '3'

services:
  e2e:
    build:
```

e2e_tests/src/e2e_conduit/runner.py
```diff
@@ -0,0 +1,103 @@
...
+        self.keep_temps = keep_temps
+
+    def setup_scenario(self, scenario):
+
+        # Setup conduit_dir for conduit data dir
+        scenario.conduit_dir = tempfile.mkdtemp()
+        if not self.keep_temps:
+            atexit.register(shutil.rmtree, scenario.conduit_dir, onerror=logger.error)
+        else:
+            logger.info("leaving temp dir %r", scenario.conduit_dir)
```
tzaffi, line 30: I'm not seeing the reason to print`scenario.coundit_dir`'s `repr()` with `%r` rather than the more typical `%s` which uses `str()`.

e2e_tests/src/e2e_conduit/runner.py
```diff
@@ -0,0 +1,103 @@
...
+
+        scenario.accumulated_config = {
+            "conduit_dir": scenario.conduit_dir,
+        }
+
+        for plugin in [scenario.importer, *scenario.processors, scenario.exporter]:
+            plugin.setup(scenario.accumulated_config)
+            plugin.resolve_config_input()
+            plugin.resolve_config_output()
```
tzaffi, line 39: not even a nit: is this the expected usage pattern? If so, maybe it's worth having the base case implement a new method `resolve_config()` which would be called instead.

tzaffi (approved):
Approving even though I had some minor suggestions. Regardless, they wouldn't hold up approval.

e2e_tests/pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[build-system]
```
winder, line 1: You already mentioned this, but a README for how to use this would be helpful. Especially focusing on how to support a new plugin.

e2e_tests/docker/indexer/Dockerfile
```diff
@@ -13,14 +13,17 @@ RUN mkdir -p /opt/go/indexer
 COPY . /opt/go/indexer
 WORKDIR /opt/go/indexer
 RUN rm -f $HOME/go/bin/algod
-RUN rm /opt/go/indexer/cmd/algorand-indexer/algorand-indexer
+RUN rm -f /opt/go/indexer/cmd/algorand-indexer/algorand-indexer
```
winder, line 16: Do we need to remove the conduit binary as well?

e2e_tests/docker/indexer/Dockerfile
```diff
@@ -13,14 +13,17 @@ RUN mkdir -p /opt/go/indexer
...
 WORKDIR /opt/go/indexer/third_party/go-algorand
 RUN make install
 WORKDIR /opt/go/indexer
 RUN make
 ENV PATH="${HOME}/go/bin/:${PATH}"
-RUN pip3 install -r misc/requirements.txt
+COPY ./e2e_tests ./e2e_tests
```
winder, line 22: On line 13 we have `COPY . /opt/go/indexer`. So this directory could already be available at `/opt/go/indexer/e2e_tests`

e2e_tests/src/e2e_conduit/subslurp.py
```diff
@@ -0,0 +1,60 @@
...
+import gzip
+import io
+import logging
+import re
+
+logger = logging.getLogger(__name__)
+
+
+class subslurp:
+    # accumulate stdout or stderr from a subprocess and hold it for debugging if something goes wrong
```
winder, line 11: nit: or is this form only for functions?
```suggestion
    """ accumulate stdout or stderr from a subprocess and hold it for debugging if something goes wrong """
```

e2e_tests/src/e2e_conduit/fixtures/importers/algod.py
```diff
@@ -0,0 +1,97 @@
...
+                for f in files:
+                    if f == "node.log":
+                        p = os.path.join(root, f)
+                        logger.error("found node.log: {}".format(p))
+                        with open(p) as nf:
+                            for line in nf:
+                                logger.error("   {}".format(line))
+            raise
+
+        # atexitrun(["goal", "network", "stop", "-r", tempnet])
```
winder, line 96: If you run this locally, does the network ever get stopped?

winder (approved):
LGTM, just left a few comments for you to consider.

PR #1224, opened 2022-09-14
Replace `*string` by `ledgercore.ValueDelta{Data:}` in `kvmods`

tzaffi:
## Summary

## Test Plan

### DIFF
--- a/idb/postgres/internal/writer/writer.go
+++ b/idb/postgres/internal/writer/writer.go
@@ -303,19 +303,19 @@ func writeAccountDeltas(round basics.Round, accountDeltas *ledgercore.AccountDel
 
 }
 
-func writeBoxMods(kvMods map[string]*string, batch *pgx.Batch) error {
+func writeBoxMods(kvMods map[string]ledgercore.ValueDelta, batch *pgx.Batch) error {
 	// INSERT INTO / UPDATE / DELETE FROM `app_box`
 	// WARNING: kvMods can in theory support more general storage types than app boxes.
 	// However, here we assume that all the provided kvMods represent app boxes.
 	// If a non-box is encountered inside kvMods, an error will be returned and
 	// AddBlock() will fail with the import getting stuck at the corresponding round.
-	for key, value := range kvMods {
+	for key, valueDelta := range kvMods {
 		app, name, err := logic.SplitBoxKey(key)
 		if err != nil {
 			return fmt.Errorf("writeBoxMods() err: %w", err)
 		}
-		if value != nil {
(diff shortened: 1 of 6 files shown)


idb/postgres/internal/writer/writer.go
```diff
@@ -303,19 +303,19 @@ func writeAccountDeltas(round basics.Round, accountDeltas *ledgercore.AccountDel
...
 	// AddBlock() will fail with the import getting stuck at the corresponding round.
-	for key, value := range kvMods {
+	for key, valueDelta := range kvMods {
 		app, name, err := logic.SplitBoxKey(key)
 		if err != nil {
 			return fmt.Errorf("writeBoxMods() err: %w", err)
 		}
-		if value != nil {
-			batch.Queue(upsertAppBoxStmtName, app, []byte(name), []byte(*value))
+		if valueDelta.Data != nil {
```
tzaffi, line 317: NOTE: the indexer box update code _DOES NOT_  populate `ValueDelta.OldData`. I believe that this isn't necessary as rows are deleted from the `app_box` and no information is retained about old box values.

idb/postgres/postgres.go
```diff
@@ -48,7 +48,7 @@ func OpenPostgres(connection string, opts idb.IndexerDbOptions, log *log.Logger)
 
 	postgresConfig, err := pgxpool.ParseConfig(connection)
 	if err != nil {
-		return nil, nil, fmt.Errorf("Couldn't parse config: %v", err)
+		return nil, nil, fmt.Errorf("couldn't parse config: %v", err)
```
tzaffi, line 51: @AlgoStephenAkiki - okay with this unrelated change?

winder (approved):

idb/postgres/postgres.go
jannotti: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

idb/postgres/internal/writer/writer.go
michaeldiamant: @tzaffi As a point of clarification and to test my understanding:  I agree indexer does _not_ populate `OldData`, but I don't understand the rationale that follows.

Testing my understanding - Does the following make sense to you?

I expect indexer to be provided `OldData` populated by go-algorand indexer.  Which is to say, I don't think indexer _ever_ populates `OldData`.

michaeldiamant (approved):
@tzaffi Matches my expectations - thanks for propagating the change.

jannotti (approved):
Seems right to me.

PR #1229, opened 2022-09-19
Conduit: block generator fixes

Eric-Warehime:
## Summary

Conduit combines the catchup and follow loops which were previously separated in `fetcher.go`. It does this by calling algod's `v2/status/wait-for-block-after` endpoint before making a `GetBlock` call. 

Since block-gen didn't have this mocked, it was causing Indexer to fail requests.

Side Note: I've also increased the wait time for the health endpoint to return 200s. Not sure if this is specific to my machine, or the startup process for Conduit is significantly slower than Indexer's. 10 seconds safely allows it to become healthy for me locally.

## Test Plan

Running block-gen

### DIFF
--- a/cmd/block-generator/generator/server.go
+++ b/cmd/block-generator/generator/server.go
@@ -57,6 +57,7 @@ func MakeServerWithMiddleware(configFile string, addr string, blocksMiddleware B
 	mux.HandleFunc("/v2/accounts/", getAccountHandler(gen))
 	mux.HandleFunc("/genesis", getGenesisHandler(gen))
 	mux.HandleFunc("/report", getReportHandler(gen))
+	mux.HandleFunc("/v2/status/wait-for-block-after/", getStatusWaitHandler(gen))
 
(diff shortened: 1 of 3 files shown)


cmd/block-generator/generator/server.go
```diff
@@ -57,6 +57,7 @@ func MakeServerWithMiddleware(configFile string, addr string, blocksMiddleware B
 	mux.HandleFunc("/v2/accounts/", getAccountHandler(gen))
 	mux.HandleFunc("/genesis", getGenesisHandler(gen))
 	mux.HandleFunc("/report", getReportHandler(gen))
+	mux.HandleFunc("/v2/status/wait-for-block-after/", getStatusWaitHandler(gen))
```
winder, line 60: Fetcher had a two phase import process, it would skip `wait-for-block-after` until the first missing block.

I'm not sure how much of an impact this has to the performance.

winder (approved):

PR #1235, opened 2022-09-21
Conduit init subcommand

winder:
## Summary

Note: this is building ontop of #1232 so there are some unrelated changes in the diff.

Thinking out loud a little with this PR. I wanted to try out Conduit and getting the basic configuration setup was a little tricky.

Run `./conduit init` or `./conduit init -d /path/to/data/directory` and a new directory is created along with some instructions:

```
~$ ./conduit init -d data/conduit_data
A data directory has been created at 'data/conduit_data'.

Before it can be used, the config file needs to be updated
by setting the algod address/token and the block-dir path where
Conduit should write the block files.

Once the config file is updated, start Conduit with:
  ./conduit -d data/conduit_data
```

## Test Plan

New unit test and manual testing.

### DIFF
--- a/cmd/conduit/main.go
+++ b/cmd/conduit/main.go
@@ -152,8 +152,7 @@ func runConduitInit(path string) error {
 		return fmt.Errorf("runConduitInit(): failed to write sample config: %w", err)
 	}
 
-	fmt.Printf("A data directory has been created\n")
-	fmt.Printf("%s\n", location)
+	fmt.Printf("A data directory has been created %s.\n", location)
 	fmt.Printf("\nBefore it can be used, the config file needs to be updated\n")
 	fmt.Printf("by setting the algod address/token and the block-dir path where\n")
 	fmt.Printf("Conduit should write the block files.\n")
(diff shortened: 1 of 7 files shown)


cmd/conduit/main.go
```diff
@@ -111,14 +118,70 @@ func conduitCmd() *cobra.Command {
 		SilenceUsage: true,
 	}
 
-	cfg.Flags = conduitCmd.PersistentFlags()
+	cfg.Flags = cmd.Flags()
```
winder, line 121: This passes the flags to the sub commands, which I didn't want.

cmd/conduit/main.go
```diff
@@ -111,14 +118,70 @@ func conduitCmd() *cobra.Command {
...
 	cfg.Flags.StringVarP(&cfg.ConduitDataDir, "data-dir", "d", "", "set the data directory for the conduit binary")
 
-	return conduitCmd
+	return cmd
+}
+
+//go:embed conduit.yml.example
```
AlgoStephenAkiki, line 127: Very very cool

PR #1237, opened 2022-09-22
enhancement: prune historical data 

shiqizng:
## Summary

Adding an option for postgresql exporter to discard old transaction data by round.

## Test Plan

Adding new tests, run indexer for multiple days to validate data are removed correctly for the recurring option.

### DIFF
--- a/exporters/postgresql/postgresql_exporter.go
+++ b/exporters/postgresql/postgresql_exporter.go
@@ -1,6 +1,7 @@
 package postgresql
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/algorand/indexer/exporters/util"
@@ -27,6 +28,7 @@ type postgresqlExporter struct {
 	db             idb.IndexerDb
 	logger         *logrus.Logger
 	deleteDataChan chan uint64
+	ctx            context.Context
 }
 
 var postgresqlExporterMetadata = exporters.ExporterMetadata{
@@ -71,15 +73,16 @@ func (exp *postgresqlExporter) Init(cfg plugins.PluginConfig, logger *logrus.Log
 	if rnd, err := exp.db.GetNextRoundToAccount(); err == nil {
 		exp.round = rnd
 	}
+	exp.ctx = context.Background()
 	// if data pruning is enabled
 	if exp.cfg.Delete.Rounds > 0 {
-		dm, err := util.MakeDataManager(&exp.cfg.Delete, dbName, exp.cfg.ConnectionString, logger)
+		dm, err := util.MakeDataManager(&exp.cfg.Delete, exp.cfg.ConnectionString, logger)
 		if err != nil {
 			logger.Warnf("skipping data pruning. %w", err)
 		} else {
 			logger.Info("exporter running with data pruning mode")
 			exp.deleteDataChan = make(chan uint64)
-			go dm.Delete(exp.deleteDataChan)
+			go dm.Delete(exp.ctx, exp.deleteDataChan)
 		}
 	}
 	return err
@@ -91,6 +94,7 @@ func (exp *postgresqlExporter) Config() plugins.PluginConfig {
 }
 
 func (exp *postgresqlExporter) Close() error {
+	exp.ctx.Done()
 	if exp.db != nil {
 		exp.db.Close()
 	}
--- a/exporters/postgresql/postgresql_exporter.go
+++ b/exporters/postgresql/postgresql_exporter.go
@@ -81,7 +81,7 @@ func (exp *postgresqlExporter) Init(cfg plugins.PluginConfig, logger *logrus.Log
 	if !exp.cfg.Test && exp.cfg.Delete.Rounds > 0 {
 		exp.wg.Add(1)
 		exp.dm = util.MakeDataManager(exp.ctx, &exp.cfg.Delete, exp.db, logger)
-		go exp.dm.Delete(&exp.wg)
+		go exp.dm.Delete(&exp.wg, nil)
 	}
 	return err
 }
--- a/exporters/postgresql/postgresql_exporter.go
+++ b/exporters/postgresql/postgresql_exporter.go
@@ -1,6 +1,7 @@
 package postgresql
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/algorand/indexer/exporters/util"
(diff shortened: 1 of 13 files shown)


AlgoStephenAkiki:
How are we planning on merging the exporter changes to the conduit branch?

exporters/util/prune.go
```diff
@@ -0,0 +1,132 @@
...
+						p.logger.Warnf("exec: data pruning err: %v", err)
+						return
+					}
+				} else if p.config.Interval == disabled {
+					p.logger.Infof("Interval %d. Data pruning is disabled", p.config.Interval)
+					return
+				} else if p.config.Interval < once {
+					p.logger.Infof("Interval %d. Invalid Interval value", p.config.Interval)
+					return
+				}
```
winder, line 121: These can both be errors, let's avoid the concurrency completely unless it's required (i.e. Interval > 0)

exporters/util/prune.go
```diff
@@ -0,0 +1,132 @@
...
+			counter++
+		}
+	}
+}
+
+// Closed - if true, the data pruning process has exited and
+// stopped reading round from <-roundch
+func (p *postgresql) Closed() bool {
+	return p.close
+}
```
winder, line 132: As above, we can simplify to avoid the case of an expected early termination. The logic becomes simpler if the go routine should always live for the duration of the conduit process.

idb/postgres/postgres.go
```diff
@@ -2495,3 +2495,58 @@ func (db *IndexerDb) SetNetworkState(genesis bookkeeping.Genesis) error {
...
+		// update metastate
+		status := types.DeleteStatus{
+			// format time, "2006-01-02T15:04:05Z07:00"
+			LastPruned:  t.Format(time.RFC3339),
+			OldestRound: keep,
+		}
+		if err2 != nil {
+			return 0, fmt.Errorf("deleteTxns(): transaction delete err %w", err2)
+		}
+		err2 = db.setMetastate(tx, schema.DeleteStatusKey, string(encoding.EncodeDeleteStatus(&status)))
```
winder, line 2530: :+1:  Nice touch adding status to the metastate. We sometimes get into trouble reverse engineering a weird DB state, we should do stuff like this more often.

idb/postgres/postgres.go
```diff
@@ -2495,3 +2495,58 @@ func (db *IndexerDb) SetNetworkState(genesis bookkeeping.Genesis) error {
...
+		if err2 = tx.Commit(ctx); err2 != nil {
+			return 0, fmt.Errorf("deleteTxns(): delete transactions: %w", err2)
+		}
+		db.log.Infof("%d transactions deleted, last pruned at %s", cmd.RowsAffected(), status.LastPruned)
+		return cmd.RowsAffected(), nil
+	}
+	// retry
+	var rows int64
+	var err error
+	for i := 1; i <= 3; i++ {
```
winder, line 2544: See the `Retry` function in `dbutil.go`, it's configured so that it would only retry if there is a specific type of error. I think most errors probably don't need a retry.

idb/postgres/postgres.go
```diff
@@ -2495,3 +2495,58 @@ func (db *IndexerDb) SetNetworkState(genesis bookkeeping.Genesis) error {
 	}
 	return db.setNetworkState(nil, &networkState)
 }
+
+// DeleteTransactions removes old transactions
+// keep is the number of rounds to keep in db
+func (db *IndexerDb) DeleteTransactions(ctx context.Context, keep uint64, timeout time.Duration) (int64, error) {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
```
winder, line 2503: I don't think it makes sense to have a timeout as part of the interface here, the caller can pass in a timeout context if they want.

exporters/postgresql/postgresql_exporter.go
```diff
@@ -66,10 +75,30 @@ func (exp *postgresqlExporter) Init(cfg plugins.PluginConfig, logger *logrus.Log
...
-	if rnd, err := exp.db.GetNextRoundToAccount(); err == nil {
+	rnd, err := exp.db.GetNextRoundToAccount()
+	if err == nil {
 		exp.round = rnd
+	} else if err == idb.ErrorNotInitialized {
+		exp.round = 0
+	} else {
+		return fmt.Errorf("Init() err getting next round: %v", err)
+	}
+	exp.ctx, exp.cf = context.WithCancel(context.Background())
```
winder, line 86: hopefully we should have a context passed into this function soon. Moving this to the top of the Init function should make it easier to swap out the `context.Background` later.

winder (changes requested):
The overall logic looks good, but the concurrency needs to be simplified. I tried to leave a lot of feedback, happy to talk more if it would help.

exporters/util/prune.go
```diff
@@ -0,0 +1,132 @@
...
+	for {
+		select {
+		case <-p.ctx.Done():
+			return
+		case currentRound := <-roundch:
+			p.logger.Debugf("Delete: received round %d", currentRound)
+			if currentRound > p.config.Rounds {
+				keep := currentRound - p.config.Rounds + 1
+				if (p.config.Interval == once || p.config.Interval > 0) && counter == 0 {
+					// always run a delete at startup
```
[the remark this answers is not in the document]
shiqizng: we should still use a channel to pass the round from exporter to pruning?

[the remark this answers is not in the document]
winder: There was some discussion here which is relevant: https://docs.google.com/document/d/1SDZkdE-3FdfGSXttewtXDt7PigGwPML24fLCNYREnGM/edit?disco=AAAAecMJlV4

exporters/postgresql/postgresql_exporter.go
AlgoStephenAkiki: +1

idb/postgres/postgres.go
```diff
@@ -2495,3 +2496,64 @@ func (db *IndexerDb) SetNetworkState(genesis bookkeeping.Genesis) error {
 	}
 	return db.setNetworkState(nil, &networkState)
 }
+
+// DeleteTransactions removes old transactions
+// keep is the number of rounds to keep in db
+func (db *IndexerDb) DeleteTransactions(ctx context.Context, keep uint64) (int64, error) {
+	ctx, cancel := context.WithCancel(ctx)
```
AlgoStephenAkiki, line 2503: Why do we create another context from the passed in context?

shiqizng: updated.

exporters/util/prune.go
winder: Is the delete completely non blocking? i.e. you're still able to write new rows while deleting the old ones?

shiqizng: yes. it only locks the rows that are being deleted, writes to the table will still happen.

PR #1238, opened 2022-09-22
Changed tag format

AlgoStephenAkiki:
Resolves #1238

Changes tag format and adds numerical filtering

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 all: conduit cmd/algorand-indexer/algorand-indexer go-algorand idb/postgres/internal/schema/setup_postgres_sql.go idb/mocks/IndexerDb.go
 
 conduit: go-algorand
-	cd cmd/conduit && go build
+	go generate ./... && cd cmd/conduit && go build
 
 cmd/algorand-indexer/algorand-indexer: idb/postgres/internal/schema/setup_postgres_sql.go go-algorand
 	cd cmd/algorand-indexer && go build -ldflags="${GOLDFLAGS}"
@@ -60,7 +60,7 @@ fakepackage: go-algorand
 test: idb/mocks/IndexerDb.go cmd/algorand-indexer/algorand-indexer
 	go test -coverpkg=$(COVERPKG) ./... -coverprofile=coverage.txt -covermode=atomic ${TEST_FLAG}
 
-lint: go-algorand
+lint: conduit
 	golint -set_exit_status ./...
 	go vet ./...
 
--- a/processors/filterprocessor/expression/expression.go
+++ b/processors/filterprocessor/expression/expression.go
@@ -2,7 +2,10 @@ package expression
 
 import (
 	"fmt"
+	"github.com/algorand/go-algorand/data/basics"
+	"reflect"
 	"regexp"
+	"strconv"
 )
 
 // FilterType is the type of the filter (i.e. const, regex, etc)
@@ -13,31 +16,51 @@ const (
 	ExactFilter FilterType = "exact"
 	// RegexFilter a filter that applies regex rules to the matching
 	RegexFilter FilterType = "regex"
+
+	// LessThanFilter a filter that applies numerical less than operation
+	LessThanFilter FilterType = "less-than"
+	// LessThanEqualFilter a filter that applies numerical less than or equal operation
+	LessThanEqualFilter FilterType = "less-than-equal"
+	// GreaterThanFilter a filter that applies numerical greater than operation
+	GreaterThanFilter FilterType = "greater-than"
+	// GreaterThanEqualFilter a filter that applies numerical greater than or equal operation
+	GreaterThanEqualFilter FilterType = "greater-than-equal"
+	// EqualToFilter a filter that applies numerical equal to operation
+	EqualToFilter FilterType = "equal"
+	// NotEqualToFilter a filter that applies numerical NOT equal to operation
+	NotEqualToFilter FilterType = "not-equal"
 )
 
 // TypeToFunctionMap maps the expression-type with the needed function for the expression.
 // For instance the exact or regex expression-type might need the String() function
+// A blank string means a function is not required
 // Can't make this const because there are no constant maps in go...
 var TypeToFunctionMap = map[FilterType]string{
-	ExactFilter: "String",
-	RegexFilter: "String",
+	ExactFilter:            "String",
+	RegexFilter:            "String",
+	LessThanFilter:         "",
+	LessThanEqualFilter:    "",
+	GreaterThanFilter:      "",
+	GreaterThanEqualFilter: "",
+	EqualToFilter:          "",
+	NotEqualToFilter:       "",
 }
 
 // Expression the expression interface
 type Expression interface {
-	Search(input interface{}) bool
+	Search(input interface{}) (bool, error)
 }
 
 type regexExpression struct {
 	Regex *regexp.Regexp
 }
 
-func (e regexExpression) Search(input interface{}) bool {
-	return e.Regex.MatchString(input.(string))
+func (e regexExpression) Search(input interface{}) (bool, error) {
+	return e.Regex.MatchString(input.(string)), nil
 }
 
 // MakeExpression creates an expression based on an expression type
-func MakeExpression(expressionType FilterType, expressionSearchStr string) (*Expression, error) {
+func MakeExpression(expressionType FilterType, expressionSearchStr string, targetKind reflect.Kind) (*Expression, error) {
 	switch expressionType {
 	case ExactFilter:
 		{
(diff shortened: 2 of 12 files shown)


processors/filterprocessor/gen/generate.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"bytes"
+	"fmt"
+	"go/format"
+	"os"
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/algorand/go-algorand/data/transactions"
```
winder, line 12: Does this work if you use the transaction object from the go SDK?

processors/filterprocessor/fields/searcher.go
```diff
@@ -1,12 +1,12 @@
 package fields
 
+//go:generate go run ../gen/generate.go fields ./generated_signed_txn_map.go
```
winder, line 3: Could you check-in generated_signed_txn_map.go?

processors/filterprocessor/gen/generate.go
```diff
@@ -0,0 +1,128 @@
...
+
+	initialStr := `// Code generated via go generate. DO NOT EDIT.
+
+package %s
+
+import "github.com/algorand/go-algorand/data/transactions"
+
+// SignedTxnMap generates a map with the key as the codec tag and the value as a function
+// that returns the associated variable
+var SignedTxnMap = map[string]func(*transactions.SignedTxnInBlock) interface{}{
```
winder, line 86: Could you create a benchmark to compare the performance of a switch statement vs the map of functions? From our conversation I was expecting something like
```go
switch tag {
case "txn.snd":
    return t.SignedTxn.Sender
case "txn.rcv":
...
```

The top 2 results for "golang map vs switch" seem to say opposite things:
https://hashrocket.com/blog/posts/switch-vs-map-which-is-the-better-way-to-branch-in-go
https://stackoverflow.com/questions/46789259/map-vs-switch-performance-in-go

I wonder if the extra overhead of a function call would change things.

processors/filterprocessor/gen/generate.go
AlgoStephenAkiki: Perhaps.  Is it easy to test by importing the sdk?

winder: Indexer uses the SDK, so I think you could swap in the other type: https://github.com/algorand/go-algorand-sdk/blob/develop/types/transaction.go#L25

It looks like `SignedTxnWithAD` isn't available, but I'm not sure that one is especially useful so maybe this is OK?

processors/filterprocessor/gen/generate.go
```diff
@@ -0,0 +1,145 @@
...
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/algorand/go-algorand/data/transactions"
+)
+
+// recursiveTagFields recursively gets all field names in a struct
+// Output will contain a key of the full tag along with the fully qualified struct
+func recursiveTagFields(theStruct interface{}, Output *map[string]string, tagLevel []string, fieldLevel []string) {
```
tzaffi, line 17: Does capitalizing `Output` provide any functional difference or follow some golang convention? A minute of googling didn't reveal anything to me.

tzaffi: I believe it is more idiomatic to pass maps directly as they are reference types, so no need to pass a pointer to a `map[string]string`.

processors/filterprocessor/gen/generate.go
```diff
@@ -0,0 +1,145 @@
...
+// go run generate.go packagename outputfile
+func main() {
+
+	var packageName string
+	var outputFilepath string
+
+	if len(os.Args) == 3 {
+		packageName = os.Args[1]
+		outputFilepath = os.Args[2]
+	}
```
tzaffi, line 63: should there be and `else` case here that exits gracefully?

processors/filterprocessor/gen/generate.go
```diff
@@ -0,0 +1,145 @@
...
+	sort.Strings(keys)
+
+	_, err = fmt.Fprintf(&bb, "switch tag {\n")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to write to %s: %v\n", outputFilepath, err)
+		os.Exit(1)
+	}
+
+	for _, k := range keys {
+		fmt.Fprintf(&bb, "case \"%s\":\nreturn &((*input).%s), nil\n", k, output[k])
```
tzaffi, line 114:
```suggestion
		fmt.Fprintf(&bb, "case \"%s\":\nreturn &(input.%s), nil\n", k, output[k])
```

I believe go's reference semantics make the following suggestions equivalent (and arguably a bit easier to read)

tzaffi: playground example: https://go.dev/play/p/kIB_h2s7UAQ

processors/filterprocessor/gen/generate.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+"fmt"
+
+"github.com/algorand/go-algorand/data/transactions"
+)
+
+// SignedTxnFunc takes a tag and associated SignedTxnInBlock and returns the value 
+// referenced by the tag.  An error is returned if the tag does not exist
+func SignedTxnFunc(tag string, input *transactions.SignedTxnInBlock) (interface{}, error) {
```
tzaffi, line 90:
```suggestion
func SignedTxnFunc(tag string, input *transactions.SignedTxnInBlock) (typePath interface{}, err error) {
```

By naming the output variable one can achieve a slightly simpler switch statement. For example, the individual clauses could be re-written as:
```go
case "txn.sp.S.td":
		typePath = &input.SignedTxnWithAD.SignedTxn.Txn.StateProofTxnFields.StateProof.SigProofs.TreeDepth
```

while the error default case would look like

```go
default:
		err = fmt.Errorf("unknown tag: %s", tag)
```

You may need a `return` after the entire switch though.

Makefile
```diff
@@ -60,7 +60,7 @@ fakepackage: go-algorand
 test: idb/mocks/IndexerDb.go cmd/algorand-indexer/algorand-indexer
 	go test -coverpkg=$(COVERPKG) ./... -coverprofile=coverage.txt -covermode=atomic ${TEST_FLAG}
 
-lint: go-algorand
+lint: conduit
```
tzaffi, line 63: nice!

processors/filterprocessor/fields/searcher.go
```diff
@@ -1,12 +1,12 @@
...
+
 import (
 	"fmt"
-	"reflect"
-	"strings"
-
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/indexer/processors/filterprocessor/expression"
+	"reflect"
```
tzaffi, line 9: nit: I think the `reflect` import should go back where it came from - right after the `fmt` import

tzaffi (commented):
Looks good. Suggested minor code simplifications.

processors/filterprocessor/gen/generate.go
AlgoStephenAkiki: Done

processors/filterprocessor/gen/generate.go
AlgoStephenAkiki: The way the logic works is that we build up the "string" (of which only needs the file path).  Then at the end we decide if we print to stdout or the file itself.  The reason for that was running the generate command by itself will allow you to see its output without you having to run "go generate" over and over.

processors/filterprocessor/gen/generate.go
AlgoStephenAkiki: Done

processors/filterprocessor/gen/generate.go
winder: Good point about the IDs, that might be the only place it is available when things are being created.

processors/filterprocessor/fields/searcher.go
AlgoStephenAkiki: Done

processors/filterprocessor/gen/generate.go
AlgoStephenAkiki: I'll make a note, but I think this might be fine in its current format unless this has substantial performance benefits.

tzaffi (approved):
LGTM

tzaffi:
> Resolves #1238
> 
> Changes tag format and adds numerical filtering

Sorry, which issue is this actually resolving?

processors/filterprocessor/expression/expression.go
```diff
@@ -59,6 +83,69 @@ func MakeExpression(expressionType FilterType, expressionSearchStr string) (*Exp
...
+	case GreaterThanEqualFilter:
+		fallthrough
+	case EqualToFilter:
+		fallthrough
+	case NotEqualToFilter:
+		{
+			var exp Expression
+
+			switch targetKind {
+			case reflect.TypeOf(basics.MicroAlgos{}).Kind():
```
winder, line 102: You could put this check in the generator. If the leaf type is MicroAlgos add the extra Raw reference:
```diff
	case "txn.fee":
-		return &input.SignedTxnWithAD.SignedTxn.Txn.Header.Fee, nil
+		return &input.SignedTxnWithAD.SignedTxn.Txn.Header.Fee.Raw, nil
```

winder (approved):
LGTM

PR #1240, opened 2022-09-23
Fix inaccurate `BoxesResponse` `next-token` documentation

jasonpaulos:
## Summary

I noticed the `BoxesResponse` API definition has an inaccurate description for the `next-token` field. The current description implies that the user must manually prefix the result with `b64:` before using it in a subsequent request. Perhaps at one time we considered this implementation, but my current understanding is that that step is not necessary due to this code:

https://github.com/algorand/indexer/blob/35e87fa72773dc6503f8aa23012ab0876e3f5d27/api/handlers.go#L673-L681

To fix the inaccuracy and maintain consistency with other definitions, I've adopted the existing description that all other `next-token` fields use.

## Test Plan

No code changes.


tzaffi (approved):
Thanks for catching this inaccuracy. In a previous discussion I had committed to reverting the way next token is communicated to the way it is done throughout indexer, but I didn't find this vestige!

Issue #332, opened 2021-02-17
How to return blocks with many transactions


winder:
## Summary
In the future, we expect to see blocks with many thousands of transactions. Returning thousands of transactions is unnecessary overhead for anyone who just wants to get the block header.

## Scope/Requirements
There are a couple approaches:
1) Simply remove the transaction array from the block header, users that want the transactions can use the transactions endpoint. Combined with #327 it is easy to verify that all transactions are accounted for.

2) Conditionally enable the transactions-in-block feature for some deployments which can accept the overhead. With already enable some features with the `--dev-mode` feature, this could be more fine grained now that we have a configuration file.

## Urgency/Relative Priority
It's harder to remove features than add them, so setting the precedent now would be a good idea.

# Acceptance Criteria

1) Add a new `header-only` option to the blocks endpoint which skips the transaction query.
2) When building the block response and querying for transactions, set the transactions limit to the server setting for max number of transactions.
3) If the transactions response contains more transactions than the limit, return an error. Follow the pattern used for the Accounts endpoint, which generates an error if the account has too many assets and/or applications. The error should indicate that the `header-only` flag should be enabled.

brianolson:
At an API level do something like this? 
`GET /v2/blocks/{block number}?header-only=1`

PR #1241, opened 2022-09-23
REST API: Add Header-Only parameter to blocks endpoint, return error if there are too many transactions.

shiqizng:
## Summary

#332 Adding header-only parameter to GET /v2/blocks/{round-number} which would return a block excluding the transactions. If the number of the transactions in a request block exceeds MaxTransactionsLimit, the endpoint return err code 400 when header-only parameter isn't set. 

## Test Plan
new unit tests and integration tests.


AlgoStephenAkiki (approved):

idb/idb.go
```diff
@@ -401,3 +408,11 @@ type Health struct {
 type NetworkState struct {
 	GenesisHash crypto.Digest `codec:"genesis-hash"`
 }
+
+// MaxTransactionsError records the error when transaction counts exceeds MaxTransactionsLimit.
+type MaxTransactionsError struct {
+}
+
+func (e MaxTransactionsError) Error() string {
+	return "Max transactions limit exceeded. header-only flag should be enabled"
```
winder, line 417: Could you put this string in `error_messages.go`?

idb/postgres/postgres_integration_test.go
```diff
@@ -375,8 +375,7 @@ func TestBlockWithTransactions(t *testing.T) {
 	//////////
 	// When // We call GetBlock and Transactions
 	//////////
-	_, txnRows0, err := db.GetBlock(
-		context.Background(), round, idb.GetBlockOptions{Transactions: true})
+	_, txnRows0, err := db.GetBlock(context.Background(), round, idb.GetBlockOptions{Transactions: true, MaxTransactionsLimit: 100})
```
winder, line 378: Could you add another test which triggers the MaxTransactionError case?

idb/idb.go
shiqizng: added.

idb/postgres/postgres.go
```diff
@@ -480,6 +481,9 @@ func (db *IndexerDb) GetBlock(ctx context.Context, round uint64, options idb.Get
 		results := make([]idb.TxnRow, 0)
 		for txrow := range out {
 			results = append(results, txrow)
+			if uint64(len(results)) > options.MaxTransactionsLimit {
```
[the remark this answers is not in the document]
shiqizng: I think with the limit filter added, we're ok with `out` channel. When line 484 is true, `out` would be empty.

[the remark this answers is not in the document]
winder: I think you're right, but the return inside the range seems dangerous. Here's an example of clearing out the channel (I thought there were others but this is the only one I could find): https://github.com/algorand/indexer/blob/develop/api/handlers.go#L1216

Could you move the if statement to the outside of the range statement? It should still be efficient, thanks to the limit, but would also ensure that the channel is closed.

winder (approved):

PR #1251, opened 2022-09-28
Add panic recovers

### LINKED ISSUE #263: Catch Panic's and Write To Log
## Summary

If a panic occurs make sure it gets written to the log file.

AlgoStephenAkiki:
Resolves #263

Adds recovery functions in main loop of conduit as well as in a go-routine in the start function of the pipeline

## Summary

## Test Plan

### DIFF
--- a/cmd/conduit/main.go
+++ b/cmd/conduit/main.go
@@ -50,6 +50,11 @@ func init() {
 
 // runConduitCmdWithConfig run the main logic with a supplied conduit config
 func runConduitCmdWithConfig(cfg *conduit.Config) error {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Fatalf("conduit framework experienced a panic: %v", r)
+		}
+	}()
 
 	// From docs:
 	// BindEnv takes one or more parameters. The first parameter is the key name, the rest are the name of the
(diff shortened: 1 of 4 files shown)


cmd/conduit/main.go
```diff
@@ -50,6 +50,11 @@ func init() {
 
 // runConduitCmdWithConfig run the main logic with a supplied conduit config
 func runConduitCmdWithConfig(cfg *conduit.Config) error {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Fatalf("conduit framework experienced a panic: %v", r)
+		}
+	}()
```
winder, line 57: Do you think it's worth making a helper function? I'm thinking that could be something we document / reuse in plugins.

```
	func HandlePanicAndFail(logger *log.Logger) {
		if r := recover(); r != nil {
			logger.Panicf("conduit framework experienced a panic: %v", r)
		}
	}
```

Also, it looks like logrus has a "Panic" level. The difference compared to Fatal is that it calls panic after logging instead of Exit.

AlgoStephenAkiki: Re-worked, let me know what you think.

winder (approved):
LGTM

PR #1253, opened 2022-09-30
Conduit: filereader / filewriter plugins and compatibility changes.

winder:
## Summary

* New filereader plugin.
* Small changes / improvements to the filewriter plugin:
  1. Add tags to the BlockData type so that the serialization is explicit.
  2. Change the block encoder function to use go-codec, the `StateDelta` object contains a map with non-string keys which caused a problem with the json encoder.
  3. Use t.TempDir() for tests so that the tests don't leave artifacts behind.
  4. Save a genesis.json file
  5. Some new configurations - Removing the vote certificate (should be in the filter plugin?), specifying the filename format.

## Test Plan

New unit tests.

### DIFF
--- a/importers/filereader/filereader_test.go
+++ b/importers/filereader/filereader_test.go
@@ -1,0 +1,189 @@
+package fileimporter
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+
+	"github.com/algorand/indexer/data"
+	"github.com/algorand/indexer/exporters/filewriter"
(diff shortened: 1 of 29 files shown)


util/ledger_util.go
```diff
@@ -64,6 +65,10 @@ func createInitState(genesis *bookkeeping.Genesis) (ledgercore.InitState, error)
 // MakeLedger opens a ledger, initializing if necessary.
 func MakeLedger(logger *log.Logger, inMemory bool, genesis *bookkeeping.Genesis, dataDir string) (*ledger.Ledger, error) {
 	const prefix = "ledger"
+	err := os.MkdirAll(dataDir, 0755)
+	if err != nil {
+		return nil, fmt.Errorf("MakeProcessor() failed to create '%s': %w", dataDir, err)
+	}
```
winder, line 71: Automatically create the local ledger directory if it's missing. While I was testing different configurations, this kept coming up.

importers/filereader/filereader_test.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func TestConfigDefault(t *testing.T) {
+	testImporter = New()
+	expected, err := yaml.Marshal(&Config{})
+	require.NoError(t, err)
+	assert.Equal(t, plugins.PluginConfig(expected), testImporter.Config())
+}
+
+// TODO: remaining tests
```
shiqizng, line 122: this can be removed?

shiqizng:
LGTM

PR #1255, opened 2022-10-03
fix: use CI_E2E_FILENAME in s3 prefix 

shiqizng:
## Summary
This PR fixes a string parsing error in e2e test script. commit hash wasn't included in the s3 file prefix and so e2e test was only using the latest artifact from nightly build. 

## Test Plan
check CI build to make sure test artifact is downloaded from the correct file path.

### DIFF
--- a/misc/e2elive.py
+++ b/misc/e2elive.py
@@ -83,12 +83,16 @@ def main():
         from botocore.config import Config
         from botocore import UNSIGNED
         s3 = boto3.client("s3", config=Config(signature_version=UNSIGNED))
+        prefix = "indexer/e2e4"
         if "/" in tarname:
-            tarname = tarname.split("/")[1]
+            cmhash_tarnme = tarname.split("/")
+            cmhash = cmhash_tarnme[0]
+            tarname =cmhash_tarnme[1]
+            prefix+="/"+cmhash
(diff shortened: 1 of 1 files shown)


misc/e2elive.py
```diff
@@ -83,12 +83,16 @@ def main():
...
         from botocore import UNSIGNED
         s3 = boto3.client("s3", config=Config(signature_version=UNSIGNED))
+        prefix = "indexer/e2e4"
         if "/" in tarname:
-            tarname = tarname.split("/")[1]
+            cmhash_tarnme = tarname.split("/")
+            cmhash = cmhash_tarnme[0]
+            tarname =cmhash_tarnme[1]
+            prefix+="/"+cmhash
             tarpath = os.path.join(tempdir, tarname)
```
winder, line 92: Could you elaborate on whats happening here?

If I'm reading right it looks like there were multiple slashes in the name and this supports a tarname with 1 slash. Should we support multiple slashes?

shiqizng: I don't think it's possible to support filenames with `/` without having to update the prefix.

AlgoStephenAkiki (approved):

winder (approved):

PR #1256, opened 2022-10-03
Condensed Duplicate logger output

### LINKED ISSUE #1245: Conduit: Duplicate logger output
When starting Conduit with no arguments, there are duplicate error outputs:

```
{"__type":"Conduit","_name":"main","level":"info","msg":"Data Directory:  ","time":"2022-09-16T15:50:51-04:00"}
Error: supplied data directory was empty
ERRO[0000] supplied data directory was empty
```

Some initial investigation:
```
The first line is us trying to log the config after we've loaded it: https://github.com/algorand/indexer/blob/conduit/cmd/conduit/main.go#L69

The Error message is the validity function finding the data directory in the config empty: https://github.com/algorand/indexer/blob/conduit/conduit/config.go#L27

And the third line is probably from https://github.com/algorand/indexer/blob/conduit/cmd/conduit/main.go#L122
Seems like we've got a bunch of duplicate logs of the same thing I guess? Since we are using a logger initialized in the init function we should probably always be using that.
```

We are also using `RunE` instead of `Run` in conduit, which is probably where one of the error outputs comes from. But it is not clear to me why the format is different.

AlgoStephenAkiki:
Resolves #1245 

Explicitly spell out empty data directory as well as condensed error messages when running conduit

## Summary

## Test Plan

### DIFF
--- a/cmd/conduit/main.go
+++ b/cmd/conduit/main.go
@@ -117,6 +117,8 @@ func makeConduitCmd() *cobra.Command {
 			return runConduitCmdWithConfig(cfg)
 		},
 		SilenceUsage: true,
+		// Silence errors because our logger will catch and print any errors
+		SilenceErrors: true,
 	}
 
 	cfg.Flags = cmd.Flags()
@@ -183,7 +185,7 @@ func makeInitCmd() *cobra.Command {
 
 func main() {
 	if err := conduitCmd.Execute(); err != nil {
-		log.Errorf("%v", err)
+		logger.Errorf("%v", err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/conduit/main.go
```diff
@@ -117,6 +117,8 @@ func makeConduitCmd() *cobra.Command {
 			return runConduitCmdWithConfig(cfg)
 		},
 		SilenceUsage: true,
+		// Silence errors because our logger will catch and print any errors
+		SilenceErrors: true,
```
winder, line 121: Whats the difference between this and changing from `RunE` to `Run`?

AlgoStephenAkiki: I think the big difference is that we don't need to call `os.Exit(1)` inside the command function.  Instead we can return an error (which causes defer'd functions to run) and return an appropriate error code to the command line at the main function level.

winder (approved):

PR #1258, opened 2022-10-05
adding block metadata to pipeline

shiqizng:
## Summary

adding metadata to pipeline

### DIFF
--- a/cmd/conduit/conduit.yml.example
+++ b/cmd/conduit/conduit.yml.example
@@ -10,6 +10,11 @@ Importer:
 # One or more processors may be defined to manipulate what data
 # reaches the exporter.
 Processors:
+    Name: block_evaluator
+    Config:
+        - indexer-data-dir: "path/to/indexer-data-dir"
+          algod-addr: "importer algod address here"
+          algod-token: "importer algod token here"
 
 # An exporter is defined to do something with the data.
 # Here the filewriter is defined which writes the raw block
--- a/cmd/conduit/conduit.yml.example
+++ b/cmd/conduit/conduit.yml.example
@@ -10,11 +10,6 @@ Importer:
 # One or more processors may be defined to manipulate what data
 # reaches the exporter.
 Processors:
-    Name: block_evaluator
-    Config:
-        - indexer-data-dir: "path/to/indexer-data-dir"
-          algod-addr: "importer algod address here"
-          algod-token: "importer algod token here"
 
 # An exporter is defined to do something with the data.
 # Here the filewriter is defined which writes the raw block
--- a/conduit/pipeline.go
+++ b/conduit/pipeline.go
@@ -148,12 +148,12 @@ type pipelineImpl struct {
 	processors []*processors.Processor
 	exporter   *exporters.Exporter
 
-	blockMetadata         BlockMetaData
-	blockMetadataFilePath string
+	pipelineMetadata         PipelineMetaData
+	pipelineMetadataFilePath string
 }
 
-// BlockMetaData contains the metadata for block file storage
-type BlockMetaData struct {
+// PipelineMetaData contains the metadata for the pipeline
+type PipelineMetaData struct {
 	GenesisHash string `json:"genesis-hash"`
 	Network     string `json:"network"`
 	NextRound   uint64 `json:"next-round"`
@@ -200,13 +200,6 @@ func (p *pipelineImpl) Init() error {
 
 	// TODO Need to change interfaces to accept config of map[string]interface{}
 
-	// load pipeline metadata
-	var err error
-	p.blockMetadata, err = p.loadBlockMetadata()
-	if err != nil {
-		return fmt.Errorf("Pipeline.Start(): could not read metadata: %w", err)
-	}
-
 	// Initialize Importer
 	importerLogger := log.New()
 	importerName := (*p.importer).Metadata().Name()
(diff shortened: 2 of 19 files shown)


conduit/pipeline.go
```diff
@@ -187,56 +200,81 @@ func (p *pipelineImpl) Init() error {
...
-	// Initialize Exporter first since the pipeline derives its round from the Exporter
-	exporterLogger := log.New()
-	exporterLogger.SetFormatter(makePluginLogFormatter(plugins.Exporter, (*p.exporter).Metadata().Name()))
-
-	jsonEncode := string(json.Encode(p.cfg.Exporter.Config))
-	err := (*p.exporter).Init(plugins.PluginConfig(jsonEncode), exporterLogger)
-	exporterName := (*p.exporter).Metadata().Name()
+	// load pipeline metadata
+	var err error
+	p.blockMetadata, err = p.loadBlockMetadata()
```
winder, line 205: nit: `initializeOrLoadBlockMetadata` might be a better name. While reading, I wondered how the metadata file would be initialized.

winder: Is it possible to initialize this after the importer, so that it can have access to the genesis file?

conduit/pipeline.go
```diff
@@ -143,7 +147,16 @@ type pipelineImpl struct {
...
 	processors []*processors.Processor
 	exporter   *exporters.Exporter
-	round      basics.Round
+
+	blockMetadata         BlockMetaData
+	blockMetadataFilePath string
+}
+
+// BlockMetaData contains the metadata for block file storage
+type BlockMetaData struct {
```
winder, line 156: nit: could you call this `Metadata` or `PipelineMetadata` to indicate that it has more than just "block" metadata?

exporters/postgresql/postgresql_exporter.go
```diff
@@ -66,9 +64,8 @@ func (exp *postgresqlExporter) Init(cfg plugins.PluginConfig, logger *logrus.Log
 	}
 	exp.db = db
 	<-ready
-	if rnd, err := exp.db.GetNextRoundToAccount(); err == nil {
-		exp.round = rnd
-	}
+	importer.EnsureInitialImport(exp.db, *initProvider.GetGenesis())
+	exp.round = uint64(initProvider.NextDBRound())
```
winder, line 68: You can return an error here if `NextDBRound() != GetNextRoundToAccount()`

winder (changes requested):
Just a few small suggestions, otherwise the changes look great.

conduit/pipeline.go
```diff
@@ -339,12 +369,13 @@ func (p *pipelineImpl) Start() {
...
 					// Ignore round 0 (which is empty).
-					if p.round > 0 {
+					if p.pipelineMetadata.NextRound > 0 {
 						p.addMetrics(blkData, importTime)
 					}
 					// Increment Round
 					p.setError(nil)
-					p.round++
+					p.pipelineMetadata.NextRound++
+					p.encodeMetadataToFile()
```
Eric-Warehime, line 378: This is potentially a huge bottleneck on how quickly we can fully catchup from 0 on a new Indexer node. Not only are we doing blocking file IO between each round, but we are also creating a temp file and renaming it.

An idea: Maybe send completed rounds to a channel which is read by a goroutine that asynchronously keeps the metadata file up to date. We would need to ensure that the channel is fully flushed before properly exiting Conduit otherwise DB and metadata could be unsynched. If the process does not terminate gracefully of course they will have that problem anyways. It's potentially a bit more complex, but it seems to me like we might want that complexity here to get these operations off the main run loop.

conduit/pipeline_test.go
```diff
@@ -422,3 +433,123 @@ func TestPipelineErrors(t *testing.T) {
...
+				{
+					Name:   "",
+					Config: map[string]interface{}{},
+				},
+			},
+			Exporter: NameConfigPair{
+				Name:   "",
+				Config: map[string]interface{}{},
+			},
+		},
```
Eric-Warehime, line 465: Minor: It's potentially worth adding this to a test util since I see us using an empty config in multiple places.

Eric-Warehime (commented):
Would like to get more discussion on the file io happening every conduit loop.

Also would be good to get some more test coverage--many files show decreases in coverage %.

cmd/conduit/conduit.yml.example
```diff
@@ -10,6 +10,11 @@ Importer:
 # One or more processors may be defined to manipulate what data
 # reaches the exporter.
 Processors:
+    Name: block_evaluator
+    Config:
+        - indexer-data-dir: "path/to/indexer-data-dir"
+          algod-addr: "importer algod address here"
+          algod-token: "importer algod token here"
```
winder, line 17:
```suggestion
#    optional plugin to add in additional information in the block files.
#    Name: block_evaluator
#    Config:
#        - indexer-data-dir: "path/to/indexer-data-dir"
#          algod-addr: "importer algod address here"
#          algod-token: "importer algod token here"
```

conduit/pipeline_test.go
shiqizng: I'm leaving this as is because moving PipelineConfig to test util creates a dependency cycle.

conduit/pipeline.go
shiqizng: I'm leaving this unchanged for now. I ran a 15min block generator test locally and the average import time is no worse than one in conduit branch. I will set up an ec2 to run a longer test.

winder (approved):

PR #1271, opened 2022-10-12
Conduit: ensure thread safety across multiple loggers.

### LINKED ISSUE #1245: Conduit: Duplicate logger output

AlgoStephenAkiki:
Resolves #1245 
    
 Adds a multi-threaded logger instance for conduit usage


importers/importer.go
```diff
@@ -17,7 +17,7 @@ type Importer interface {
 	// Init will initialize each importer with a given config. This config will contain the Unmarhsalled config file specific to this plugin.
 	// It is called during initialization of an importer plugin such as setting up network connections, file buffers etc.
 	// Importers will also be responsible for returning a valid Genesis object pointer
-	Init(ctx context.Context, cfg plugins.PluginConfig, logger *logrus.Logger) (*bookkeeping.Genesis, error)
+	Init(ctx context.Context, cfg plugins.PluginConfig, logger *loggers.MT) (*bookkeeping.Genesis, error)
```
winder, line 20: Could you help me understand why the new interface is needed here? Does the plugin writer need to know that we've done something special with the logger?

loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+
+// LoggerManager a manager that can produce loggers that are synchronized internally
+type LoggerManager struct {
+	internalWriter ThreadSafeWriter
+}
+
+// AdaptLogger will take a logger instance and wrap it for synchronization
+func AdaptLogger(log *log.Logger) *LoggerManager {
+	return MakeLoggerManager(log.Out)
+}
```
winder, line 23: I don't think you need this function:
* If `MakeLogger` returns log.Logger, then you don't need to change all the interfaces.
* Now you don't need `MakeLoggerManager` to wrap the writer, all the tests could continue using `log.New`.

loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+			Mutex:  &sync.Mutex{},
+		},
+	}
+}
+
+// ThreadSafeWriter a struct that implements io.Writer in a threadsafe way
+type ThreadSafeWriter struct {
+	Writer io.Writer
+	Mutex  *sync.Mutex
+}
```
winder, line 46: Are there any tests that could be written for this? i.e. Take a logger that uses the ThreadSafeWriter and fire a bunch of logs at it. Maybe also a good opportunity to write a very large log message and see that it isn't broken into parts?

loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Write writes p bytes with the mutex
+func (w ThreadSafeWriter) Write(p []byte) (n int, err error) {
+	w.Mutex.Lock()
+	defer w.Mutex.Unlock()
+	return w.Writer.Write(p)
+}
```
winder, line 53: that should do it! :+1:

loggers/loggers.go
AlgoStephenAkiki: See updated test

loggers/loggers_test.go
```diff
@@ -0,0 +1,89 @@
...
+		go func() {
+			// Sleep a random number of milliseconds before and after to test
+			// that creating a logger doesn't affect thread-safety
+			time.Sleep(time.Duration(rand.Intn(50)) * time.Millisecond)
+			l := lMgr.MakeLogger()
+			l.SetFormatter(&log.JSONFormatter{
+				// We want to disable timestamps to stop logrus from sorting our output
+				DisableTimestamp: true,
+			})
+			time.Sleep(time.Duration(rand.Intn(50)) * time.Millisecond)
```
winder, line 50: One trick for starting all of the loggers at the same time is to have them all wait on the same wait group. That you mark as Done once the goroutines have been created

loggers/loggers_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"testing"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+)
+
+type FakeIoWriter struct {
+	// A list of strings that represent ints
+	ListOfInts []string
```
winder, line 17: Should you allocate the array to the expected size?

winder (approved):
Just had a couple minor comments for you to consider. Looks good.

loggers/loggers_test.go
AlgoStephenAkiki: I want to make sure that they don't start at the same time though.  I want to make sure that making the logger doesn't affect thread-safety.

loggers/loggers_test.go
AlgoStephenAkiki: The slice is easier from a testing perspective since I don't have to keep track of the latest entry (I can simply append).  Plus I dont think the performance is that bad with the current test.  We can always change it if we want to later.

shiqizng (approved):

PR #1282, opened 2022-10-19
testing: Validator improvements.

winder:
## Summary

* Report the canonical address name even if the b64 address is used.
* Avoid reporting skipped accounts unless using new '--print-skipped' flag.
* Fix REST API error encoding.
* Detect algod result limit reached errors.

## Test Plan

Added unit tests.


algobarb (approved):
Tested this although I didn't have an account with an error to check there's no issues there, I was able to check the "--print-skipped" flag.

shiqizng (approved):

PR #1284, opened 2022-10-19
adding metrics endpoint

shiqizng:
## Summary
adding /metrics endpoint to conduit and adding metrics to pipeline. 

## Test Plan
run conduit and check /metrics endpoint

### DIFF
--- a/cmd/conduit/main.go
+++ b/cmd/conduit/main.go
@@ -7,18 +7,15 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/algorand/indexer/util/metrics"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	"github.com/algorand/indexer/conduit"
-	"github.com/algorand/indexer/loggers"
-)
-
-import (
-	// We need to import these so that the package wide init() function gets called
 	_ "github.com/algorand/indexer/exporters/all"
 	_ "github.com/algorand/indexer/importers/all"
+	"github.com/algorand/indexer/loggers"
 	_ "github.com/algorand/indexer/processors/all"
 )
 
@@ -48,6 +45,7 @@ func init() {
 	logger.SetFormatter(&formatter)
 
 	conduitCmd.AddCommand(initCmd)
+	metrics.RegisterPrometheusMetrics()
 }
 
 // runConduitCmdWithConfig run the main logic with a supplied conduit config
--- a/conduit/pipeline.go
+++ b/conduit/pipeline.go
@@ -3,14 +3,20 @@ package conduit
 import (
 	"context"
 	"fmt"
-	"github.com/algorand/indexer/util/metrics"
-	log "github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
+	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"runtime/pprof"
 	"sync"
 	"time"
+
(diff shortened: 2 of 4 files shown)


conduit/pipeline.go
```diff
@@ -296,42 +310,52 @@ func (p *pipelineImpl) addMetrics(block data.BlockData, importTime time.Duration
...
 func (p *pipelineImpl) Start() {
 	p.wg.Add(1)
+	retry := 0
 	go func() {
 		defer p.wg.Done()
 		// We need to add a separate recover function here since it launches its own go-routine
 		defer HandlePanic(p.logger)
 		for {
 		pipelineRun:
+			metrics.PipelineRetryGauge.Set(float64(retry))
```
winder, line 320: Because of the way Prometheus collects metrics, I think this needs to be a counter. Since the metrics server scrapes metrics at a configurable interval, we could easily miss retry attempts unless we measure retries across all rounds.

winder: A histogram/summary might also work

cmd/conduit/main.go
```diff
@@ -7,18 +7,15 @@ import (
...
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	"github.com/algorand/indexer/conduit"
-	"github.com/algorand/indexer/loggers"
-)
-
-import (
-	// We need to import these so that the package wide init() function gets called
```
Eric-Warehime, line 19: Nit: I think this is a useful comment in relation to the `_` imports--explains why they're there.

util/metrics/metrics.go
```diff
@@ -88,4 +100,34 @@ var (
 			Name:      GetAlgodRawBlockTimeName,
 			Help:      "Total response time from Algod's raw block endpoint in seconds.",
 		})
+
+	ImporterTimeSeconds = prometheus.NewSummary(
+		prometheus.SummaryOpts{
+			Subsystem: "indexer_daemon",
```
Eric-Warehime, line 106: Are these specific to the indexer or will they also be available in any conduit pipeline?

shiqizng: they will be available to any conduit pipeline. I've thought about setting subsystem to "conduit" but thought it would be confusing to have both indexer_daemon and conduit subsystems showing up in the metrics.

conduit/pipeline.go
shiqizng: updated

cmd/conduit/main.go
shiqizng: my code editor keeps merging these import blocks into a single block. it seems that gofmt doesn't support using multiple import blocks.

shiqizng: updated.

conduit/pipeline.go
```diff
@@ -387,18 +412,21 @@ func (p *pipelineImpl) Start() {
...
 					if p.pipelineMetadata.NextRound > 0 {
-						p.addMetrics(blkData, importTime)
+						p.addMetrics(blkData, time.Since(start))
 					}
 					// Increment Round
 					p.setError(nil)
 					p.pipelineMetadata.NextRound++
 					p.encodeMetadataToFile()
+					retry = 0
+					p.pipelineMetadata.NextRound++
```
winder, line 429: Double increment here
```suggestion
					p.setError(nil)
					p.encodeMetadataToFile()
					retry = 0
					p.pipelineMetadata.NextRound++
```

winder (approved):

PR #1285, opened 2022-10-20
Conduit: Dynamic on complete

winder:
## Summary

It seemed like this function is useful to any plugin, so I have implemented it in a dynamic way.

Any plugin can now add an `OnComplete` callback if they need it, and do not need to provide it if they do not.

## Test Plan

Update tests.

### DIFF
--- a/conduit/pipeline.go
+++ b/conduit/pipeline.go
@@ -143,7 +143,7 @@ type pipelineImpl struct {
 	importer         *importers.Importer
 	processors       []*processors.Processor
 	exporter         *exporters.Exporter
-	completeCallback []Completed
+	completeCallback []OnCompleteFunc
 
 	round basics.Round
 }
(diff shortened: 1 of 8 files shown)


shiqizng (approved):

conduit/pipeline.go
```diff
@@ -140,10 +140,12 @@ type pipelineImpl struct {
...
 	initProvider *data.InitProvider
 
-	importer   *importers.Importer
-	processors []*processors.Processor
-	exporter   *exporters.Exporter
-	round      basics.Round
+	importer         *importers.Importer
+	processors       []*processors.Processor
+	exporter         *exporters.Exporter
+	completeCallback []Completed
```
winder, line 146: I'm not sure if this needs to be `[]*Completed` or not. I think all of our plugins use pointer receivers, so they are already pointers, but maybe that isn't always the case.

Eric-Warehime (approved):

Eric-Warehime:
Looks like the lint is failing, but otherwise really cool refactoring!

PR #1290, opened 2022-10-21
plugin metrics

shiqizng:
## Summary

adding some metrics to plugins.

### DIFF
--- a/processors/blockprocessor/block_processor.go
+++ b/processors/blockprocessor/block_processor.go
@@ -74,7 +74,7 @@ func (proc *blockProcessor) Init(ctx context.Context, initProvider data.InitProv
 	proc.logger = logger
 
 	// First get the configuration from the string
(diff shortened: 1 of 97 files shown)


processors/blockprocessor/block_processor.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/algorand/indexer/util/metrics"
```
winder, line 8: Can you register this without adding a common dependency?

Two possible options:
1) similar to `OnComplete(rnd) error`, a dynamic `ProvideMetrics() []prometheus.Collector` function could be used to fetch the metrics during init.
2) we could say the global metrics are always used and call `prometheus.Register` in the init function.

shiqizng: implemented option 1.

winder (approved):

PR #1293, opened 2022-10-26
e2e scripts readme

winder:
## Summary

Basic documentation for e2e testing scripts.

## Test Plan

N/A


e2e_tests/README.md
```diff
@@ -0,0 +1,65 @@
...
+Running e2elive
+
+**Note**: you will need to restart this container between tests, otherwise the database will already be initialized. Being out of sync with the data directory may cause unexpected behavior and failing tests.
+```sh
+e2elive --connection-string "host=127.0.0.1 port=5555 user=pguser password=pgpass dbname=mydb" --s3-source-net "fafa8862/rel-nightly" --indexer-bin ../cmd/algorand-indexer/algorand-indexer --indexer-port 9890
+```
+
+### e2econduit
+
+This runs a pipeline with conduit. The database is mocked so a real DB is not required.
```
Eric-Warehime, line 45: The db is not mocked, it's the same as the indexer e2e tests ( a postegresql container), it just manages that container's lifecycle for you instead of you having to spin it up either via docker-compose or manually.

winder: Cool, I missed that. So the only difference is that it doesn't call `e2equeries` at the end?

winder: updated

Eric-Warehime: > So the only difference is that it doesn't call e2equeries at the end? 

Yeah doesn't introspect into the db to make all of the queries that we do in e2elive, but it plays al of the same data against the indexer conduit pipeline and ensures we hit the expected round in the pipeline w/o error.

The hope (expectation) is that we'll extend the set of test cases though as we get new plugins. In that case `e2econduit` would run the entire set of test cases. Potentially it will be useful in the future to add a CLI command to be able to specify a single test scenario to run instead of all of them.

Eric-Warehime (approved):

PR #1295, opened 2022-10-27
Conduit: config updates

shiqizng:
## Summary
Add an option to the config file to write logs to a file instead of stdout and other config changes. 

## Test Plan
regression testing, run conduit daemon to validate expected behavior.

### DIFF
--- a/conduit/pipeline.go
+++ b/conduit/pipeline.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path"
@@ -108,7 +107,7 @@ func MakePipelineConfig(logger *log.Logger, cfg *Config) (*PipelineConfig, error
 
 	logger.Infof("Auto-loading Conduit Configuration: %s", autoloadParamConfigPath)
 
-	file, err := ioutil.ReadFile(autoloadParamConfigPath)
+	file, err := os.ReadFile(autoloadParamConfigPath)
 	if err != nil {
 		return nil, fmt.Errorf("MakePipelineConfig(): reading config error: %w", err)
 	}


cmd/conduit/main.go
```diff
@@ -78,10 +78,6 @@ func runConduitCmdWithConfig(cfg *conduit.Config) error {
 
 	logger.Info(cfg)
 
-	if err := cfg.Valid(); err != nil {
-		return err
-	}
-
```
shiqizng, line 84: removing this since it's also done in MakePipelineConfig.

conduit/pipeline.go
```diff
@@ -106,18 +108,11 @@ func MakePipelineConfig(logger *log.Logger, cfg *Config) (*PipelineConfig, error
 
 	logger.Infof("Auto-loading Conduit Configuration: %s", autoloadParamConfigPath)
 
-	file, err := os.Open(autoloadParamConfigPath)
-	if err != nil {
-		return nil, fmt.Errorf("MakePipelineConfig(): error opening file: %w", err)
-	}
-	defer file.Close()
-
-	err = viper.ReadConfig(file)
```
shiqizng, line 115: viper doesn't support yaml tags that are different from the field name. for example, viper doesn't pick up `log-file` field when unmarshalling the config file.

conduit/pipeline.go
```diff
@@ -507,6 +502,15 @@ func MakePipeline(ctx context.Context, cfg *PipelineConfig, logger *log.Logger)
...
 		return nil, fmt.Errorf("MakePipeline(): logger was empty")
 	}
+
+	if cfg.LogFile != "" {
+		f, err := os.OpenFile(cfg.LogFile, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
+		if err != nil {
+			return nil, fmt.Errorf("MakePipeline(): %w", err)
+		}
+		logger.SetOutput(f)
+	}
```
Eric-Warehime, line 512: We can probably add a really simple unit test that just ensures the expected log file gets created in the location we tell it to here.

conduit/pipeline.go
Eric-Warehime: I think @AlgoStephenAkiki is more familiar with how we're using viper. I'm wondering if this breaks something that we were depending on--either with how you can set config values via environment variables or I recall discussing whether we accept `.yml` and `.yaml` extensions?

AlgoStephenAkiki: Re: viper vs yaml - I think it just was a matter of preference from Will to consistently use viper instead of yaml.  We might need to get his opinion on that to go forward.  For the .yml/.yaml thing, I just had a list of extensions to check that I iterated through.

conduit/pipeline.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
```
winder, line 8: This package is deprectated

conduit/pipeline.go
winder: I think I see part of the problem. The global viper config is used in `cmd/conduit/main.go`, I bet there are some weird side when we use it again here.

Earlier I was imagining we could use these features to simplify the pipeline init:
https://github.com/spf13/viper#extracting-a-sub-tree
https://github.com/spf13/viper#establishing-defaults

Since we're not using those, maybe we should remove viper completely and re-introduce it later if it ever makes sense.

shiqizng: it would fail to load CPUProfile if the config file used `cpu-profile` instead of CPUprofile. it's the same problem for the other fields with kebab case tags, like PIDFilePath or PipelineLogLevel.

conduit/pipeline.go
shiqizng: removed this package.

conduit/pipeline.go
shiqizng: added a test.

winder (approved):

PR #1296, opened 2022-10-28
List subcommand

### LINKED ISSUE #1270: Conduit: list subcommand which prints available plugins
A subcommand to list all of the plugins, grouped by type.

*list* - use the metadata object to get the description.

If a plugin is marked as deprecated, make sure the output makes that clear.

```bash
~$ ./conduit list
Importers:
    algod            Importer for fetching blocks from an algod REST API.
    
Processors:
    block_processor ...
    filter          ...

Exporters:
    file_writer ..
    postgresql  ..
```

*list <plugin>" - details. Not exactly sure how this works, maybe we should add a sample config in each plugin directory and include that as part of the metadata.
```
~$ ./conduit list algod
    Name: algod
    Config:
        - netaddr: "Algod netaddr string"
          token: "...."
```

AlgoStephenAkiki:
Resolves #1270

Adds the list subcommand with error checking and output

## Summary

## Test Plan


AlgoStephenAkiki:
I've only implemented sample configs on the processors to get feedback if this is an ok way to structure it.  Will add more configs once I get the green light.

cmd/conduit/list.go
```diff
@@ -0,0 +1,214 @@
...
+			// search exporters
+			ctor, err := exporters.ExporterBuilderByName(strippedName)
+			if err != nil {
+				return err
+			}
+
+			meta := ctor.New().Metadata()
+			fmt.Fprint(out, meta.SampleConfig())
+			return nil
+		}
```
Eric-Warehime, line 86: We should combine all of these if possible. Maybe something like this?
```go
func matchOrSuggest(pluginName string, metadataSearch func(string) (*plugins.PluginMetadata, error)) (string, error) 
```

Eric-Warehime (commented):
I like how you've used sample configs embedded in the metadata for this! 

I think either readme docs or maybe adding a help subcommand (like `conduit list help`) would help me get an idea of the behavior expected from the CLI and whether or not everything here conforms with those expectations. As examples, the difference between listing `conduit list importer.pluginName` versus just `conduit list pluginName` versus `conduit list`, should you be able to run `conduit list importers` for all importers, should there be different levels of output for something like `conduit list importers.algod [--verbose]`, etc.

processors/blockprocessor/sample.yaml
```diff
@@ -0,0 +1,2 @@
+Name: block_evaluator
+Config:
```
winder, line 2: I like this approach, it gives us a lot of flexibility to include inline documentation:

```suggestion
# Execute block transactions to compute the new account states.
name: block_evaluator
config:
    # each block evaluator field and its comment.
```

We should add a unit test for each of these and make sure:
* every field is set (i.e. make sure the sample is updated when a new field is added.
* the sample is definitely valid.

Maybe there's a way to use reflection to make a generic function that accepts the string and the concrete type?

importers/metadata.go
```diff
@@ -4,9 +4,10 @@ import "github.com/algorand/indexer/plugins"
 
 // ImporterMetadata returns fields relevant to identification and description of plugins.
 type ImporterMetadata struct {
```
winder, line 6: Thinking out loud. This might not be the right PR.

Maybe `PluginMetadata` should be a struct instead of an interface. Do we get anything from having an implementation of that interface in each plugin type? If they all shared the same type then you would only need to update one place and we could extend it without breaking the interface.

cmd/conduit/list.go
```diff
@@ -0,0 +1,214 @@
...
+
+		fmt.Fprint(out, output)
+	}
+
+	var potentialNames []string
+	var sampleConfig string
+
+	for _, pluginName := range pluginNames {
+
+		if strings.HasPrefix(pluginName, "importer.") {
```
winder, line 49: It might be possible to embed a separate `Metadata()` interface in each of the plugin types and use a helper function to print the list of each plugin type.

cmd/conduit/list.go
```diff
@@ -0,0 +1,214 @@
...
+			if err != nil {
+				return err
+			}
+
+			meta := ctor.New().Metadata()
+			fmt.Fprint(out, meta.SampleConfig())
+			return nil
+		}
+
+		if strings.HasPrefix(pluginName, "exporter.") {
```
shiqizng, line 75: these can probably be if-else block and repeated code can be reduced.

AlgoStephenAkiki:
Changed around the workflow according to slack convos and comments, please re-review.

processors/blockprocessor/sample.yaml
AlgoStephenAkiki: I can make sure that the sample is valid by initializing a plugin and using `init()` to consume and check for errors.  I'm not sure how to do the other part where we check that every field is filled out.  I'll look at generics/reflection to see what we can do.

importers/metadata.go
AlgoStephenAkiki: I think it makes sense.  But perhaps another issue like you said.

cmd/conduit/list.go
```diff
@@ -0,0 +1,178 @@
...
+		return "", err
+	}
+	importerNames := importers.ImporterNames()
+	for _, importerName := range importerNames {
+		ctor, err := importers.ImporterBuilderByName(importerName)
+		// belt and suspenders
+		if err != nil {
+			return "", err
+		}
+		meta := ctor.New().Metadata()
```
shiqizng, line 109: maybe we could also store the metadata when register the constructor and have a helper like GetMetadataByName(name string)

cmd/conduit/list.go
```diff
@@ -0,0 +1,178 @@
...
+		fmtString := "%s - %s\n"
+
+		if meta.Deprecated() {
+			fmtString = "[DEPRECATED] " + fmtString
+		}
+
+		_, err = fmt.Fprintf(&sb, "  "+fmtString, processorName, meta.Description())
+		if err != nil {
+			return "", err
+		}
```
shiqizng, line 146: this block of code is repeated. it can probably be wrapped in a method.

AlgoStephenAkiki: Done

cmd/conduit/list.go
AlgoStephenAkiki: Done

processors/processor_factory.go
```diff
@@ -24,14 +25,23 @@ func (f ProcessorConstructorFunc) New() Processor {
...
+
 // RegisterProcessor is used to register ProcessorConstructor implementations. This mechanism allows
 // for loose coupling between the configuration and the implementation. It is extremely similar to the way sql.DB
 // driver's are configured and used.
 func RegisterProcessor(name string, constructor ProcessorConstructor) error {
 	if _, ok := processorImpls[name]; ok {
 		return fmt.Errorf("processor already exists")
 	}
+
+	if _, ok := processorMetaData[name]; ok {
```
shiqizng, line 39: I don't think this can happen. if name already exists, it should've exited in line 36?

shiqizng (approved):

processors/processor_factory.go
AlgoStephenAkiki: It's more of a belt and suspenders thing

cmd/conduit/list.go
```diff
@@ -0,0 +1,173 @@
...
+		Short: "lists all plugins available to conduit",
+		Long:  "lists all plugins available to conduit",
+		Args:  cobra.MaximumNArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runListCmd(os.Stdout, args)
+		},
+		SilenceUsage: true,
+		// Silence errors because our logger will catch and print any errors
+		SilenceErrors: true,
+	}
```
winder, line 29: We seem to be doing something wrong with the SilenceErrors / SilenceUsage / RunE configuration.

This command is interactive, output should be human readable. And we should print out the usage information when there's an error to help people use the command:

```
./conduit list importers
{"__type":"Conduit","_name":"main","level":"error","msg":"invalid argument, specify [importers, processors, exporters] before plugin name","time":"2022-11-03T10:22:58-04:00"}

./conduit list importers file_writer
{"__type":"Conduit","_name":"main","level":"error","msg":"no importer by name: file_writer","time":"2022-11-03T10:23:10-04:00"}


./conduit list exporter file_writer
{"__type":"Conduit","_name":"main","level":"error","msg":"invalid argument type (exporter), specify [importers, processors, exporters] before plugin name","time":"2022-11-03T10:23:15-04:00"}
```

processors/blockprocessor/sample_test.go
```diff
@@ -0,0 +1 @@
+package blockprocessor
```
Eric-Warehime, line 1: What's the purpose of this file?

cmd/conduit/list.go
```diff
@@ -0,0 +1,173 @@
...
+
+		fmt.Fprint(out, output)
+		return nil
+	}
+
+	if len(listArgs) == 1 {
+		return fmt.Errorf("invalid argument, specify [importers, processors, exporters] before plugin name")
+	}
+
+	pluginTypeSpecifier := listArgs[0]
```
Eric-Warehime, line 49: Can we downcase and remove space from this so that `conduit list  Importers   foobar` still works as expected. It's possible that cobra is already removing spaces.

processors/blockprocessor/sample_test.go
winder: Lets drop the per-plugin sample tests. [This general test gets us most of the way there](https://github.com/algorand/indexer/pull/1315/files#diff-03dff77f7636e05b34cf95cbb0ecd02024030d89eb953e260c937418c05ee69c).

Eric-Warehime (approved):

winder:
There's a test failing to cleanup after itself causing the tests to fail. I'll fix it in a followup.

PR #1297, opened 2022-10-31
Added none operation

### LINKED ISSUE #1283: Conduit: filter plugin 'none' operation
In addition to `any` and `all`, `none` would be useful and probably straightforward to add.

AlgoStephenAkiki:
Resolves #1283

Adds none operation to fields

## Summary

## Test Plan

### DIFF
--- a/processors/filterprocessor/fields/filter.go
+++ b/processors/filterprocessor/fields/filter.go
@@ -12,10 +12,11 @@ type Operation string
 
 const anyFieldOperation Operation = "any"
 const allFieldOperation Operation = "all"
+const noneFieldOperation Operation = "none"
 
 // ValidFieldOperation returns true if the input is a valid operation
 func ValidFieldOperation(input string) bool {
-	if input != string(anyFieldOperation) && input != string(allFieldOperation) {
+	if input != string(anyFieldOperation) && input != string(allFieldOperation) && input != string(noneFieldOperation) {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


processors/filterprocessor/fields/filter.go
```diff
@@ -33,6 +34,28 @@ func (f Filter) SearchAndFilter(input data.BlockData) (data.BlockData, error) {
...
 	switch f.Op {
+	case noneFieldOperation:
+		for _, txn := range input.Payset {
+
+			allFalse := true
+			for _, fs := range f.Searchers {
+				b, err := fs.search(txn)
+				if err != nil {
+					return data.BlockData{}, err
+				}
```
Eric-Warehime, line 45: This looks like the only bit that isn't covered by the unit tests. What are the failure conditions which would cause the searcher to return an error? Is this conditional something that should never happen or it could happen if you provide an invalid pattern?

winder (approved):

processors/filterprocessor/filter_processor_test.go
```diff
@@ -15,6 +15,141 @@ import (
...
+					},
+				},
+			},
+		},
+	)
+
+	output, err := fp.Process(bd)
+	assert.NoError(t, err)
+	assert.Equal(t, len(output.Payset), 1)
+	assert.Equal(t, output.Payset[0].SignedTxnWithAD.SignedTxn.Txn.PaymentTxnFields.Receiver, sampleAddr2)
```
shiqizng, line 146: nit: maybe assign the txn to a var and then assert equal on output.Payset[0] and that txn.

shiqizng (approved):

processors/filterprocessor/fields/filter.go
AlgoStephenAkiki: It just allows the searcher interface to propagate errors if the implementation needs to.  A quick example is the microalgos searcher which can return an error if the field is not a microalgo type (sort of belt and suspenders, but I think it is useful).

PR #1311, opened 2022-11-02
conduit: next-round command line flag

shiqizng:
## Summary

Add a command line flag to set the "next round" instead of using the round specified in the metadata.json file.

## Test Plan
add new unit tests


conduit/pipeline.go
```diff
@@ -262,6 +262,14 @@ func (p *pipelineImpl) Init() error {
...
 		return fmt.Errorf("Pipeline.Start(): genesis hash in metadata does not match expected value: actual %s, expected %s", gh, p.pipelineMetadata.GenesisHash)
 	}
+	// overwriting NextRound if next-round arg is set
+	if p.cfg.ConduitConfig.Flags != nil && p.cfg.ConduitConfig.Flags.Changed("next-round") {
+		if round, _ := p.cfg.ConduitConfig.Flags.GetUint64("next-round"); err == nil && round > p.pipelineMetadata.NextRound {
+			p.pipelineMetadata.NextRound = round
+		} else if round < p.pipelineMetadata.NextRound {
+			return fmt.Errorf("Pipeline.Start(): cannot set starting round to %d, it is behind next round %d", round, p.pipelineMetadata.NextRound)
+		}
+	}
```
winder, line 272: A few things here.

* I don't think we should be using flags at this level of the program. @Eric-Warehime could you weigh in on why we have flags in the config?
* Please add a log `"Overriding default next round from %d to %d."`
* It's not an error to set the metadata to an earlier round. Allow the plugins to decide if thats a supported case (i.e. the block processor plugin, and postgres writers can generate an error, the rest are more stateless and shouldn't care)

shiqizng: do we want to remove Flags from ConduitConfig?

winder: For this PR, just add a field for NextRoundOverride. I think there's a bigger question about viper/yaml/pflags/env-vars to consider as a whole.

winder: Although... it's nice that you're able to distinguish between 0 and unset with this logic...

Eric-Warehime: It seems like the Flags are set in the cfg so we can later merge values set via the CLI with values set via environment variables. https://github.com/algorand/indexer/blob/conduit/cmd/conduit/main.go#L75

We can get rid of it, but it would mean that any variables set via the environment would override those set via the CLI--which is not the current behavior.

winder (approved):

Eric-Warehime (approved):

PR #1324, opened 2022-11-09
Ignore no op asset transfers.

winder:
## Summary

If an asset transfer moves no assets, and is not an optin or closeout transaction, do not prefetch its resources.

## Test Plan

CI and manual testing.

### DIFF
--- a/accounting/eval_preload.go
+++ b/accounting/eval_preload.go
@@ -27,7 +27,7 @@ func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD, assetsReq map[ba
 			!fields.AssetCloseTo.IsZero() && // closeout
 			!fields.AssetSender.IsZero() // clawback
 
-		if fields.XferAsset == 0 && !noOpXfer {
+		if fields.XferAsset != 0 && !noOpXfer {
 			assetsReq[fields.XferAsset] = struct{}{}
 		}
 	case protocol.AssetFreezeTx:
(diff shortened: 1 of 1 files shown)


michaeldiamant (approved):
@winder PR matches my understanding for a tactical fix.  Given that a more principled change is already underway, a quick fix seems _reasonable_ to me.

Prefer 1+ approvals from folks more familiar with go-algorand ledger to vouch that the change is _sensible_.

accounting/eval_preload.go
```diff
@@ -22,7 +22,12 @@ func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD, assetsReq map[ba
 		}
 	case protocol.AssetTransferTx:
 		fields := &txn.AssetTransferTxnFields
-		if fields.XferAsset != 0 {
+
+		noOpXfer := (fields.AssetAmount == 0) &&
```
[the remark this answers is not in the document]
jannotti: Right - since we certainly grabbed the asset to check the clawback addr, even if the movement is 0 amount (I check, as I suppose we don't _have_ to check a 0 clawback)

accounting/eval_preload.go
```diff
@@ -22,7 +22,12 @@ func addToCreatorsRequest(stxnad *transactions.SignedTxnWithAD, assetsReq map[ba
...
+			(txn.Sender != fields.AssetReceiver) &&
+			(fields.AssetCloseTo != basics.Address{})
+
+		if fields.XferAsset == 0 && !noOpXfer {
```
jannotti, line 30: isn't it `fields.XferAsset != 0` ?

winder: yup, thanks

jannotti (approved):
Seems good

algorandskiy (approved):
Do we need to add a test?

Issue #1320, opened 2022-11-08
refactor idb module


shiqizng:
Replace go-algorand dependencies with sdk in idb/postgres and document types that can't be replaced or are not available in the sdk. The focus of this one is to replace "go-algorand/data/transactions"

Acceptance Criteria 
- regression testing passes

shiqizng:
[PR](https://github.com/algorand/indexer/pull/1334) merged

PR #1325, opened 2022-11-09
[WIP] enhancement: remove go-algorand imports

shiqizng:
the main PR for #1320. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203413680067756


conduit/pipeline/pipeline_test.go
```diff
@@ -178,15 +178,15 @@ func (m *mockImporter) OnComplete(input data.BlockData) error {
 	if m.onCompleteError {
 		err = fmt.Errorf("on complete")
 	}
-	m.finalRound = input.BlockHeader.Round
```
Eric-Warehime, line 181: I'm fine with committing things like this for now as incremental improvements. But eventually the `data.BlockData` input won't be using go-algorand types and so doing `input.BlockHeader.Round` would just return a `types.Round` anyways.

idb/postgres/postgres.go
```diff
@@ -204,7 +204,7 @@ func (db *IndexerDb) AddBlock(vb *ledgercore.ValidatedBlock) error {
 		}
 		defer w.Close()
 
-		if block.Round() == basics.Round(0) {
+		if uint64(block.Round()) == uint64(0) {
```
Eric-Warehime, line 207: Same comment as above--the block object we're using will eventually return the proper type. Except here we can't even get rid of the `basics` import so I'm not sure how much benefit this gives us.

PR #1353, opened 2022-11-30
metrics: Configurable conduit metrics prefix.

winder:
## Summary

Add a configuration to the metrics prefix so that conduit can be different than Indexer.

### DIFF
--- a/util/metrics/metrics.go
+++ b/util/metrics/metrics.go
@@ -2,9 +2,15 @@ package metrics
 
 import "github.com/prometheus/client_golang/prometheus"
 
+// This is helpful for tests to ensure there are never uninitialized values.
+func init() {
+	RegisterPrometheusMetrics("uninitialized")
+}
+
 // RegisterPrometheusMetrics register all prometheus metrics with the global
 // metrics handler.
 func RegisterPrometheusMetrics(subsystem string) {
+	deregister()
 	instantiageCollectors(subsystem)
 
 	_ = prometheus.Register(GetAlgodRawBlockTimeSeconds)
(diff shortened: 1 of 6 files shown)


util/metrics/metrics.go
```diff
@@ -2,59 +2,71 @@ package metrics
...
+}
+
 // RegisterPrometheusMetrics register all prometheus metrics with the global
 // metrics handler.
-func RegisterPrometheusMetrics() {
-	for _, c := range collectors {
-		prometheus.Register(c)
+func RegisterPrometheusMetrics(subsystem string) {
+	deregister()
+	instantiageCollectors(subsystem)
```
shiqizng, line 14: typo?

util/metrics/metrics.go
```diff
@@ -2,59 +2,71 @@ package metrics
...
+	RegisterPrometheusMetrics("uninitialized")
+}
+
 // RegisterPrometheusMetrics register all prometheus metrics with the global
 // metrics handler.
-func RegisterPrometheusMetrics() {
-	for _, c := range collectors {
-		prometheus.Register(c)
+func RegisterPrometheusMetrics(subsystem string) {
+	deregister()
```
shiqizng, line 13: why is deregister() needed?

winder: It's mainly for tests, which need to make sure these objects are instantiated even if the main init functions are not called. I'll update with a comment.
```suggestion
	// deregister metric objects in case the register function was called more than once.
	// This helps with testing.
	deregister()
```

Eric-Warehime (approved):
LGTM after you fix typo and resolve merge conflicts.

PR #1357, opened 2022-11-30
conduit: Pass data dir to plugin config.

winder:
## Summary

Pass a data directory to plugins so that path arguments can be optional.

## Test Plan

New unit tests.


mciccarello (approved):
Looks good to me. I don't see anything that directly conflicts with what I've done with the plugin config strings. However be aware I've changed how they are indented slightly to make it work with conduit init, and I am dependent on that particular format (which I'm not super happy about, but will take time to generalize)

PR #1360, opened 2022-12-01
Conduit: init parameters added to conduit init command

mciccarello:
## Summary

Added parameters (actually Cobra calls them flags) for user to set importer, exporter and processors and have them show up in the same yaml configuration file.

## Test Plan

Modifications to unit test to come (reason for the draft status)


cmd/conduit/main.go
```diff
@@ -155,7 +162,37 @@ func runConduitInit(path string) error {
...
+		exporter = defaultExporter
+	} else {
+		exporter = twoSpaces + "name: " + exporterFlag + "\n" + twoSpaces + "config:"
+	}
+
+	var processors string
+	for _, processor := range processorList {
+		processors = processors + twoSpaces + "- "
+		processors = processors + "name: " + processor + "\n" + twoSpaces + twoSpaces + "config:\n"
+	}
```
winder, line 190: could you add:
* validation - make sure the requested plugin exists.
* use the metadata template to fill in a sample config.

There are some helpers to fetch the metadata:
https://github.com/algorand/indexer/blob/develop/conduit/pipeline/metadata.go

The `./conduit list ...` command shows this data being used:
https://github.com/algorand/indexer/blob/develop/cmd/conduit/internal/list/list.go

cmd/conduit/main.go
```diff
@@ -14,6 +15,7 @@ import (
 	"github.com/algorand/indexer/cmd/conduit/internal/list"
 	"github.com/algorand/indexer/conduit"
 	"github.com/algorand/indexer/conduit/pipeline"
+
```
winder, line 18: the lint error is related to the import order in this file. It is fixed on develop.

mciccarello: Thanks, the error message was so helpful.

mciccarello: Actually I merged develop and it doesn't change anything about the import order as far as I can tell, and the problem persists.

PR #1363, opened 2022-12-02
New Feature: Removes logic dependency

### LINKED ISSUE #1328: Remove go-algorand/data/transactions/logic dependency.
### Subject of the issue

As part of boxes support a dependency to `github.com/algorand/go-algorand/data/transactions/logic` was added. We are currently in the process of removing go-algorand dependencies, so it is time to decide what to do with this package.

Previous conversation regarding this dependency:
https://github.com/algorand/indexer/pull/1117#discussion_r961975811

AlgoStephenAkiki:
Resolves #1328

Removes the logic dependency from the indexer with the exception of AssembleStringWithVersion in account_testutil.go

## Summary

## Test Plan


util/test/account_testutil.go
```diff
@@ -216,6 +216,7 @@ func MakeCreateAppTxn(sender basics.Address) transactions.SignedTxnWithAD {
 // MakeComplexCreateAppTxn makes a transaction that creates an arbitrary app. When assemblerVersion is set to 0, use the AssemblerDefaultVersion.
 func MakeComplexCreateAppTxn(sender basics.Address, approval, clear string, assemblerVersion uint64) (transactions.SignedTxnWithAD, error) {
 	// Create a transaction with ExtraProgramPages field set to 1
+	// TODO how to handle this?
```
AlgoStephenAkiki, line 219: @shiqizng @winder This function complies teal programs from strings.  Is this a function that needs to be refactored or replaced completely?  It currently is only used in testing in Indexer

winder (changes requested):
AFAIK we need the following functions, not the entire logic package:

```
// MakeBoxKey creates the key that a box named `name` under app `appIdx` should use.
func MakeBoxKey(appIdx basics.AppIndex, name string) string {
	/* This format is chosen so that a simple indexing scheme on the key would
	   allow for quick lookups of all the boxes of a certain app, or even all
	   the boxes of a certain app with a certain prefix.

	   The "bx:" prefix is so that the kvstore might be usable for things
	   besides boxes.
	*/
	key := make([]byte, boxNameIndex+len(name))
	copy(key, boxPrefix)
	binary.BigEndian.PutUint64(key[boxPrefixLength:], uint64(appIdx))
	copy(key[boxNameIndex:], name)
	return string(key)
}

// SplitBoxKey extracts an appid and box name from a string that was created by MakeBoxKey()
func SplitBoxKey(key string) (basics.AppIndex, string, error) {
	if len(key) < boxNameIndex {
		return 0, "", fmt.Errorf("SplitBoxKey() cannot extract AppIndex as key (%s) too short (length=%d)", key, len(key))
	}
	if key[:boxPrefixLength] != boxPrefix {
		return 0, "", fmt.Errorf("SplitBoxKey() illegal app box prefix in key (%s). Expected prefix '%s'", key, boxPrefix)
	}
	keyBytes := []byte(key)
	app := basics.AppIndex(binary.BigEndian.Uint64(keyBytes[boxPrefixLength:boxNameIndex]))
	return app, key[boxNameIndex:], nil
}

// AppCallBytes represents an encoding and a value of an app call argument.
type AppCallBytes struct {
	Encoding string `codec:"encoding"`
	Value    string `codec:"value"`
}

// NewAppCallBytes parses an argument of the form "encoding:value" to AppCallBytes.
func NewAppCallBytes(arg string) (AppCallBytes, error) {
	parts := strings.SplitN(arg, ":", 2)
	if len(parts) != 2 {
		return AppCallBytes{}, fmt.Errorf("all arguments and box names should be of the form 'encoding:value'")
	}
	return AppCallBytes{
		Encoding: parts[0],
		Value:    parts[1],
	}, nil
}
```

AlgoStephenAkiki:
> AFAIK we need the following functions, not the entire logic package:
> 
> ```
> // MakeBoxKey creates the key that a box named `name` under app `appIdx` should use.
> func MakeBoxKey(appIdx basics.AppIndex, name string) string {
> 	/* This format is chosen so that a simple indexing scheme on the key would
> 	   allow for quick lookups of all the boxes of a certain app, or even all
> 	   the boxes of a certain app with a certain prefix.
> 
> 	   The "bx:" prefix is so that the kvstore might be usable for things
> 	   besides boxes.
> 	*/
> 	key := make([]byte, boxNameIndex+len(name))
> 	copy(key, boxPrefix)
> 	binary.BigEndian.PutUint64(key[boxPrefixLength:], uint64(appIdx))
> 	copy(key[boxNameIndex:], name)
> 	return string(key)
> }
> 
> // SplitBoxKey extracts an appid and box name from a string that was created by MakeBoxKey()
> func SplitBoxKey(key string) (basics.AppIndex, string, error) {
> 	if len(key) < boxNameIndex {
> 		return 0, "", fmt.Errorf("SplitBoxKey() cannot extract AppIndex as key (%s) too short (length=%d)", key, len(key))
> 	}
> 	if key[:boxPrefixLength] != boxPrefix {
> 		return 0, "", fmt.Errorf("SplitBoxKey() illegal app box prefix in key (%s). Expected prefix '%s'", key, boxPrefix)
> 	}
> 	keyBytes := []byte(key)
> 	app := basics.AppIndex(binary.BigEndian.Uint64(keyBytes[boxPrefixLength:boxNameIndex]))
> 	return app, key[boxNameIndex:], nil
> }
> 
> // AppCallBytes represents an encoding and a value of an app call argument.
> type AppCallBytes struct {
> 	Encoding string `codec:"encoding"`
> 	Value    string `codec:"value"`
> }
> 
> // NewAppCallBytes parses an argument of the form "encoding:value" to AppCallBytes.
> func NewAppCallBytes(arg string) (AppCallBytes, error) {
> 	parts := strings.SplitN(arg, ":", 2)
> 	if len(parts) != 2 {
> 		return AppCallBytes{}, fmt.Errorf("all arguments and box names should be of the form 'encoding:value'")
> 	}
> 	return AppCallBytes{
> 		Encoding: parts[0],
> 		Value:    parts[1],
> 	}, nil
> }
> ```

The AppCallBytes struct requires the `Raw()` method in handlers `LookupApplicationBoxByIDAndName` and `SearchForApplicationBoxes`.  If we don't pull the struct over then we still maintain a link to the original package which is why I've pulled over as much of the logic package as I have.

I also pulled out the basics package to try and remove more dependencies with these functions but we can curtail that if you want.

Eric-Warehime (commented):
We do not need any of the `abi/` code redefined since this package was already pulled into a separate repo which should be usable as a dependency. 

https://github.com/algorand/avm-abi

util/test/account_testutil.go
shiqizng: I think this method needs to use the new logic package which should include methods like `AssembleStringWithVersion` and `AssembleString`

shiqizng: it's not ideal to change this method to take raw teal bytes since this would require users to first compile their teal program using something like goal then copy the bytes to the tests.

api/app_boxes_fixtures_test.go
```diff
@@ -96,7 +97,7 @@ func setupLiveBoxes(t *testing.T, proc func(cert *rpcs.EncodedBlockCert) error,
 	newBoxValue := "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
 	boxTxns := make([]*transactions.SignedTxnWithAD, 0)
 	for _, boxName := range boxNames {
-		expectedAppBoxes[firstAppid][logic.MakeBoxKey(firstAppid, boxName)] = newBoxValue
+		expectedAppBoxes[firstAppid][logic.MakeBoxKey(indexerBasics.AppIndex(firstAppid), boxName)] = newBoxValue
```
shiqizng, line 100: sdk has AppIndex. reuse types from sdk whenever possible.

util/test/account_testutil.go
winder: Wouldn't that would require copying the entire teal assembler into this program?

I'm generally expecting that all of these `Make*Txn` functions need to be replaced with `Make*StateDelta` (or `Add*ToStateDelta`) to accommodate the new interfaces. So this may end up being a non-issue with some of the followup work.

winder (commented):
Updates look good so far :+1:

go.mod
```diff
@@ -29,6 +29,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
```
michaeldiamant, line 31: @AlgoStephenAkiki + other reviewers:  For my understanding - What's the plan to remove remaining dependencies so that go-algorand `go.mod` entry can be removed?

* I understand the PR description marks AVM assembler as out-of-scope.  However, I can imagine some solutions that change based on considering the assembler.
* I'm gauging if paths have been considered, if it's an open question or something else.

michaeldiamant: Resolving - Thanks to discussion with @winder and @AlgoStephenAkiki, I learned the plan is:  In a forthcoming PR, remove the need for assembler usage by switching to state delta API.

PR #1359, opened 2022-11-30
merge: 2.15.0 release changes

Eric-Warehime:
## Summary

Merges changes committed during the 2.15.0 release. Includes:
#1341 
#1324

And some final changes to fix golangci-lint failures.

## Test Plan

CI

PR #1364, opened 2022-12-04
Bug-Fix: pass failing `nightly_test_indexer_vs_algod`

tzaffi:
## Summary

The `nightly_test_indexer_vs_algod` started failing recently because of the following chain of coincidences:
1. Introduction of `golanci-lint` in #1359 
2. The installed `golangci-lint` requires go $\geq$ 1.17
3. `nightly_test_indexer_vs_algod` runs on the default Circle golang orb version (1.16)

## Solution

Break `install_linter` out of `install_dependencies` and have all the jobs run `install_linter` after `install_dependencies` except for `nightly_test_indexer_vs_algod` which is doesn't need linting, and therefore doesn't need to upgrade its golang.

## Test Plan

Once this is merged into develop, the nightly test will confirm🤞 that the fix is correct. In the meanwhile, this [commit](https://github.com/algorand/indexer/commit/c9d6eb707859b9cdcab2440970ff0743c97c6d48) (the diff is confusing because of the previous state of the repo) introduced a temporary test that simulates the `indexer_vs_algod_nightly_TESTING123TESTING` and we can see that [the test passed on circle](https://app.circleci.com/pipelines/github/algorand/indexer/2965/workflows/f243f829-bea3-4baf-96db-76a9e8beca79/jobs/4337). Here's a screenshot of the temporarily introduced test:

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,10 +119,15 @@ jobs:
       go_version:
         type: string
     steps:
+      - restore_cache:
+        keys:
       - go/install:
         version: << parameters.go_version >>
       - install_dependencies
       - run_indexer_vs_algod
+      - save_cache:
+        key: just-kidding
+        paths:
 
 commands:
   install_dependencies:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,7 +113,6 @@ jobs:
               ]
             }
   indexer_vs_algod_nightly_BOO:   # TEMPorarily adding for test purpose
-    no_cache: true
     machine:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  go: circleci/go@1.7.0
+  go: circleci/go@1.7.1
```
tzaffi, line 4:
```suggestion
  go: circleci/go@1.7.0
```

PR #1369, opened 2022-12-07
Conduit: Simplify and fix Conduit init.

winder:
## Summary

While adding a startup banner, I noticed that the config file was not being updated as expected. Additionally the log level was not being applied to the root logger. While debugging I attempted to simplify things until I could get everything working as expected.

## Test Plan

Existing unit tests.


conduit/pipeline/pipeline.go
```diff
@@ -67,64 +61,60 @@ type Config struct {
...
-	paramConfigFound := err == nil
-
-	if !paramConfigFound {
-		return nil, fmt.Errorf("MakePipelineConfig(): could not find %s in data directory (%s)", conduit.DefaultConfigName, cfg.ConduitDataDir)
+	autoloadParamConfigPath, err := util.GetConfigFromDataDir(args.ConduitDataDir, conduit.DefaultConfigBaseName, []string{"yml", "yaml"})
+	if err != nil || autoloadParamConfigPath == "" {
+		return nil, fmt.Errorf("MakePipelineConfig(): could not find %s in data directory (%s)", conduit.DefaultConfigName, args.ConduitDataDir)
 	}
 
-	logger.Infof("Auto-loading Conduit Configuration: %s", autoloadParamConfigPath)
```
AlgoStephenAkiki, line 111: Would this be useful to keep for debugging purposes?

AlgoStephenAkiki (approved):

conduit/pipeline/pipeline.go
winder: I think so, but the logger isn't initialized at this point (we don't know the log level until after processing the config file)

PR #1372, opened 2022-12-08
Add Conduit startup banner and fix default log level.

winder:
## Summary

Add a startup banner to Conduit. It prints to stdout and indicates that we're currently in a pre-release mode.

Optional configuration value to disable the banner. Indexer's automatically configured conduit disables the banner because that configuration has been tested much more thoroughly than other configurations.

Also fix the log-level setting.

## Test Plan

New unit test.


loggers/loggers_test.go
```diff
@@ -3,25 +3,42 @@ package loggers
...
 
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type FakeIoWriter struct {
-	// A list of strings that represent ints
-	ListOfInts []string
+	Entries []string
```
winder, line 18: I'm not sure why this was named `ListOfInts`, so I renamed it.

shiqizng (approved):
since the banner is default to hidden, the release will also have a pre-release tag, right?

Eric-Warehime (approved):

AlgoStephenAkiki (approved):

winder:
@shiqizng yes. It should only defaults to hidden when running via Indexer, I was it would be OK because that configuration has had much more testing.

PR #1377, opened 2022-12-09
conduit: Add version flag to conduit root command.

winder:
## Summary

Simple version flag that hooks into the same system Indexer uses.

## Test Plan

Manually.


mciccarello (approved):
Looks ok.

Issue #1347, opened 2022-11-28
License - final pass


shiqizng:
Do a final pass on indexer repo 

- [x] all go-algorand references are removed. 
- [x] remove temporary type conversion helpers(e.g. ConvertParams)
- [x] tidy account_testutil.go. remove methods that create go-algorand types, rename "V2" methods. 
- [x] remove go-algorand from the mod file
- [x] delete third_party/go-algorand submodule, remove references from Makefile
- [x] delete fetcher and block_processor packages, and the `GetAlgodRawBlockTimeSeconds` metric.

These should all be simple tasks, if we missed anything challenging create a separate issue.

PR #1378, opened 2022-12-12
Conduit: use correct subsystem in plugin metrics.

winder:
## Summary

Make sure plugin metrics are able to use the same subsystem as the rest of Conduit.

## Test Plan

New unit tests.


conduit/plugins/importers/algod/metrics.go
```diff
@@ -2,17 +2,5 @@ package algodimporter
...
-var (
-	GetAlgodRawBlockTimeSeconds = prometheus.NewSummary(
-		prometheus.SummaryOpts{
-			Subsystem: "indexer_daemon",
-			Name:      GetAlgodRawBlockTimeName,
-			Help:      "Total response time from Algod's raw block endpoint in seconds.",
-		})
-)
+// getAlgodRawBlockTimeSeconds is used to record how long it took to fetch the block.
+var getAlgodRawBlockTimeSeconds prometheus.Summary
```
shiqizng, line 6: it looks like GetAlgodRawBlockTimeSeconds can be removed from [util/metrics/metrics.go](https://github.com/algorand/indexer/blob/develop/util/metrics/metrics.go#L18)

shiqizng (approved):

conduit/plugins/importers/algod/metrics.go
winder: It's used by fetcher, but maybe fetcher is completely deprecated now?

winder: Added a note to remove this metric when fetcher is removed: https://github.com/algorand/indexer/issues/1347

AlgoStephenAkiki (approved):

PR #1382, opened 2022-12-14
docs: Docs automation

Eric-Warehime:
## Summary

Adds a hidden command to conduit which will generate the cobra CLI docs.

Also adds a new `conduit-docs` command which will be used via `go:generate` on config files to generate the markdown pages for our developer portal.  
This portion is still WIP, I'll include renders of some of the more interesting configs people can see what they look like for a given revision.

For now I've just got the config graphs working. For example, the filter processor page is as follows:

Sample docs page generated with `conduit-docs`. Also see https://github.com/algorand/docs/pull/964 for the docs pages.
Screenshot 2022-12-20 at 9 26 27 AM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,11 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 # This is the default target, build everything:
 all: conduit cmd/algorand-indexer/algorand-indexer go-algorand idb/postgres/internal/schema/setup_postgres_sql.go idb/mocks/IndexerDb.go
 
-conduit: go-algorand
+conduit: go-algorand conduit-docs
 	go generate ./... && cd cmd/conduit && go build -ldflags="${GOLDFLAGS}"
+
+conduit-docs:
+	go install ./cmd/conduit-docs/
 
 cmd/algorand-indexer/algorand-indexer: idb/postgres/internal/schema/setup_postgres_sql.go go-algorand
 	cd cmd/algorand-indexer && go build -ldflags="${GOLDFLAGS}"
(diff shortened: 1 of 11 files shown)


Eric-Warehime:
Some caveats...There's a tradeoff that I've made by parsing the AST instead of using runtime reflection. It enables us to grab comments from the code (which I believe are erased at runtime), but it makes it pretty difficult to parse objects in other packages like `expression.FilterType`. 

It also means we can't use `go:embed` to load the sample config--although I think putting examples in comments as I've done for filter processor is a decent substitute for this.

Makefile
```diff
@@ -25,9 +25,12 @@ export GO_IMAGE = golang:$(shell go version | cut -d ' ' -f 3 | tail -c +3 )
 # This is the default target, build everything:
 all: conduit cmd/algorand-indexer/algorand-indexer go-algorand idb/postgres/internal/schema/setup_postgres_sql.go idb/mocks/IndexerDb.go
 
-conduit: go-algorand
+conduit: go-algorand conduit-docs
 	go generate ./... && cd cmd/conduit && go build -ldflags="${GOLDFLAGS}"
 
+conduit-docs:
+	go install ./cmd/conduit-docs/
```
winder, line 32: I think you can use `go run ./cmd/conduit-docs` in the go-generate command to avoid needing an install step.

shiqizng: are these docs for dev portal or are they readmes staying in the conduit repo?

shiqizng:
this is great! is it be possible to fix the table width so the type is in a single line.  Right now, the type value for filters look confusing.

Makefile
Eric-Warehime: I intend them for the dev portal, but we could use them for our readmes as well--there is obviously a lot of overlap.

Is it bad to use `go install`? I've wondered why we don't have an install target for indexer in our makefile.

AlgoStephenAkiki (approved):

PR #1383, opened 2022-12-16
Bug Fix: Hides all log messages until config is complete

### LINKED ISSUE #1332: Conduit: CLI errors still print in logger format
The logger should not be used until it has been configured.

Here is an example where the logger is used even though no configuration is even provided:
```
./conduit 
{"__type":"Conduit","_name":"main","level":"info","msg":"Data Directory: [EMPTY] ","time":"2022-11-17T16:17:16Z"}
MakePipelineConfig(): supplied data directory was empty
```

AlgoStephenAkiki:
Resolves #1332

Makes it so that informational log messages are only printed once the Init() function is finished in the pipeline bring up sequence.

This results in error messages being printed to the screen to help the user diagnose those first.

### DIFF
--- a/conduit/pipeline/pipeline.go
+++ b/conduit/pipeline/pipeline.go
@@ -164,6 +164,14 @@ type pipelineImpl struct {
 	pipelineMetadata state
 
 	metricsCallback []conduit.ProvideMetricsFunc
+
+	// preConfigQueue is a pointer to a list of strings that should only be logged
+	// after all configuration of the pipeline is complete (i.e. after init() is successfully called).
+	// The main function of conduit will amass logging strings that will only be output once
+	// everything has been configured.  This is to make it so that the end user sees the direct
+	// error first when launching from the command line. Only after the pipeline has been successfully
+	// configured will log messages be recorded.
+	preConfigQueue *[]string
 }
 
 // state contains the pipeline state.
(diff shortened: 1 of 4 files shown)


conduit/pipeline/pipeline.go
```diff
@@ -164,6 +164,14 @@ type pipelineImpl struct {
...
 
 	metricsCallback []conduit.ProvideMetricsFunc
+
+	// preConfigQueue is a pointer to a list of strings that should only be logged
+	// after all configuration of the pipeline is complete (i.e. after init() is successfully called).
+	// The main function of conduit will amass logging strings that will only be output once
+	// everything has been configured.  This is to make it so that the end user sees the direct
+	// error first when launching from the command line. Only after the pipeline has been successfully
+	// configured will log messages be recorded.
+	preConfigQueue *[]string
```
Eric-Warehime, line 174: I don't know if this is a good idea, but it might be interesting to store `logrus.Entry` here. 

If we generate the log entries with [logger.WithTime](https://github.com/sirupsen/logrus/blob/master/logger.go#L145), when we flush them to the log we'll maintain the time data instead of having everything logged at once when the `init` finishes.

Eric-Warehime (approved):
LGTM, let me know what you think about storing logrus entries instead of strings. Happy to merge it as is though.

winder (changes requested):
Let's fix the logic instead of adding a buffer. There's no reason the logger can't be initialized prior to calling init. I thought this issue was resolved here: https://github.com/algorand/indexer/pull/1369

PR #1384, opened 2022-12-19
enhancement: refactor postgres integration tests

shiqizng:
## Summary

refactor postgres_integration_test.go. replace proc() calls with hard-coded state deltas. 

## Test Plan

regression testing.


types/types.go
```diff
@@ -29,3 +30,9 @@ type SpecialAddresses struct {
 	FeeSink     sdk.Address
 	RewardsPool sdk.Address
 }
+
+// LegercoreValidatedBlock for serialization
+type LegercoreValidatedBlock struct {
+	Blk   bookkeeping.Block
+	Delta ledgercore.StateDelta
+}
```
winder, line 38: So we'll run these vb files through a "convert" function to update tests when we change the interfaces from go-algorand types?

idb/postgres/postgres_integration_test.go
```diff
@@ -172,24 +167,14 @@ func TestAssetCloseReopenTransfer(t *testing.T) {
...
 	// Given // A round scenario requiring subround accounting: AccountA is funded, closed, opts back, and funded again.
 	///////////
-	createAsset := test.MakeAssetConfigTxn(0, total, uint64(6), false, "mcn", "my coin", "http://antarctica.com", test.AccountD)
-	optInA1 := test.MakeAssetOptInTxn(assetid, test.AccountA)
-	optInA2 := test.MakeAssetOptInTxn(assetid, test.AccountA)
-	fundA := test.MakeAssetTransferTxn(assetid, amt, test.AccountD, test.AccountA, basics.Address{})
-	optInB := test.MakeAssetOptInTxn(assetid, test.AccountB)
-	optInC := test.MakeAssetOptInTxn(assetid, test.AccountC)
-	closeA := test.MakeAssetTransferTxn(assetid, 1000, test.AccountA, test.AccountB, test.AccountC)
-	payMain := test.MakeAssetTransferTxn(assetid, amt, test.AccountD, test.AccountA, basics.Address{})
```
winder, line 182: I'm curious what others thing here.

Should we keep this code commented out to serve as documentation for whats inside the validated blocks?

idb/postgres/postgres_integration_test.go
```diff
@@ -545,23 +457,14 @@ func TestIgnoreDefaultFrozenConfigUpdate(t *testing.T) {
...
-		&optin)
+	vb, err := test.ReadValidatedBlockFromFile("test_resources/validated_blocks/IgnoreDefaultFrozenConfigUpdate.vb")
 	require.NoError(t, err)
+	block := ledgercore.MakeValidatedBlock(vb.Blk, vb.Delta)
 
 	//////////
 	// When // We commit the round accounting to the database.
 	//////////
-	err = proc(&rpcs.EncodedBlockCert{Block: block})
+	err = db.AddBlock(&block)
```
winder, line 467: nit: consider moving these lines next to each other since they'll need to be updated together during the interface migration.

winder (commented):
Looks great. Just had some minor comments.

idb/postgres/postgres_integration_test.go
Eric-Warehime: Yeah I think it's fine to keep these as comments for now. Will they be restored once we fully migrate?

types/types.go
shiqizng: I was thinking that vb could be deserialized to[ types.ValidatedBlock](https://github.com/algorand/indexer/blob/develop/types/types.go#L13)

idb/postgres/postgres_integration_test.go
shiqizng: I don't think we'll restore these until we have a way to create the state deltas for tests.

PR #1388, opened 2022-12-20
tech-debt: Update submodule, e2e tests

Eric-Warehime:
## Summary
Updates the submodule to the most recent nightly build.

Also removed the needle in the haystack logging because it seems to cause more confusion than help. Updated the logging to still provide the same set of information. And I updated the circle CI rel-nightly e2e build tar to pull from.

## Test Plan

Ran `e2elive` locally against the updated rel-nightly tar.

### DIFF
--- a/e2e_tests/src/e2e_common/util.py
+++ b/e2e_tests/src/e2e_common/util.py
@@ -164,18 +164,6 @@ def firstFromS3Prefix(
     haystack = []
     found_needle = False
 
-    def report():
-        print(
-            f"""Hello from firstFromS3Prefix() !!!!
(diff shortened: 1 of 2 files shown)


e2e_tests/src/e2e_common/util.py
```diff
@@ -164,18 +164,6 @@ def firstFromS3Prefix(
     haystack = []
     found_needle = False
 
-    def report():
```
AlgoStephenAkiki, line 167: Thank you...I hated this message

AlgoStephenAkiki (approved):

PR #1389, opened 2022-12-22
encoding: Use LenientDecode for genesis objects from algod

Eric-Warehime:
## Summary

The py-algorand-sdk uses sandbox to launch indexer, and has been hitting failures such as [this one](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/1172/workflows/cf2ee38b-8be9-4899-81df-e9e3e2ec0ba1/jobs/5997).

It is due to [this change](https://github.com/algorand/indexer/commit/f56ae83fa3259710e840e681e45d4bfef7ec7d9b#diff-cc2152c5941ca76fdb35ad52ce46b45e7d4c5a3dee1e9f900bf9a1397db6d4b9R92) which uses the SDK's strict json decoder to convert from genesis bytes to the SDK's genesis type. 

The different fields in go-algorand's Genesis type which are not present in the SDK type (`Comment`, and `DevMode`) cause the decode failure.

I've made decoding of the genesis type in Indexer lenient against unknown fields for now.

## Test Plan

Added some fields to the mock algod we use. In the future these will need to be changed to just use bytes instead of the go-algorand Genesis type. But for the short term these test the fix.

### DIFF
--- a/conduit/plugins/exporters/filewriter/file_exporter_test.go
+++ b/conduit/plugins/exporters/filewriter/file_exporter_test.go
@@ -168,7 +168,7 @@ func TestExporterReceive(t *testing.T) {
 		assert.FileExists(t, path)
 
 		var blockData data.BlockData
-		err := util.DecodeFromFile(path, &blockData)
+		err := util.DecodeFromFile(path, &blockData, true)
 		require.Equal(t, basics.Round(i), blockData.BlockHeader.Round)
 		require.NoError(t, err)
 		require.NotNil(t, blockData.Certificate)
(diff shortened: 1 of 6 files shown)


algochoi (approved):
Thank you for the fix here - I can confirm that the changes work on dev mode again.

Is the long term goal to bring parity between the `genesis` type in the SDK and `go-algorand` and re-enable strict decoding?

Eric-Warehime:
> Is the long term goal to bring parity between the genesis type in the SDK and go-algorand and re-enable strict decoding?

I think the Genesis definition in go-algorand and the SDK should be identical, yes. Even though fields like `DevMode` are not very relevant to things outside of the node, it's important to maintain parity since so often we find ourselves copying `genesis.json` files around in different ways.

I don't if strict decoding should be enabled here even after we update the SDK...it seems like most places across our tools have "fail on unknown fields" so I guess it should be the same here as well.

AlgoStephenAkiki (approved):

PR #1390, opened 2022-12-26
REST API: Add compression to REST API

algoidan:
## Summary

## Test Plan

### DIFF
--- a/api/handlers_e2e_test.go
+++ b/api/handlers_e2e_test.go
@@ -1670,7 +1670,7 @@ func TestGetBlockWithCompression(t *testing.T) {
 	serverCtx, serverCancel := context.WithCancel(context.Background())
 	defer serverCancel()
 	opts := defaultOpts
-	listenAddr := "localhost:8888"
+	listenAddr := "localhost:8889"
 	go Serve(serverCtx, listenAddr, db, nil, logrus.New(), opts)
 
 	waitForServer(t, listenAddr)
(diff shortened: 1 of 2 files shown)


api/handlers_e2e_test.go
```diff
@@ -1669,6 +1637,151 @@ func TestGetBlocksTransactionsLimit(t *testing.T) {
...
+
+		client := &http.Client{}
+		req, err := http.NewRequest("GET", "http://"+listenAddr+path, nil)
+		require.NoError(t, err)
+		q := req.URL.Query()
+		if headerOnly {
+			q.Add("header-only", "true")
+		}
+		if useCompression {
+			q.Add("compress", "true")
```
Eric-Warehime, line 1689:
```suggestion
			req.Header.Add("Accept-Encoding", "gzip")
```

api/handlers_e2e_test.go
```diff
@@ -1669,6 +1637,151 @@ func TestGetBlocksTransactionsLimit(t *testing.T) {
...
+	serverCtx, serverCancel := context.WithCancel(context.Background())
+	defer serverCancel()
+	opts := defaultOpts
+	listenAddr := "localhost:8889"
+	go Serve(serverCtx, listenAddr, db, nil, logrus.New(), opts)
+
+	waitForServer(t, listenAddr)
+
+	getBlockFunc := func(t *testing.T, headerOnly bool, useCompression bool) *generated.BlockResponse {
+		path := "/v2/blocks/1"
```
Eric-Warehime, line 1679: If we make this a parameter, we can generalize this across a bunch of different APIs and test to make sure compression works in general and not just for blocks.

Eric-Warehime (changes requested):

api/server.go
```diff
@@ -104,6 +105,13 @@ func Serve(ctx context.Context, serveAddr string, db idb.IndexerDb, dataError fu
 
 	e.Use(middlewares.MakeLogger(log))
 	e.Use(middleware.CORS())
+	e.Use(middleware.GzipWithConfig(middleware.GzipConfig{
+		// we currently support compressed result only for GET /v2/blocks/ API
+		Skipper: func(c echo.Context) bool {
+			return !strings.Contains(c.Path(), "/v2/blocks/") || c.QueryParam("compress") != "true"
+		},
```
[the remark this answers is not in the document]
algoidan: I limited the ability to use gzip just for the `/v2/blocks` endpoint because I think we need to add tests for all endpoints. In order to avoid that and just address the problem we have, I think it is a good start. If you still think we should support all endpoints I can do that, but adding tests for all endpoints might take a while.

As for your second point, do you suggest we use gzip every time the client support `gzip` encoding? even if it didn't request it? My approach was only to add the support and not replace it (don't want to break any 3rd party code)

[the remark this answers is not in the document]
Eric-Warehime: That's fair, let's limit it to the blocks endpoint for now then.

Existing code probably isn't supplying the `Accept-Encoding: gzip` header. Without specifying the header explicitly, the middleware will won't compress anything, so I expect almost all clients to have the same behavior by default.

api/handlers_e2e_test.go
Eric-Warehime: Since we're just testing the blocks endpoint for now, we can pull this into a separate function for other APIs later.

Eric-Warehime (approved):
LGTM

PR #1392, opened 2022-12-27
API: Change block txn default

Eric-Warehime:
## Summary

In #1241 we started using the `MaxTransactionsLimit` parameter to limit the number of transactions returned in the `/v2/block` API. That caused users who expected to receive the whole block to start getting errors when blocks with high numbers of txns were requested.

In order to enable users to set this limit but also enable those who would like to retrieve the entire block at once, I've set this limit to unlimited by default.

Since the `MaxTransactionsLimit` parameter was also previously being used in the `/v2/transactions` endpoint, I've separated the param to a new config option which will only control transactions returned in blocks.


Eric-Warehime:
Will reopen a new PR against the hotfix branch.

PR #1396, opened 2022-12-29
performance: Add /v2/blocks benchmark

Eric-Warehime:
## Summary

The `/v2/blocks` endpoint consumes a large amount of memory when handling blocks with high transactions which have large numbers of inner transactions. This benchmarks memory usage of the API when processing different variations of transactions/inner transactions from a block.

### DIFF
--- a/idb/postgres/postgres_integration_test.go
+++ b/idb/postgres/postgres_integration_test.go
@@ -343,6 +343,76 @@ func TestMultipleWriters(t *testing.T) {
 	assert.Equal(t, amt, balance)
 }
 
+// BenchmarkBlockTransactionsLimit benchmarks memory usage of the block endpoint when exceeding MaxTransactionsLimit.
+func BenchmarkBlockTransactionsLimit(b *testing.B) {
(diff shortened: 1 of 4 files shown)


jhawk28:
It would be useful to have a local pprof http endpoint to pull the heap.

idb/postgres/postgres_integration_test.go
```diff
@@ -343,6 +343,75 @@ func TestMultipleWriters(t *testing.T) {
...
+		for n := uint64(0); n < benchmark.testTxns; n++ {
+			txn := test.MakeAppCallWithInnerTxns(test.AccountA, appAddr, test.AccountB, benchmark.innerTxns)
+			txns = append(txns, &txn)
+		}
+
+		block, err := test.MakeBlockForTxns(prevBlockHeader, txns...)
+		require.NoError(b, err)
+		prevBlockHeader = block.BlockHeader
+
+		err = proc(&rpcs.EncodedBlockCert{Block: block})
```
shiqizng, line 401: this test breaks after block_processor is refactored to query the state deltas.

winder:
Closing because the test would soon become obsolete. It depends on the block processor.

Issue #1395, opened 2022-12-28
Investigate /v2/blocks high memory usage


Eric-Warehime:
Using the `/v2/blocks` endpoint (with Indexer v2.15.0) will currently return an error when the total number of transactions in the returned block is greater than the `MaxTransactionsLimit`. See [here](https://github.com/algorand/indexer/blob/develop/idb/postgres/postgres.go#L494) for the implementation.

This becomes a problem when there are a large number of inner transactions since the returned data recursively reconstructs transactions. By that I mean that a root txn will have inner txns nested inside of it, and an inner txn will have a copy of the root txn which also has inner txns.

The result is that not only does the memory usage scale with the number of transactions in a block, but also with the number of inner transactions. 

Based on benchmarking, a `MaxTransactionsLimit` of 1000   will consume ~3GB of memory if the transactions in the block each have 200 inner transactions.

Desired Result:

Identify and implement a solution to limit memory consumption on blocks containing large numbers of inner transactions.

Eric-Warehime:
Benchmark results
```
BenchmarkBlockTransactionsLimit//v2/blocks_txns:_10_innerTxns:_0-10         	     333	   3558479 ns/op	  132312 B/op	     380 allocs/op
BenchmarkBlockTransactionsLimit//v2/blocks_txns:_100_innerTxns:_0-10        	     153	   7806742 ns/op	 1277780 B/op	    3376 allocs/op
BenchmarkBlockTransactionsLimit//v2/blocks_txns:_1000_innerTxns:_0-10       	      24	  46342839 ns/op	12674837 B/op	   33306 allocs/op
BenchmarkBlockTransactionsLimit//v2/blocks_txns:_10_innerTxns:_10-10        	     139	   8603460 ns/op	 1688362 B/op	    1357 allocs/op
BenchmarkBlockTransactionsLimit//v2/blocks_txns:_100_innerTxns:_100-10      	       3	 412493597 ns/op	128043600 B/op	  103353 allocs/op
BenchmarkBlockTransactionsLimit//v2/blocks_txns:_1000_innerTxns:_200-10     	       1	8535142250 ns/op	2976411216 B/op	 2034852 allocs/op
```

Eric-Warehime:
WIP PR for the benchmarks used to generate the above #1396

jhawk28:
[...]

PR #1397, opened 2023-01-03
API: Hotfix block endpoint

Eric-Warehime:
#1395 Shows more details. 

At a high level, the SQL used to fetch transactions for the `/v2/blocks` endpoint balloons the amount of memory usage when the block contains large numbers of inner transactions. These inner transactions end up getting discarded anyways after the query returns and is deserialized. This patch improves the query and memory usage by only fetching root transactions to begin with.

## Test Plan

Ran benchmarks from #1396 to get memory usage data. The result of a block containing 250 transactions, each of which has 250 inner transactions, w/ a `MaxTransactionsLimit` set to the default of 1000 shows pre-patch memory usage around  9GB, and post-patch memory usage around 2GB.

Before patch:
```
BenchmarkBlockTransactions//v2/blocks_txns:_250_innerTxns:_250-10         	       1	9067736792 ns/op	3213189536 B/op	 2537356 allocs/op
```

After patch:
```
BenchmarkBlockTransactions//v2/blocks_txns:_250_innerTxns:_250-10         	       1	2378328417 ns/op	800444504 B/op	  633329 allocs/op
```

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -457,7 +457,7 @@ func (db *IndexerDb) GetBlock(ctx context.Context, round uint64, options idb.Get
 
 	if options.Transactions {
 		out := make(chan idb.TxnRow, 1)
-		query, whereArgs, err := buildTransactionQuery(idb.TransactionFilter{Round: &round, Limit: options.MaxTransactionsLimit + 1})
+		query, whereArgs, err := buildTransactionQuery(idb.TransactionFilter{Round: &round, Limit: options.MaxTransactionsLimit + 1, ReturnRootTxnsOnly: true})
 		if err != nil {
 			err = fmt.Errorf("txn query err %v", err)
 			out <- idb.TxnRow{Error: err}
@@ -658,6 +658,9 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	if tf.RekeyTo != nil && (*tf.RekeyTo) {
 		whereParts = append(whereParts, "(t.txn -> 'txn' -> 'rekey') IS NOT NULL")
 	}
+	if tf.ReturnRootTxnsOnly {
+		whereParts = append(whereParts, "t.txid IS NOT NULL")
+	}
 
 	// If returnInnerTxnOnly flag is false, then return the root transaction
 	if !tf.ReturnInnerTxnOnly {
@@ -671,7 +674,7 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	}
 
 	// join in the root transaction if the returnInnerTxnOnly flag is false
-	if !tf.ReturnInnerTxnOnly {
+	if !tf.ReturnInnerTxnOnly || tf.ReturnRootTxnsOnly {
 		query += " LEFT OUTER JOIN txn root ON t.round = root.round AND (t.extra->>'root-intra')::int = root.intra"
 	}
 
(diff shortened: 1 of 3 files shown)


idb/postgres/postgres.go
```diff
@@ -671,7 +674,7 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	}
 
 	// join in the root transaction if the returnInnerTxnOnly flag is false
-	if !tf.ReturnInnerTxnOnly {
+	if !tf.ReturnInnerTxnOnly || tf.ReturnRootTxnsOnly {
```
shiqizng, line 677: is this join needed when tf.ReturnRootTxnsOnly is true? I think `t.txid IS NOT NULL` returns all the root txns already?

shiqizng: at this time, when tf.ReturnInnerTxnOnly is true, it overrides tf.ReturnRootTxnsOnly setting at line 669.  is it what we want? do we return only the inner txns if we get a "bad" query where tf.ReturnInnerTxnOnly==true and tf.ReturnRootTxnsOnly==true? maybe, a valid query should only contain ReturnRootTxnsOnly or ReturnInnerTxnOnly.

Eric-Warehime: I think you're correct, we don't need the extra join here.

winder: Yes, I think we specifically want to avoid the join
```suggestion
	if !tf.ReturnInnerTxnOnly && !tf.ReturnRootTxnsOnly {
```

idb/postgres/postgres.go
```diff
@@ -658,6 +658,9 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	if tf.RekeyTo != nil && (*tf.RekeyTo) {
 		whereParts = append(whereParts, "(t.txn -> 'txn' -> 'rekey') IS NOT NULL")
 	}
+	if tf.ReturnRootTxnsOnly {
+		whereParts = append(whereParts, "t.txid IS NOT NULL")
+	}
 
 	// If returnInnerTxnOnly flag is false, then return the root transaction
 	if !tf.ReturnInnerTxnOnly {
```
winder, line 666: This is also needed to make sure we don't try to reference `root`:
```suggestion
	if !tf.ReturnInnerTxnOnly && !tf.ReturnRootTxnsOnly {
```

idb/postgres/postgres.go
```diff
@@ -658,9 +658,12 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	if tf.RekeyTo != nil && (*tf.RekeyTo) {
 		whereParts = append(whereParts, "(t.txn -> 'txn' -> 'rekey') IS NOT NULL")
 	}
+	if tf.ReturnRootTxnsOnly {
+		whereParts = append(whereParts, "t.txid IS NOT NULL")
+	}
 
 	// If returnInnerTxnOnly flag is false, then return the root transaction
-	if !tf.ReturnInnerTxnOnly {
+	if !(tf.ReturnInnerTxnOnly || tf.ReturnRootTxnsOnly) {
```
Eric-Warehime, line 666: The `ReturnInnerTxnOnly` variable name does not really represent what this is doing--it makes it much harder to reason about these conditions.

Let's update the name on the develop branch--avoiding a bunch of refactoring on this hotfix branch seems more important.

I've updated the most recent patch to represent what we want, which is that if neither of these options is enabled we will fetch `root.txn` and _also_ if neither is enabled we'll add the JOIN which references the intra index.

winder (approved):
Changes look good. Tested them locally and it works great. Some big blocks that used to timeout are now returning almost instantly.

PR #1412, opened 2023-01-11
tests: Add a way to pause the e2e docker container for debugging.

winder:
## Summary

I frequently find myself wanting to inspect the e2elive temp files in the docker container. These changes allow you to do so by setting `export EXTRA=--keep-alive` prior to running `make e2e`.

## Test Plan

Manual testing.

### DIFF
--- a/e2e_tests/src/e2e_indexer/e2elive.py
+++ b/e2e_tests/src/e2e_indexer/e2elive.py
@@ -36,6 +36,8 @@ def main():
 
     ap = argparse.ArgumentParser()
     ap.add_argument("--keep-temps", default=False, action="store_true")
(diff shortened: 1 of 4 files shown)


e2e_tests/src/e2e_indexer/e2elive.py
```diff
@@ -131,6 +137,8 @@ def main():
                     with open(p) as nf:
                         for line in nf:
                             logger.error("   {}".format(line))
+        if found is False:
```
tzaffi, line 140: It seems that when arriving to this line it will always be the case that `found is False`. Did you intend to set `found = True` after line 135 ?

tzaffi (approved):
LGTM

Eric-Warehime (approved):

PR #1415, opened 2023-01-13
New Feature: Adds the sync mode importer 

AlgoStephenAkiki:
Keeps all interfaces the same but adds specific request functionality to the algod importer.


helpers/helpers.go
```diff
@@ -31,6 +37,625 @@ func ConvertParams(params basics.AssetParams) sdk.AssetParams {
 	}
 }
 
+func L2cBlockHeader(input bookkeeping.BlockHeader) sdk.BlockHeader {
```
Eric-Warehime, line 40: Lets remove these until the followup PR where we start using them.

Eric-Warehime (changes requested):
Can you add a description to the PR also?

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -32,6 +32,11 @@ type algodImporter struct {
 	cancel  context.CancelFunc
 }
 
+func (algodImp *algodImporter) OnComplete(input data.BlockData) error {
+	_, err := algodImp.aclient.SetSyncRound(input.Round() + 1).Do(algodImp.ctx)
```
shiqizng, line 36: what if input.Round() is < current round in ledger? I think you would also need to call SetSyncRound in Init? 
and does algod node need to have a large enough `MaxAcctLookback` for indexer to import older rounds?

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -32,6 +32,11 @@ type algodImporter struct {
 	cancel  context.CancelFunc
 }
 
+func (algodImp *algodImporter) OnComplete(input data.BlockData) error {
```
winder, line 35: The archival mode importer is still useful for a lot of things. I think the Sync Mode importer should be a new plugin

shiqizng:
can we rebase this PR to develop instead of feature/sync-mode?

conduit/plugins/importers/algod/algod_importer.go
Eric-Warehime: The node will be required to have the delta in the cache. This means the indexer's round must be within `MaxAcctLookback` of the node's round. The data node will [set the sync round](https://github.com/algorand/go-algorand/pull/5009/files#diff-c10bfda5adaaf4573b0ae153fe9eba6a3be539853363b7b630d2f6da5f94c9adR118) to the latest round on startup to make sure it doesn't get ahead of the indexer.

AlgoStephenAkiki:
> can we rebase this PR to develop instead of feature/sync-mode?

I have rebased this against develop

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/algorand/avm-abi v0.2.0
 	github.com/algorand/go-algorand v0.0.0-20220211161928-53b157beb10f
 	github.com/algorand/go-algorand-sdk v1.22.1-0.20221209152459-223656f08456
```
Eric-Warehime, line 10: Do we still need v1?

conduit/plugins/importers/syncmode/syncmode_importer_config.go
```diff
@@ -0,0 +1,9 @@
+package syncmode
+
+// Config specific to the sync mode importer
+type Config struct {
+	// Algod netaddr string
+	NetAddr string `yaml:"netaddr"`
+	// Algod rest endpoint token
+	Token string `yaml:"token"`
+}
```
Eric-Warehime, line 9: Can you add the docs generation? I've just copied this from the algod importer. The readme is in https://github.com/algorand/indexer/tree/develop/cmd/conduit-docs if you want to change or add anything.
```suggestion

//go:generate conduit-docs ../../../../conduit-docs/

//Name: conduit_importers_syncmode

// Config specific to the sync mode importer
type Config struct {
	// <code>netaddr</code> is the Algod network address. It must be either an <code>http</code> or <code>https</code> URL.
	NetAddr string `yaml:"netaddr"`
	// <code>token</code> is the Algod API endpoint token.
	Token string `yaml:"token"`
}
```

conduit/plugins/importers/syncmode/syncmode_importer.go
```diff
@@ -0,0 +1,168 @@
+package syncmode
```
Eric-Warehime, line 1: Should we make the package something like the importer name? `syncalgod` or something like that. I think that "syncmode" could be confusing as a name without the algod context.

conduit/plugins/importers/syncmode/metrics.go
```diff
@@ -0,0 +1,18 @@
...
+// Prometheus metric names
+const (
+	GetAlgodRawBlockTimeName = "get_algod_raw_block_time_sec"
+)
+
+// Initialize the prometheus objects
+var (
+	GetAlgodRawBlockTimeSeconds = prometheus.NewSummary(
+		prometheus.SummaryOpts{
+			Subsystem: "indexer_daemon",
```
shiqizng, line 14: maybe the subsystem/prefix here should be the plugin name. same for algod_importer.

conduit/plugins/importers/syncmode/syncmode_importer.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/algorand/indexer/conduit/plugins/importers"
+	"github.com/algorand/indexer/data"
+
+	"github.com/algorand/go-algorand-sdk/v2/client/v2/algod"
+	sdk "github.com/algorand/go-algorand-sdk/v2/types"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/rpcs"
+)
+
+const importerName = "sync-algod"
```
shiqizng, line 25: it looks like we have been using snake case for these, e.g. `block_evaluator` and `filter_processor`

conduit/plugins/importers/syncmode/syncmode_importer.go
```diff
@@ -0,0 +1,168 @@
...
+	_, err := sm.aclient.SetSyncRound(input.Round() + 1).Do(sm.ctx)
+	return err
+}
+
+//go:embed sample.yaml
+var sampleConfig string
+
+var syncModeImporterMetadata = conduit.Metadata{
+	Name:         importerName,
+	Description:  "Importer for fetching blocks from an algod REST API.",
```
shiqizng, line 45: this description is the same as algod_importer. does anyone else have a better description for this plugin?

conduit/plugins/importers/syncmode/syncmode_importer.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	return nil
+}
+
+func (sm *syncModeImporter) GetBlock(rnd uint64) (data.BlockData, error) {
+	var blockbytes []byte
+	var err error
+	var blk data.BlockData
+
+	for retries := 0; retries < 3; retries++ {
```
shiqizng, line 121: what happens if the node is behind the indexer by many rounds? the pipeline exits after n retries?

AlgoStephenAkiki: I guess so.  Thoughts @Eric-Warehime?

idb/postgres/postgres_integration_test.go
```diff
@@ -11,6 +11,9 @@ import (
 	"testing"
 	"time"
 
+	crypto2 "github.com/algorand/go-algorand-sdk/v2/crypto"
+	"github.com/algorand/go-algorand-sdk/v2/encoding/msgpack"
+	sdk "github.com/algorand/go-algorand-sdk/v2/types"
```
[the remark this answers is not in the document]
AlgoStephenAkiki: I didn't get an error when linting, but I've re-arranged them to be better (hopefully).

helpers/helpers.go
```diff
@@ -1,3 +1,4 @@
+//nolint
```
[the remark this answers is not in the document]
AlgoStephenAkiki: Initially the helper functions I wrote had so many errors that I put this in to side-step the issue (since we were deleting them anyway).  I've removed it now since I've removed what I wrote.

go.mod
AlgoStephenAkiki: I don't think we do but I'd rather save that for another PR.

conduit/plugins/importers/syncmode/syncmode_importer.go
AlgoStephenAkiki: Good idea.  Done.

conduit/plugins/importers/syncmode/metrics.go
AlgoStephenAkiki: Good idea...done.

conduit/plugins/importers/syncmode/syncmode_importer.go
AlgoStephenAkiki: Done

conduit/plugins/importers/syncmode/syncmode_importer.go
AlgoStephenAkiki: I changed it a little.  Let me know what you think.  FYI @Eric-Warehime

conduit/plugins/importers/syncmode/syncmode_importer.go
Eric-Warehime: I think so. If we wanted to we could add in catchpoint/slow catchup to the init method for the sync importer. That way it would match what we have for the local ledger. 

But I think it's a better answer to just require the node to be in the correct state. Manually run catchup on your node and set the sync round to be whichever round you want to start from. Then start the indexer.

go.mod
Eric-Warehime: Sure, sounds fine.

conduit/plugins/importers/syncmode/syncmode_importer.go
winder: It probably makes sense to add that in a follow up.

For this PR a clear error message that recommends how to fix the bad state would be good.

conduit/plugins/importers/syncalgod/syncalgod_importer.go
```diff
@@ -0,0 +1,168 @@
...
+	_, err := sm.aclient.SetSyncRound(input.Round() + 1).Do(sm.ctx)
+	return err
+}
+
+//go:embed sample.yaml
+var sampleConfig string
+
+var syncAlgodModeImporterMetadata = conduit.Metadata{
+	Name:         importerName,
+	Description:  "Importer for fetching blocks from an algod REST API using synch round and ledger delta algod calls.",
```
shiqizng, line 45: synch, typo?

conduit/plugins/importers/syncalgod/syncalgod_importer.go
```diff
@@ -0,0 +1,168 @@
...
+	sm.aclient = client
+
+	genesisResponse, err := client.GetGenesis().Do(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	genesis := sdk.Genesis{}
+
+	err = protocol.DecodeJSON([]byte(genesisResponse), &genesis)
```
shiqizng, line 96: please use Decode() from "github.com/algorand/go-algorand-sdk/v2/encoding/json". Let's not introduce new go-algorand dependencies if possible.

conduit/plugins/importers/syncalgod/syncalgod_importer.go
```diff
@@ -0,0 +1,168 @@
...
+func (sm *syncModeImporter) Config() string {
+	s, _ := yaml.Marshal(sm.cfg)
+	return string(s)
+}
+
+func (sm *syncModeImporter) Close() error {
+	if sm.cancel != nil {
+		sm.cancel()
+	}
+	return nil
```
shiqizng, line 113: should we call UnsetSyncRound in Close()?

Eric-Warehime: If we want to let the node continue catching up once conduit stops then yes.

I think in our case it's best to leave it set so that we can start/stop conduit and the node will still be usable. If we let it catchup freely then when conduit starts again the deltas we need will likely no longer be available.

Because we're making it so much more lightweight to run an algod node for indexer by removing the requirement of it being archival, I think it's not unreasonable to assume that algod nodes service indexer/conduit are dedicated instead of shared across multiple services/use cases.

conduit/plugins/importers/syncalgod/syncalgod_importer.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/algorand/indexer/data"
+
+	"github.com/algorand/go-algorand-sdk/v2/client/v2/algod"
+	"github.com/algorand/go-algorand-sdk/v2/encoding/json"
+	sdk "github.com/algorand/go-algorand-sdk/v2/types"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/rpcs"
+)
+
+const importerName = "sync_algod"
```
Eric-Warehime, line 26: See https://github.com/algorand/indexer/pull/1431/files#diff-6efc74da31cfe4605350d8ccb909cc366261084bb90581104d5652fb494046aaR7
You need to add the import to `all.go` so it gets loaded in by the conduit command.

conduit/plugins/importers/syncalgod/syncalgod_importer.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			return blk, err
+		}
+
+		// We aren't going to do anything with the new delta until we get everything
+		// else converted over
+		_, err = sm.aclient.GetLedgerStateDelta(rnd).Do(sm.ctx)
+		if err != nil {
+			return blk, err
+		}
```
Eric-Warehime, line 152: The block processor skips round 0. We need to for the delta endpoint as well since it won't be able to fetch a delta for round 0 ever.
https://github.com/algorand/indexer/blob/develop/conduit/plugins/processors/blockprocessor/block_processor.go#L129

```suggestion
		// Round 0 has no delta associated with it
		if rnd != 0 {
			_, err = sm.aclient.GetLedgerStateDelta(rnd).Do(sm.ctx)
			if err != nil {
				return blk, err
			}
		}
```

Eric-Warehime (changes requested):

conduit/plugins/importers/syncalgod/syncalgod_importer.go
winder: We'll also want to align this name with the new name. Have you decided on Follower?

AlgoStephenAkiki: So would we name this `algod_follower`?

AlgoStephenAkiki: I've renamed to `algod_follower`

Eric-Warehime (approved):

shiqizng (approved):

PR #1417, opened 2023-01-17
db: Remove AccountTotals from metastate

### LINKED ISSUE #1349: License - remove account totals from idb
```
db.setMetastate(
tx, schema.AccountTotals, string(encoding.EncodeAccountTotals(&totals)))
```
and related code can be removed.

Eric-Warehime:
## Summary

Resolves #1349  
Removes the AccountTotals from the metastate table. These are not referenced by any API, so it should be a no-op from the end user standpoint.

I've removed the related tests which reference this.

### DIFF
--- a/idb/postgres/internal/encoding/encoding.go
+++ b/idb/postgres/internal/encoding/encoding.go
@@ -639,22 +639,6 @@ func DecodeMigrationState(data []byte) (types.MigrationState, error) {
 	return state, nil
 }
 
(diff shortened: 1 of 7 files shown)


shiqizng (approved):
look good. thanks for removing these.

AlgoStephenAkiki (approved):

winder (approved):
Thanks, LGTM.

There would still be stale account totals in old installations, but it looks like we would never read them so they wouldn't cause any problems.

PR #1418, opened 2023-01-17
tech debt: Rename and simplify app log transactions query.

winder:
## Summary

Rename `ReturnInnerTxnOnly` to `IncludeInnerTxns`, because you could still match on the root transaction.

I was torn between calling this variable `IncludeInnerTxns` and `SkipRootTxnJoin`, in the end I chose the former because it's closer to the behavior we want where the latter is the implementation.

## Test Plan

Existing regression tests.

### DIFF
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -253,7 +253,10 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 	}
 
 	var stxn *sdk.SignedTxnWithAD
-	if row.Txn != nil {
+	if row.RootTxn != nil && row.Txn != nil {
+		// see postgres.go:yieldTxnsThreadSimple
+		return generated.Transaction{}, fmt.Errorf("%d:%d Txn and RootTxn should be mutually exclusive", row.Round, row.Intra)
+	} else if row.Txn != nil {
 		stxn = row.Txn
 	} else if row.RootTxn != nil {
 		stxn = row.RootTxn
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -246,6 +246,7 @@ type rowData struct {
 }
 
 // txnRowToTransaction parses the idb.TxnRow and generates the appropriate generated.Transaction object.
+// If the TxnRow contains a RootTxn, the generated.Transaction object will be the root txn.
 func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 	if row.Error != nil {
 		return generated.Transaction{}, row.Error
@@ -283,12 +284,7 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 		Sig:      sigToTransactionSig(stxn.Sig),
 	}
 
-	var txid string
-	if row.Extra.RootIntra.Present {
-		txid = row.Extra.RootTxid
-	} else {
-		txid = crypto.TransactionIDString(stxn.Txn)
-	}
+	txid := crypto.TransactionIDString(stxn.Txn)
 	txn.Id = &txid
 	txn.Signature = &sig
 
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -287,7 +287,12 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 		Sig:      sigToTransactionSig(stxn.Sig),
 	}
 
-	txid := crypto.TransactionIDString(stxn.Txn)
+	var txid string
+	if row.Extra.RootIntra.Present {
+		txid = row.Extra.RootTxid
+	} else {
+		txid = crypto.TransactionIDString(stxn.Txn)
+	}
 	txn.Id = &txid
 	txn.Signature = &sig
 
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1451,10 +1451,9 @@ func (si *ServerImplementation) fetchTransactions(ctx context.Context, filter id
 				return err
 			}
 
-			// The root txn has already been added.
-			// If we also want to return inner txns, we cannot deduplicate the
-			// results as inner txns all share the same txn ID as its root txn.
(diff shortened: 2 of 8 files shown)


api/handlers.go
```diff
@@ -1451,15 +1451,10 @@ func (si *ServerImplementation) fetchTransactions(ctx context.Context, filter id
...
 
-			// Do not return inner transactions.
-			if tx.Id == nil {
-				continue
-			}
-
 			// The root txn has already been added.
 			// If we also want to return inner txns, we cannot deduplicate the
 			// results as inner txns all share the same txn ID as its root txn.
-			if _, ok := rootTxnDedupeMap[*tx.Id]; ok && !filter.ReturnInnerTxnOnly {
```
winder, line 1462: I believe this extra check is no longer needed after modifying `txnRowToTransaction` to compute a txid from the inner transaction.

api/converter_utils.go
```diff
@@ -283,12 +287,7 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 		Sig:      sigToTransactionSig(stxn.Sig),
 	}
 
-	var txid string
-	if row.Extra.RootIntra.Present {
-		txid = row.Extra.RootTxid
-	} else {
-		txid = crypto.TransactionIDString(stxn.Txn)
-	}
+	txid := crypto.TransactionIDString(stxn.Txn)
```
algochoi, line 290: Unfortunately, this will change the behavior on how inner txn txids are returned in this API: https://github.com/algorand/indexer/blob/develop/api/handlers.go#L752

It feels like the intended behavior is for the inner txid to always match the root txid, especially since inner txns have different logic to calculate their "ID"s. I added a test here to check this TxID behavior: https://github.com/winder/indexer/pull/2

winder: AFAIK inner transactions aren't returned in the API. Are they returned in the logs endpoint?

algochoi: The log endpoint response returns the logs and the TxID that logged it: https://github.com/algorand/indexer/blob/develop/api/generated/common/types.go#L249

So if an inner txn calls app #12345 that logs something, then the log response for app #12345 would have an array of `{log(s)-TxID}`, where TxID is currently the root TxID.

winder: Thanks for the explanation, and the test case. I'll revert this part.

idb/postgres/postgres.go
```diff
@@ -673,18 +673,18 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	}
 
 	// If returnInnerTxnOnly flag is false, then return the root transaction
```
Eric-Warehime, line 675: nit: update comment.

idb/postgres/postgres.go
```diff
@@ -673,18 +673,18 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
...
 		query = "SELECT t.round, t.intra, t.txn, NULL, t.extra, t.asset, h.realtime FROM txn t JOIN block_header h ON t.round = h.round"
+	} else {
+		query = "SELECT t.round, t.intra, t.txn, root.txn, t.extra, t.asset, h.realtime FROM txn t JOIN block_header h ON t.round = h.round"
 	}
 
 	if joinParticipation {
 		query += " JOIN txn_participation p ON t.round = p.round AND t.intra = p.intra"
 	}
 
 	// join in the root transaction if the returnInnerTxnOnly flag is false
```
Eric-Warehime, line 686: nit: update comment

idb/idb.go
```diff
@@ -240,12 +240,12 @@ type TransactionFilter struct {
...
-	// instead of the root txn.
-	ReturnInnerTxnOnly bool
+	// If this flag is set to true, then the query returns inner transactions
+	// instead of resolving them as the root transaction.
+	IncludeInnerTxns bool
 
-	// If this flag is set to true, then the query returns only root txns
-	// and no inner txns
+	// If this flag is set to true, then the query only returns root
+	// transactions and skips matching or returning inner transactions.
```
Eric-Warehime, line 248: This is confusing me--the descriptions sound like they do the same thing. That being, return only root txns instead of inner txns. Or in other words, `!IncludeInnerTxns == ReturnRootTxnsOnly`. But I think they're actually different.

winder: Aah back to the drawing board. When you say it that way these names still sound pretty bad.  I agree though, the flags are confusing. I think it's because what we're doing isn't intuitive, so there is no basis for what the variables are doing.

The unintuitive behavior: when the query matches an inner transaction the root transaction is included in the result and the API returns only the root transaction.

The fields, and obnoxious but maybe more technically correct names...

* IncludeInnerTxns -> DoNotConvertInnerTxnToRootTxn
  * Actually return the inner transaction instead of converting it into the root transaction.

* ReturnRootTxnsOnly -> SkipInnerTransactionsInQuery
  * Skip inner transactions at the query level, completely side-stepping the conversion process

idb/idb.go
```diff
@@ -240,13 +240,13 @@ type TransactionFilter struct {
...
+	// If this flag is set to true, then the query returns inner transactions
+	// instead of converting them to the root transaction.
+	SkipInnerTransactionConversion bool
 
-	// If this flag is set to true, then the query returns only root txns
-	// and no inner txns
-	ReturnRootTxnsOnly bool
+	// If this flag is set to true, then the query only returns root
+	// transactions and skips matching or returning inner transactions.
+	SkipInnerTransactions bool
```
winder, line 249: @Eric-Warehime do these sound any better?

Eric-Warehime: Yes, sorry this review sort of fell of my radar somehow.

Eric-Warehime (approved):

algochoi (commented):
I think the comments still need to be updated after the flag name change

PR #1431, opened 2023-01-21
e2e-test: E2e sync mode test

Eric-Warehime:
A working e2e test using the sync mode.

It uses my personal branch of go-algorand as the submodule since the PR for sync mode has not been merged yet.  
I also hard coded a bunch of things, you can't run the regular indexer in parallel to the sync mode, and this is based off of the PR which does nothing with the delta as of right now.


Eric-Warehime:
@AlgoStephenAkiki FYI

winder:
Could you rebase on develop? I tried to push the SDK upgrade separately to cleanup the diffs

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
```diff
@@ -0,0 +1,15 @@
...
+from e2e_conduit.scenarios import Scenario
+
+
+def follower_indexer_scenario(sourcenet):
+    return Scenario(
+        "follower_indexer_scenario",
+        importer=importers.FollowerAlgodImporter(sourcenet),
+        processors=[
+            processors.BlockEvaluator(),
+        ],
```
winder, line 13: We shouldn't need the block evaluator with this importer

e2e_tests/src/e2e_conduit/fixtures/importers/follower_algod.py
```diff
@@ -0,0 +1,115 @@
...
+                tarpath = os.path.join(conduit_dir, tarname)
+            else:
+                tarpath = os.path.join(conduit_dir, tarname)
+            success = firstFromS3Prefix(s3, bucket, prefix, tarname, outpath=tarpath)
+            if not success:
+                raise Exception(
+                    f"failed to locate tarname={tarname} from AWS S3 path {bucket}/{prefix}"
+                )
+            source_is_tar = True
+            self.sourcenet = tarpath
```
winder, line 80: This seems problematic. Can follow mode support a predefined ledger?

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
Eric-Warehime: We do for now since the importer does nothing with the retrieved delta https://github.com/algorand/indexer/blob/develop/conduit/plugins/importers/algodfollower/algodfollower_importer.go#L156

e2e_tests/src/e2e_conduit/fixtures/importers/follower_algod.py
Eric-Warehime: Kind of...The problem I had w/ the existing e2e setup was that when you run `goal network start` with predefined data you get two nodes started at round 40 (or however many rounds there are). 

This wasn't a problem when all we wanted was the block data since you can retrieve any round, but in order to get the delta for those rounds we need to start at 0 and progress to 40. 

Potentially in the future we can add some persistence to deltas so that you can start a network w/ a larger number of deltas stored instead of having to populate the cache round by round.

winder: Sounds like a larger discussion is needed to figure out whether or not we're outgrowing the e2e artifact.

Thinking out loud though... I think we could take the existing network snapshot and attach a follower node to it. You'd have to pass in a custom "phonebook" to the follow node so that it caught up from the test node.

winder: I was mistaken -- what you have here looks great.

e2e_tests/src/e2e_conduit/fixtures/importers/follower_algod.py
```diff
@@ -0,0 +1,115 @@
...
+        for root, dirs, files in os.walk(os.path.join(tempnet, 'Node', 'tbd-v1')):
+                for f in files:
+                    if ".sqlite" in f:
+                        os.remove(os.path.join(root, f))
+        cf = {}
+        with open(os.path.join(tempnet, "Node", "config.json"), "r") as config_file:
+            cf = json.load(config_file)
+            cf['EnableFollowMode'] = True
+        with open(os.path.join(tempnet, "Node", "config.json"), "w") as config_file:
+            config_file.write(json.dumps(cf))
```
winder, line 99: Very clever. A comment here might have helped, I jumped to the wrong conclusion when I saw you fetching the dataset.
```suggestion
        # Reset the secondary node, and enable follow mode.
        # This is what conduit will connect to for data access.
        for root, dirs, files in os.walk(os.path.join(tempnet, 'Node', 'tbd-v1')):
                for f in files:
                    if ".sqlite" in f:
                        os.remove(os.path.join(root, f))
        cf = {}
        with open(os.path.join(tempnet, "Node", "config.json"), "r") as config_file:
            cf = json.load(config_file)
            cf['EnableFollowMode'] = True
        with open(os.path.join(tempnet, "Node", "config.json"), "w") as config_file:
            config_file.write(json.dumps(cf))
```

e2e_tests/src/e2e_conduit/fixtures/importers/follower_algod.py
```diff
@@ -0,0 +1,115 @@
...
+        tempnet = os.path.join(conduit_dir, "net")
+        if source_is_tar:
+            xrun(["tar", "-C", conduit_dir, "-x", "-f", self.sourcenet])
+        else:
+            xrun(["rsync", "-a", self.sourcenet + "/", tempnet + "/"])
+        blockfiles = glob.glob(
+            os.path.join(conduit_dir, "net", "Primary", "*", "*.block.sqlite")
+        )
+        self.last = countblocks(blockfiles[0])
+        for root, dirs, files in os.walk(os.path.join(tempnet, 'Node', 'tbd-v1')):
```
winder, line 90: Very clever. A comment here might have helped, I jumped to the wrong conclusion when I saw you fetching the dataset.
```suggestion
        # Reset the secondary node, and enable follow mode.
        # This is what conduit will connect to for data access.
        for root, dirs, files in os.walk(os.path.join(tempnet, 'Node', 'tbd-v1')):
```

shiqizng (approved):

winder (approved):

PR #1445, opened 2023-01-30
build: Fix lint/vet

Eric-Warehime:
## Summary

It looks like `make lint` is failing on a few things--not sure how some of these ever passed CI, but here are the fixes.

* The fmt string in `conduit/plugins/processors/filterprocessor/gen/generate.go` is being skipped for linting, but `go vet` is failing. Use `bytes.Buffer.WriteString` instead of `fmt.Fprint`.
* There are leftover references to `appCall` in `handlers_e2e_test.go` where `appCall` was commented out for the refactoring work. Comment these out.
* It looks like we missed a conversion to v2 of the SDK in `api/util_test.go`.

### DIFF
--- a/api/handlers_e2e_test.go
+++ b/api/handlers_e2e_test.go
@@ -1304,7 +1304,7 @@ func TestLookupInnerLogs(t *testing.T) {
 			require.NotNil(t, response.LogData)
 			ld := *response.LogData
 			require.Equal(t, 1, len(ld))
-			require.Equal(t, appCall.Txn.ID().String(), ld[0].Txid)
+			// require.Equal(t, appCall.Txn.ID().String(), ld[0].Txid)
 			require.Equal(t, len(tc.logs), len(ld[0].Logs))
 			for i, log := range ld[0].Logs {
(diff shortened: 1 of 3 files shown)


api/handlers_e2e_test.go
```diff
@@ -1304,7 +1304,7 @@ func TestLookupInnerLogs(t *testing.T) {
 			require.NotNil(t, response.LogData)
 			ld := *response.LogData
 			require.Equal(t, 1, len(ld))
-			require.Equal(t, appCall.Txn.ID().String(), ld[0].Txid)
+			// require.Equal(t, appCall.Txn.ID().String(), ld[0].Txid)
```
winder, line 1307: @algochoi could you help us update these new tests? It looks like there was a discrepancy between master and develop since we no longer have the original app calls

algochoi: sure thing, will look into it

algochoi: PR here: https://github.com/Eric-Warehime/indexer/pull/1

I wasn't sure how the refactoring was being done on a macro scale so feel free to make any stylistic changes as well

PR #1450, opened 2023-01-31
tests: Update e2e dockerfile

Eric-Warehime:
## Summary

Changed the indexer e2e tests to download and install the go-algorand binaries via the updater script. Also updated the circleCI config to run the nightly tests using the nightly binaries--whereas the regular tests will use the stable binaries.

The conduit e2e tests will need separate changes as well--I'd like to wait on those until we land all of the follower node changes/tests.

## Test Plan

`make e2e-nightly`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ commands:
           no_output_timeout: 15m
       - run: make test-generate
       - run: make fakepackage
-      - run: make e2e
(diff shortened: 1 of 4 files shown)


winder (commented):
Looks great. The failure is my fault, I'll get upstream fixed ASAP.

winder (approved):

PR #1451, opened 2023-01-31
bugfix: Fix missing var initialize.

winder:
## Summary

Not sure how I missed this.


Eric-Warehime:
I was actually looking at this as well...seems like somehow there wasn't a git conflict despite both https://github.com/algorand/indexer/commit/92991d8db9282c14a66466eae16bf9185048f21e and https://github.com/algorand/indexer/commit/c755f42bf3b8ac7f0d3f48d0412b4ff3f3dc4636 modifying the same file.

shiqizng (approved):

PR #1452, opened 2023-01-31
algod importer: Combines algod and algod_follower

### LINKED ISSUE #1449: Combine algod-based plugin logic
Combine the algod and algod_follower plugins.  Because the two plugins share similar (but not the exact) logic, it makes sense to use the same plugin (differentiating behavior with a config switch) so that changes to one are not lost to the other.

AlgoStephenAkiki:
Resolves #1449

Combines the algod and algod_follower plugins and uses a config value to switch between the logic of them


Eric-Warehime (approved):

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -155,23 +155,18 @@ func (algodImp *algodImporter) GetBlock(rnd uint64) (data.BlockData, error) {
 	var blk data.BlockData
 
 	for retries := 0; retries < 3; retries++ {
-		// nextRound - 1 because the endpoint waits until the subsequent block is committed to return
```
shiqizng, line 123: I'm wondering if this retry loop is necessary. there's already retries at the pipeline level. in this loop there's no wait before the next try, could this loop always finish before the next block is available?

Eric-Warehime: Yep, just added retires w/ backoff to the most recent version.
```
for r := 0; r < retries; r++ {
		time.Sleep(time.Duration(waitMultiplier*float64(r)) * initialWait)
...
```

We could just rely on the pipeline-level retires like you said, but retries with backoff make more sense in the case of the follower node since the catchup service gets started/stopped so frequently.

winder (approved):

PR #1460, opened 2023-02-08
FilterProcessor performance improvement and minor refactoring.

winder:
## Summary

While looking into the filter processor I found a lot of opportunities to simplify the code and make it faster.

This PR is an initial pass has a few small changes:
* improve performance by passing a pointer instead of copying the transaction object.
* improve performance by using Sprintf("%s") instead of reflection.
* simplify code by handling the "MicroAlgo" type in the generator instead of at runtime.
* minor refactoring to unit tests. There are many integration style tests which could be converted to unit tests.

Runtime of the search function went from 1500ns to 575ns according to the benchmark.

## Test Plan

No functional changes, existing unit tests check for regressions in the refactoring.

### DIFF
--- a/conduit/plugins/processors/filterprocessor/fields/filter.go
+++ b/conduit/plugins/processors/filterprocessor/fields/filter.go
@@ -39,7 +39,7 @@ func (f Filter) SearchAndFilter(input data.BlockData) (data.BlockData, error) {
 
 			allFalse := true
 			for _, fs := range f.Searchers {
-				b, err := fs.search(txn)
+				b, err := fs.search(&txn)
 				if err != nil {
 					return data.BlockData{}, err
 				}
@@ -59,7 +59,7 @@ func (f Filter) SearchAndFilter(input data.BlockData) (data.BlockData, error) {
 	case anyFieldOperation:
 		for _, txn := range input.Payset {
 			for _, fs := range f.Searchers {
-				b, err := fs.search(txn)
+				b, err := fs.search(&txn)
 				if err != nil {
 					return data.BlockData{}, err
 				}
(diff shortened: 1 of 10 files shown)


conduit/plugins/processors/filterprocessor/fields/searcher.go
```diff
@@ -18,28 +18,36 @@ type Searcher struct {
...
-	if f.MethodToCall != "" {
-		// If there is a function, search what is returned
-		toSearch = e.MethodByName(f.MethodToCall).Call([]reflect.Value{})[0].Interface()
-	} else {
-		// Otherwise, get the original value
-		toSearch = e.Interface()
-	}
-
-	b, err := (*f.Exp).Search(toSearch)
+	b, err := (*f.Exp).Search(val)
```
winder, line 50: This degree of abstraction wasn't necessary. The only "MethodToCall" is "String", the only place it's used is in the `regexExpression`. So I replaced it with `fmt.Sprintf("%s", input)` in the `Search` function. This yielded a performance improvement as well.

conduit/plugins/processors/filterprocessor/fields/filter.go
```diff
@@ -39,7 +39,7 @@ func (f Filter) SearchAndFilter(input data.BlockData) (data.BlockData, error) {
 
 			allFalse := true
 			for _, fs := range f.Searchers {
-				b, err := fs.search(txn)
+				b, err := fs.search(&txn)
```
winder, line 42: passing a reference here was a ~1.5x performance improvement.

winder:
I'll open a new PR

PR #1463, opened 2023-02-13
filter-processor: move reflection into generated code

winder:
## Summary

These changes focus on updating the generated `SignedTxnFunc` function.
* Renamed to `LookupFieldByTag`.
* Cast return values to `uint64`, `int64` and `string`. 
* Do not allow sub-tags. i.e. previously `txn` was a valid tag, now only leaves are allowed such as `txn.fee`.
* Base64 encode "data" types like the note field, previously they were silently unsupported.
* Disallow unsupported types like slices (i.e. foreign accounts, app arguments) and maps (i.e. eval delta values).
* Minor refactoring outside the generate function (i.e. the `String()` function can no longer be called, because it happens inside the generated code).
* Adds initial support for switching to the SDK transaction type.

Followup:
* Because there are now only 3 types returned much of the code can be simplified.
* Stop using pointers in generated code. It is currently required because the reflect packages widely at runtime.
* Convert integration style tests to unit tests (i.e. there are no Searcher unit tests, they all begin with creating a processor).

## Test Plan

There are extensive unit tests to cover regressions.

### DIFF
--- a/conduit/plugins/processors/filterprocessor/fields/generated_signed_txn_map.go
+++ b/conduit/plugins/processors/filterprocessor/fields/generated_signed_txn_map.go
@@ -3,236 +3,211 @@
 package fields
 
 import (
+	"encoding/base64"
 	"fmt"
 
 	"github.com/algorand/go-algorand/data/transactions"
 )
 
-// SignedTxnFunc takes a tag and associated SignedTxnInBlock and returns the value
+// LookupFieldByTag takes a tag and associated SignedTxnInBlock and returns the value
 // referenced by the tag.  An error is returned if the tag does not exist
-func SignedTxnFunc(tag string, input *transactions.SignedTxnInBlock) (interface{}, error) {
-
+func LookupFieldByTag(tag string, input *transactions.SignedTxnInBlock) (interface{}, error) {
 	switch tag {
 	case "aca":
-		return &input.SignedTxnWithAD.ApplyData.AssetClosingAmount, nil
+		value := input.SignedTxnWithAD.ApplyData.AssetClosingAmount
+		return &value, nil
 	case "apid":
-		return &input.SignedTxnWithAD.ApplyData.ApplicationID, nil
+		value := uint64(input.SignedTxnWithAD.ApplyData.ApplicationID)
+		return &value, nil
 	case "ca":
-		return &input.SignedTxnWithAD.ApplyData.ClosingAmount, nil
+		value := uint64(input.SignedTxnWithAD.ApplyData.ClosingAmount.Raw)
+		return &value, nil
 	case "caid":
-		return &input.SignedTxnWithAD.ApplyData.ConfigAsset, nil
-	case "dt":
-		return &input.SignedTxnWithAD.ApplyData.EvalDelta, nil
-	case "dt.gd":
-		return &input.SignedTxnWithAD.ApplyData.EvalDelta.GlobalDelta, nil
-	case "dt.itx":
-		return &input.SignedTxnWithAD.ApplyData.EvalDelta.InnerTxns, nil
-	case "dt.ld":
-		return &input.SignedTxnWithAD.ApplyData.EvalDelta.LocalDeltas, nil
-	case "dt.lg":
-		return &input.SignedTxnWithAD.ApplyData.EvalDelta.Logs, nil
+		value := uint64(input.SignedTxnWithAD.ApplyData.ConfigAsset)
+		return &value, nil
 	case "hgh":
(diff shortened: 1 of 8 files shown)


conduit/plugins/processors/filterprocessor/gen/generate.go
```diff
@@ -35,25 +226,51 @@ func recursiveTagFields(theStruct interface{}, output map[string]string, tagLeve
...
+func LookupFieldByTag(tag string, input *transactions.SignedTxnInBlock) (interface{}, error) {
+	switch tag {
+{{ range .StructFields }}	case "{{ .TagPath }}":
+		value := {{ ReturnValue . "input" }}
+		return &value, nil
+{{ end }}	default:
+		return nil, fmt.Errorf("unknown tag: %s", tag)
+	}
+}
+`
```
winder, line 269: Now using the `text/template` package instead of string building.

winder:
Converting back into a draft. I want to make one of the error messages clearer.

conduit/plugins/processors/filterprocessor/gen/generate.go
```diff
@@ -1,31 +1,222 @@
...
+		return "int64"
+	case int8:
+		return "int64"
+	case int16:
+		return "int64"
+	case int32:
+		return "int64" //
+	// alias
+	// SDK types
+	case types.MicroAlgos:
```
shiqizng, line 101: nit: consider using renaming the alias to sdk since that's what I've been using in the other files.

conduit/plugins/processors/filterprocessor/gen/generate.go
```diff
@@ -1,31 +1,222 @@
...
+	case types.Round:
+		return "uint64"
+	case types.OnCompletion:
+		return "uint64"
+	case types.StateProofType:
+		return "uint64"
+	case types.TxType:
+		return "string"
+	// go-algorand types
+	case basics.AssetIndex:
```
shiqizng, line 117: do we need these types because we aren't using the sdk types in the importer?

Eric-Warehime (approved):
Would maybe like more unit tests here. Or maybe we should just be marking any `gen/` files as excluded since they're sort of tested by outputting the generated code itself?

conduit/plugins/processors/filterprocessor/fields/generated_signed_txn_map.go
```diff
@@ -3,236 +3,211 @@
...
-		return &input.SignedTxnWithAD.ApplyData.AssetClosingAmount, nil
+		value := input.SignedTxnWithAD.ApplyData.AssetClosingAmount
+		return &value, nil
 	case "apid":
-		return &input.SignedTxnWithAD.ApplyData.ApplicationID, nil
+		value := uint64(input.SignedTxnWithAD.ApplyData.ApplicationID)
+		return &value, nil
 	case "ca":
-		return &input.SignedTxnWithAD.ApplyData.ClosingAmount, nil
+		value := uint64(input.SignedTxnWithAD.ApplyData.ClosingAmount.Raw)
```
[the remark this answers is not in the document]
winder: This is still using go-algorand version. The SDK uses a simple uint64

conduit/plugins/processors/filterprocessor/fields/searcher.go
```diff
@@ -13,33 +13,23 @@ import (
 
 // Searcher searches the struct with an expression and method to call
 type Searcher struct {
-	Exp          *expression.Expression
-	Tag          string
-	MethodToCall string
+	Exp *expression.Expression
+	Tag string
```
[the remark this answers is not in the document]
winder: fixed in followup PR

conduit/plugins/processors/filterprocessor/gen/generate.go
winder: fixed in followup PR

conduit/plugins/processors/filterprocessor/gen/generate.go
winder: I'll remove the go-algorand types in a followup. I didn't want to change the behavior while refactoring.

conduit/plugins/processors/filterprocessor/gen/generate_test.go
```diff
@@ -0,0 +1 @@
+package main
```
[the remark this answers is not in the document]
winder: No. I was thinking about adding tests, I'll do that in the next PR.

shiqizng (approved):

PR #1468, opened 2023-02-15
filter-processor: Remove reflection

winder:
## Summary

Next round of changes after #1463

* merge `ExactFilter` and `EqualToFilter`
* (bugfix) support matching regex special characters in "exact" string search.
* searcher unit tests.
* use type casting instead of reflection.

## Test Plan

New unit tests and existing regression tests.


conduit/plugins/processors/filterprocessor/expression/expression.go
```diff
@@ -53,97 +47,74 @@ type regexExpression struct {
...
-
-			return &exp, nil
+func makeSignedExpression(searchStr string, expressionType FilterType) (Expression, error) {
+	if expressionType == RegexFilter {
+		return nil, fmt.Errorf("target type (numeric) does not support %s filters", expressionType)
+	}
+
+	v, err := strconv.ParseInt(searchStr, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("search string \"%s\" is not numeric: %w", searchStr, err)
```
shiqizng, line 78: nit: maybe the error message can be something like "not a valid int64", because the err could be an int overflow error?

conduit/plugins/processors/filterprocessor/expression/expression.go
```diff
@@ -53,97 +47,74 @@ type regexExpression struct {
...
+}
+
+func makeUnsignedExpression(searchStr string, expressionType FilterType) (Expression, error) {
+	if expressionType == RegexFilter {
+		return nil, fmt.Errorf("target type (numeric) does not support %s filters", expressionType)
+	}
+
+	v, err := strconv.ParseUint(searchStr, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("search string \"%s\" is not numeric: %w", searchStr, err)
```
shiqizng, line 94: similar to above

shiqizng (approved):
lgtm

conduit/plugins/processors/filterprocessor/expression/expression.go
winder: Thanks. Updated, and added tests for those two cases.

Eric-Warehime (approved):

Issue #1469, opened 2023-02-16
Invalid JSON written when using conduit file_writer


barnjamin:
### Subject of the issue

Using the conduit file_writer, JSON is produced with the integer value `0` as a key. This causes some programs to complain or fail altogether.

Conduit config:
```
hide-banner: true

log-level: "DEBUG"
log-file: "/tmp/conduit.log"

importer:
    name: algod
    config:
        netaddr: "http://localhost:4001"
        token: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

processors:
  - name: block_evaluator
    config:
        algod-addr: "http://localhost:4001"
        algod-token: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

exporter:
    name: file_writer 
    config: 
        block-dir: "/tmp/blocks"
        drop-certificate: true
```

![image](https://user-images.githubusercontent.com/520236/219380342-2ce02561-9a67-4d93-b906-b321b3403c69.png)

winder:
Fixing this should (hopefully) be a matter of switching to `EncodeStrict`, which is added here: [https://github.com/algorand/go-algorand-sdk/pull/467](https://github.com/algorand/go-algorand-sdk/pull/467/files#diff-3e8b8648fce88d6eecd3103f6612f362dc206ea4f95d92cc3acabae3fbe9feb0)

PR #1470, opened 2023-02-16
Conduit: add config setting to json encoder to write all keys as strings

barnjamin:
## Summary

addresses #1469 

Also changing the path for gitignore on conduit docs

## Test Plan

Tested locally and produces expected results 
image


winder:
This change has a side effect, it modifies the global encoder. The existing code also has a side effect where it converts the global encoder to pretty print output. These issues are addressed in another PR: #1488

PR #1473, opened 2023-02-17
conduit: update plugin interface types

### LINKED ISSUE #1438: Licensing - update conduit plugin interface types
Use SDK types for state delta and block in place of the go-algorand versions in the Conduit interfaces.

* Update interfaces.
* Update plugins to use new interfaces.

shiqizng:
resolves #1438


shiqizng:
I will rebase when #1468  is merged.

data/block_export_data.go
```diff
@@ -69,38 +67,46 @@ func (blkData *BlockData) UpdateFromEncodedBlockCertificate(input *rpcs.EncodedB
 
 // MakeBlockDataFromEncodedBlockCertificate makes BlockData from rpcs.EncodedBlockCert
 func MakeBlockDataFromEncodedBlockCertificate(input *rpcs.EncodedBlockCert) BlockData {
```
Eric-Warehime, line 69: Can we get rid of this now? What else needs the encoded block cert converted after this PR?

Eric-Warehime (approved):

data/block_export_data.go
shiqizng: block_processor still uses this. we can remove it when removing the block_processor which is one of the last steps.

PR #1480, opened 2023-02-21
daemon: Remove bp from indexer daemon

Eric-Warehime:
## Summary
Removes the block_processor from the indexer daemon. Sets the algod_importer to run in follower mode.

This will make indexer backwards incompatible w/ versions of algod prior to the upcoming release, and will also remove the ability to run indexer against an algod archival node which is not also running w/ `EnableFollowMode` synced w/ indexer.

Relies on the changes in #1478 to be merged.

## Test Plan

Existing e2e tests. Also, regression testing prior to release.

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -21,7 +21,6 @@ import (
 	"github.com/algorand/indexer/conduit/pipeline"
 	_ "github.com/algorand/indexer/conduit/plugins/exporters/postgresql"
 	_ "github.com/algorand/indexer/conduit/plugins/importers/algod"
-	_ "github.com/algorand/indexer/conduit/plugins/processors/blockprocessor"
 	"github.com/algorand/indexer/config"
 	"github.com/algorand/indexer/fetcher"
 	"github.com/algorand/indexer/idb"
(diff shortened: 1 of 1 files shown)


conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -154,6 +155,18 @@ func (algodImp *algodImporter) Close() error {
 	return nil
 }
 
+func (algodImp *algodImporter) getDelta(rnd uint64) (sdk.LedgerStateDelta, error) {
+	var delta sdk.LedgerStateDelta
+	params := algod.GetLedgerStateDeltaParams{Format: "msgp"}
+	data, err := (*common.Client)(algodImp.aclient).GetRaw(algodImp.ctx, fmt.Sprintf("/v2/deltas/%d", rnd), params, nil)
```
winder, line 161: We'll need to avoid this type after the SDK is updated. Not 100% sure how the requests are made, maybe this works?
```suggestion
	data, err := (*common.Client)(algodImp.aclient).GetRaw(algodImp.ctx, fmt.Sprintf("/v2/deltas/%d?format=msgp", rnd), params, nil)
```

Eric-Warehime: I guess I'll wait until the sdk changes are done and then update.

winder (approved):
LGTM

shiqizng (approved):

PR #1485, opened 2023-02-23
importer: rewrite make integration 

shiqizng:
rewrite make integration tests as handler tests and remove importer code and reference.


api/handlers_e2e_test.go
```diff
@@ -2129,3 +2129,447 @@ func runBoxCreateMutateDelete(t *testing.T, comparator boxTestComparator) {
...
+	}
+	resp, err := http.Get(req)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	return resp, body
+}
+
+// recreate tests from  common.sh/create_delete_tests()
```
shiqizng, line 2148: these tests are replicating those in common.sh/create_delete_tests(). Given how tedious it is to create the vb files, not all tests are replicated. Let me know if you see cases that should be covered.

Eric-Warehime (approved):
The one test which might be useful for backwards compatibility reasons is https://github.com/algorand/indexer/blob/develop/test/common.sh#L409

PR #1486, opened 2023-02-24
bug: Update default indexer + algod mode.

winder:
## Summary

Fix Indexer's algod importer configuration. This would need to change again soon once we remove the block processor.

## Test Plan

Existing CI


shiqizng (commented):
shouldn't it have been default to archival mode? 
https://github.com/algorand/indexer/blob/develop/conduit/plugins/importers/algod/algod_importer.go#L96

winder:
> shouldn't it have been default to archival mode? https://github.com/algorand/indexer/blob/develop/conduit/plugins/importers/algod/algod_importer.go#L96

You're right, I missed that.

PR #1487, opened 2023-02-24
filter-plugin: Inner Txn support.

winder:
## Summary

1. Generator switched to use `SignedTxnWithAD` instead of `SignedTxnInBlock`.
2. Split out a `matches` function from SearchAndFilter to avoid duplicate code for operation types (any/all/none).
3. Rename `Search` to `Match` for lowest level expressions.
4. Rename `FilterType` to `Type`.
5. Remove `Filter` from the `Filter` constants, it is in their type.
6. There is one remaining `search` function in the `Searcher` type. It now has an option to recursively search inner transactions for a match.

## Test Plan

New unit test.

Updated benchmarks.

### DIFF
--- a/conduit/plugins/processors/filterprocessor/expression/expression.go
+++ b/conduit/plugins/processors/filterprocessor/expression/expression.go
@@ -101,18 +101,22 @@ func makeUnsignedExpression(searchStr string, expressionType Type) (Expression, 
 }
 
 // MakeExpression creates an expression based on an expression type
-func MakeExpression(filterType Type, expressionSearchStr string, target interface{}) (exp Expression, err error) {
+func MakeExpression(expressionType Type, expressionSearchStr string, target interface{}) (exp Expression, err error) {
+	if _, ok := TypeMap[expressionType]; !ok {
+		return nil, fmt.Errorf("expression type (%s) is not supported", expressionType)
+	}
+
 	switch t := target.(type) {
 	case uint64:
-		return makeUnsignedExpression(expressionSearchStr, filterType)
+		return makeUnsignedExpression(expressionSearchStr, expressionType)
 	case int64:
-		return makeSignedExpression(expressionSearchStr, filterType)
+		return makeSignedExpression(expressionSearchStr, expressionType)
 	case string:
-		if filterType == EqualTo {
+		if expressionType == EqualTo {
 			// Equal to for strings is a special case of the regex pattern.
 			expressionSearchStr = fmt.Sprintf("^%s$", regexp.QuoteMeta(expressionSearchStr))
 		}
-		return makeRegexExpression(expressionSearchStr, filterType)
+		return makeRegexExpression(expressionSearchStr, expressionType)
 
 	default:
 		return nil, fmt.Errorf("unknown expression type: %T", t)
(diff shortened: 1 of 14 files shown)


conduit/plugins/processors/filterprocessor/expression/expression.go
```diff
@@ -101,18 +114,21 @@ func makeUnsignedExpression(searchStr string, expressionType FilterType) (Expres
...
 	case int64:
-		return makeSignedExpression(expressionSearchStr, filterType)
+		return makeSignedExpression(expressionSearchStr, expressionType)
 	case string:
-		if filterType == EqualToFilter {
-			// Equal to for strings is a special case of the regex pattern.
-			expressionSearchStr = fmt.Sprintf("^%s$", regexp.QuoteMeta(expressionSearchStr))
+		if expressionType == EqualTo {
+			return &stringEqualExpression{Str: expressionSearchStr}, nil
 		}
```
winder, line 130: Switching from a special regex expression to string equality was an 8-12% performance increase according to the benchmark.

winder:
Benchmark results
```
BenchmarkProcess/inner_txn_count_0
BenchmarkProcess/inner_txn_count_0-8         	 1274750	       920.6 ns/op
BenchmarkProcess/inner_txn_count_10
BenchmarkProcess/inner_txn_count_10-8        	  219909	      5759 ns/op
BenchmarkProcess/inner_txn_count_100
BenchmarkProcess/inner_txn_count_100-8       	   25620	     48310 ns/op
```

conduit/plugins/processors/filterprocessor/expression/expression.go
```diff
@@ -56,10 +56,20 @@ func (e *regexExpression) Search(input interface{}) (bool, error) {
...
+}
+
+func (e *stringEqualExpression) Match(input interface{}) (bool, error) {
+	switch v := input.(type) {
+	case string:
+		return e.Str == v, nil
+	default:
+		return false, fmt.Errorf("unexpected regex search input type (%T)", v)
 	}
+}
```
winder, line 70: Previously this was using a special regex expression. In benchmarks using string equals was 8-12% faster.

conduit/plugins/processors/filterprocessor/gen/generate.go
```diff
@@ -296,7 +296,7 @@ func main() {
 	}
 
 	// Process fields.
-	fields, errors := getFields(sdk.SignedTxnInBlock{}, ignoreTags)
+	fields, errors := getFields(sdk.SignedTxnWithAD{}, ignoreTags)
```
winder, line 299: Switched to `SignedTxnWithAD` because that is the type used for inner transactions. This way the same lookup function is used for the root transaction and the inner transactions.

shiqizng (approved):
lgtm

conduit/plugins/processors/filterprocessor/fields/searcher_test.go
```diff
@@ -5,89 +5,142 @@ import (
...
+		require.NoError(t, err)
+		assert.Equal(t, matches, true)
+
+		// Provide the root txn, no match at the root.
+		// It should have no match with searchInner: false.
+		matches, err = searcher.search(&txnWithInnerMatch)
+
+		// No match on inner txn
+		require.NoError(t, err)
+		assert.Equal(t, matches, true)
```
Eric-Warehime, line 144: Comments seem wrong? It should match on inner here right?

Eric-Warehime (approved):

conduit/plugins/processors/filterprocessor/fields/searcher_test.go
winder: Thanks, fixed.

PR #1489, opened 2023-03-01
conduit: Reorganize ahead of repo split.

winder:
## Summary

Moving functions and files around to make it easier to extract Conduit into a separate repo.

Guiding theory:
1. The conduit package will become the project root, so it should not import packages from outside of conduit.
2. Utilities that are only used by a single package can be part of that package.

Not in this PR:
The postgresql package still depends on idb and other things. That change is much larger than what I've started with so I'll handle it in another PR.

## Test Plan

Nonfunctional changes. Existing regression tests should pass.


Eric-Warehime (approved):

conduit/plugins/exporters/filewriter/util.go
```diff
@@ -0,0 +1,88 @@
...
+		writer = gz
+	} else {
+		writer = file
+	}
+
+	var handle *codec.JsonHandle
+	if pretty {
+		handle = prettyHandle
+	} else {
+		handle = protocol.JSONStrictHandle
```
shiqizng, line 55: this will need to be updated after #1492 is merge.

shiqizng (approved):

PR #1493, opened 2023-03-03
e2e-tests: Filter processor e2e test

Eric-Warehime:
## Summary
Adds a couple of filter scenarios to the conduit e2e tests.  
Also adds details on creating new e2e tests to the readme.  

For now I have not added indexer queries to these. They simply verify that the pipeline runs without error.

## Test Plan

`make e2e-conduit`


winder (approved):

e2e_tests/src/e2e_conduit/scenarios/filter_scenario.py
```diff
@@ -0,0 +1,40 @@
...
+                {"any": [
+                    {"tag": "txn.type",
+                     "expression-type": "equal",
+                     "expression": "pay"
+                    }
+                ]}
+            ]),
+        ],
+        exporter=exporters.PostgresqlExporter(),
+    )
```
shiqizng, line 40: AlgodImporter would also work for these filter tests, right?

Eric-Warehime: The `AlgodImporter` fixture is separate still since the plugins were originally two separate ones instead of one w/ a config option. In this case we still need to use the follower because the pgsql exporter will fail w/o the deltas.

shiqizng (approved):

Issue #1482, opened 2023-02-22
Conduit: filter transactions at the group level 


algoanne:
The filter processor currently allows filtering on any of a transaction's properties. For example, filter all transactions that are app calls with a certain app ID. 

This is great. However, many dapps today are designed to process transactions as transaction groups. Thus, a dapp would be interested not only in the app calls to its app ID, but also any other transactions that were the group with that app call. 

In other words, imagine that we have the following transaction group: 

- txn 1: axfer of USDC to anne.algo 
- txn 2: pay of Algo from anne.algo
- txn 2: app call to AMM 

If the AMM developer wants to report all activity for their dapp, they need to get all of these transactions through their filter processor. 

Important note: group transactions can be both at the top-level or inner transactions. The AMM txn group could also be organized as:
- txn 1: app call to AMM, which makes 2 inner transactions: 
    - txn 2: axfer of USDC to anne.algo
    - txn 3: pay of Algo from anne.algo 
    
possible prereq/related: #1329

algoanne:
**configuration options**
- include-group-transactions: if on, for any txn matched, we'll also get that txn's grouped txns. If off, just get the individual txns matched. 
- include-inner-transactions: if on, for any txn matched, we'll also get that txn's inners. If off, just get the individual txns matched. 

if matched on an inner, automatically include its top-level transaction (what if it's an inner's inner?). This case will need to be well documented because it is non-intuitive (the better design would be to have this also be a config, but it's difficult technically at this point). 

default both of the configuration options to **on**.

algoanne:
as part of this work, re-run filter processor performance test.

PR #1501, opened 2023-03-09
new-feature: filter transactions at the group level  

shiqizng:
#1482 
this PR adds a new filter flag `omit-group-transactions`. filter processor would return the matched txn and its grouped txns by default. when `omit-group-transactions` is set to false, only matched txns are returned. 

## Test
new unit tests


shiqizng:
move this PR to conduit

PR #1506, opened 2023-03-14
idb(postgres): add protocol version check before writing to database.

winder:
## Summary

Make sure the block being written is for a known protocol version.


idb/postgres/postgres.go
```diff
@@ -175,6 +175,12 @@ func (db *IndexerDb) init(opts idb.IndexerDbOptions) (chan struct{}, error) {
 
 // AddBlock is part of idb.IndexerDb.
 func (db *IndexerDb) AddBlock(vb *itypes.ValidatedBlock) error {
+	protoVersion := protocol.ConsensusVersion(vb.Block.CurrentProtocol)
+	_, ok := config.Consensus[protoVersion]
```
shiqizng, line 179: this reminds me of `EnableAssetCloseAmount`. do we still need to support `AssetCloseAmount` for the older txns? maybe it's one of the things we can deprecate in the next release?

PR #1517, opened 2023-04-13
Enhancement: add boxes to the indexer / algod validator

bbroder-algo:
This PR compares box names and contents between algod and indexer to ensure that the data between the two is synchronized.  Validator receives a list of app-id and box names to check from a running indexer postgres DB and uses this input to formulate HTTP requests to both the indexer and the algod node for the box contents, which it then compares for equality. 

The stdin line input format for validator has changed to support this functionality, so that now each line can either be:

1: an address as before (either b64 or not)
2: "address,(b64 or not encoded address)"
3: "box,(appid),(b64 encoded box name)"

The first option ensures backwards compatibility for the tool.  The next two allow some disambiguation between comparing accounts vs boxes. 

Also included is a generate_boxes.sh which will generate the box lines for validator. 

The tool was run and no box differences between mainnet indexer and mainnet algod were found.


cmd/validator/core/command.go
```diff
@@ -45,7 +45,7 @@ func init() {
 	ValidatorCmd.Flags().StringVar(&addr, "addr", "", "If provided validate a single address instead of reading Stdin.")
 	ValidatorCmd.Flags().IntVar(&threads, "threads", 4, "Number of worker threads to initialize.")
 	ValidatorCmd.Flags().IntVar(&processorNum, "processor", 0, "Choose compare algorithm [0 = Struct, 1 = Reflection]")
-	ValidatorCmd.Flags().BoolVar(&printCurl, "print-commands", false, "Print curl commands, including tokens, to query algod and indexer.")
+	ValidatorCmd.Flags().BoolVar(&printCurl, "print-commands", false, "Print curl commands, including tokens, to query algod and indexer accounts.")
```
bbroder-algo, line 48:
```suggestion
	ValidatorCmd.Flags().BoolVar(&printCurl, "print-commands", false, "Print curl commands, including tokens, to query algod and indexer contents.")
```

bbroder-algo:
why are all my tabs/spaces wrong.

cmd/validator/core/validator.go
```diff
@@ -129,6 +222,9 @@ func CallProcessor(processor Processor, addrInput string, config Params, results
 	algodDataURL := fmt.Sprintf("%s/v2/accounts/%s", config.AlgodURL, addr)
 	indexerDataURL := fmt.Sprintf("%s/v2/accounts/%s", config.IndexerURL, addr)
 
+    fmt.Println("algodDataURL: ", algodDataURL)
+    fmt.Println("indexerDataURL: ", indexerDataURL)
+
```
bbroder-algo, line 227:
```suggestion
```

cmd/validator/core/validator.go
```diff
@@ -241,6 +333,35 @@ func getData(url, token string) ([]byte, error) {
 	return data, ioErr
 }
 
+func resultResourceError(err error, resource string) Result {
+	return Result{
+		Equal:      false,
+		Error:      err,
+		SkipReason: NotSkipped,
```
bbroder-algo, line 340:
```suggestion
```

cmd/validator/core/validator.go
```diff
@@ -119,7 +123,95 @@ func Start(work <-chan string, processorID ProcessorID, threads int, config Para
...
+
+	// Retry loop.
+	for i := 0; true; i++ {
+		indexerData, err := getData(indexerDataURL, config.IndexerToken)
+		if err != nil {
+			err = fmt.Errorf("error getting indexer data (%s): %w", indexerData, err)
+			switch {
+			case strings.Contains(string(indexerData), api.ErrResultLimitReached):
+				results <- resultBoxSkip(err, appid, boxname, SkipLimitReached)
+			case strings.Contains(string(indexerData), api.ErrNoAccountsFound):
```
bbroder-algo, line 173: maybe different errors come out of the applications indexer endpoint that would tell us about a box name or app id not found.

cmd/validator/core/validator.go
```diff
@@ -119,7 +129,101 @@ func Start(work <-chan string, processorID ProcessorID, threads int, config Para
 }
 
 // CallProcessor invokes the processor with a retry mechanism.
```
winder, line 131:
```suggestion
// CallProcessor invokes the processor with a retry mechanism.
// Accepted input formats:
// 1: "(b64 or not encoded address)"
// 2: "address,(b64 or not encoded address)"
// 3: "box,(appid),(b64 encoded box name)"
```

cmd/validator/core/validator.go
```diff
@@ -119,7 +129,101 @@ func Start(work <-chan string, processorID ProcessorID, threads int, config Para
...
+			results <- resultResourceError(err, input)
+			return
+		}
+
+	}
+}
+
+func callProcessorBox(processor Processor, appid string, boxname string, config Params, results chan<- Result) {
+	boxname = strings.ReplaceAll(boxname, "+", "%2B")
+	boxname = strings.ReplaceAll(boxname, "/", "%2F")
```
winder, line 152: This library is available, not sure if it's relevant to what you have here:
https://github.com/algorand/avm-abi/blob/main/apps/box.go

cmd/validator/core/validator.go
```diff
@@ -119,7 +129,101 @@ func Start(work <-chan string, processorID ProcessorID, threads int, config Para
...
+			// Return when results are equal, or when finished retrying.
+			result.Retries = i
+			if result.Details != nil {
+				result.Details.Resource = "box"
+				result.Details.Appid = appid
+				result.Details.Boxname = boxname
+			}
+			results <- result
+			return
+		}
```
winder, line 219: It's a shame we can't get the rounds too. The vast majority of failures I've seen have mismatched rounds, so it's a quick way to filter out false negatives. We'll have to monitor whether or not this ends up being a problem for Boxes, or if it's a 1 in a million thing that we run into because there are millions of accounts.

Thinking out loud: if it does become a problem, and we don't want to add round to the boxes results, this code could run the account processor to get some extra information.

cmd/validator/core/validator.go
```diff
@@ -241,6 +345,34 @@ func getData(url, token string) ([]byte, error) {
...
+		Equal:      false,
+		Error:      err,
+		SkipReason: skip,
+		Retries:    0,
+		Details: &ErrorDetails{
+			Resource: "box",
+			Appid:    appid,
+			Boxname:  boxname,
+		},
+	}
```
winder, line 373: nit: this isn't as much of an improvement as I thought it might be. Take it or leave it
```suggestion
	result := resultResourceError(err, "box")
	result.SkipReason = skip
	result.Details.Appid = appid,
	result.Details.Boxname = boxname,
	return result
```

cmd/validator/core/command.go
```diff
@@ -45,7 +45,7 @@ func init() {
 	ValidatorCmd.Flags().StringVar(&addr, "addr", "", "If provided validate a single address instead of reading Stdin.")
```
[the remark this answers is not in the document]
winder: A resource override might be nice too, so that the shell script to process/pipe an error log back into the validator doesn't need to remember the format.

I wouldn't consider either of these required for this PR, someone can improve the tool later if/when it becomes a problem.

cmd/validator/core/validator.go
bbroder-algo: that library makes key names for the KV store but does not prepare a B64 for URL encoding to the endpoints.

cmd/validator/core/validator.go
bbroder-algo: yes we could either add the round to the box endpoint or we could sandwich each call to box with a request for the round number, and repeat if any of the four rounds are different.

cmd/validator/core/validator.go
bbroder-algo: replaced with net/url

cmd/validator/core/validator.go
```diff
@@ -119,7 +129,101 @@ func Start(work <-chan string, processorID ProcessorID, threads int, config Para
...
+			err = fmt.Errorf("error getting indexer data (%s): %w", indexerData, err)
+			switch {
+			case strings.Contains(string(indexerData), api.ErrFailedLookingUpBoxes):
+				results <- resultBoxSkip(err, appid, boxname, SkipBoxFailedLookup)
+			case strings.Contains(string(indexerData), api.ErrWrongAppidFound):
+				results <- resultBoxSkip(err, appid, boxname, SkipBoxWrongAppid)
+			case strings.Contains(string(indexerData), api.ErrMultipleBoxes):
+				results <- resultBoxSkip(err, appid, boxname, SkipBoxMultiple)
+			case strings.Contains(string(indexerData), api.ErrNoBoxesFound):
+				results <- resultBoxSkip(err, appid, boxname, SkipBoxNoBoxes)
```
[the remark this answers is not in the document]
bbroder-algo: escalated to error

cmd/validator/core/command.go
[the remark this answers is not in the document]
bbroder-algo: I put the box command in and will leave the resource override for another conversation.

cmd/validator/core/command.go
```diff
@@ -43,14 +44,15 @@ func init() {
 	ValidatorCmd.Flags().IntVarP(&config.Retries, "retries", "", 5, "Number of retry attempts when a difference is detected.")
 	ValidatorCmd.Flags().IntVarP(&config.RetryDelayMS, "retry-delay", "", 1000, "Time in milliseconds to sleep between retries.")
 	ValidatorCmd.Flags().StringVar(&addr, "addr", "", "If provided validate a single address instead of reading Stdin.")
+	ValidatorCmd.Flags().StringVar(&box, "box", "", "If provided validate a single (appid,b64boxname) instead of reading Stdin.")
```
bbroder-algo, line 47:
```suggestion
	ValidatorCmd.Flags().StringVar(&box, "box", "", "If provided validate a single box (in the format "appid,b64boxname") instead of reading Stdin.")
```

cmd/validator/core/validator.go
bbroder-algo: scytale of the opinion that we should add the round number to the box endpoint and also further add a requested round parameter to the endpoint.

winder (approved):

Issue #1432, opened 2023-01-22
validator: support validating box data.


winder:
## Problem

[Validator](https://github.com/algorand/indexer/tree/develop/cmd/validator) is used to confirm that the postgres data is consistent with data returned by go-algorand.

It should be updated to validate box data along with the account data. 

## Urgency

High. If there is an inconsistency with production data we would not currently be able to detect it.

bbroder-algo:
https://github.com/algorand/indexer/pull/1517

bbroder-algo:
Validator logic merged into indexer, however
1: requested --resource capability for redirects back into validator
2: round response incorporated into algod's box endpoint
3: round REQUEST incorporated into algod's box endpoint

bbroder-algo:
2: https://github.com/algorand/go-algorand/pull/5340

bbroder-algo:
checking the round response in the validator #1564

gmalouf:
Completed with #1517

PR #1519, opened 2023-04-28
REST: add round information to box response

bbroder-algo:
# Summary

Provide for the round information relevant to the box contents in the indexer box endpoint

## Test Plan

Altered saved endpoint response test resources for boxes to include the round number.


bbroder-algo:
I wasn't sure specifically how to generate the fixture response json from the readme so I edited the test_responses by hand

algochoi (commented):
Looks reasonable to me - were there specific use-cases for this? It could be pretty interesting to track box value changes using follower mode at different rounds, and this new field could make that more obvious.

bbroder-algo:
v2/account gives you the round number in the response and validator uses this to see if results from algod and indexer are aligned on the same round.

api/handlers.go
```diff
@@ -1104,6 +1104,7 @@ func (si *ServerImplementation) fetchApplicationBoxes(ctx context.Context, param
 			if appid == 0 {
 				appid = generated.ApplicationId(result.App)
 			}
+			result.Box.Round = uint64(round)
```
tzaffi, line 1107:
```suggestion
			result.Box.Round = round
```
`round` is already a `uint64`

algochoi (approved):
Makes sense to me! RE: https://github.com/algorand/indexer/issues/1432

api/handlers.go
bbroder-algo: [b17fbd7](https://github.com/algorand/indexer/pull/1519/commits/b17fbd75009c974d09f73404c6f3038e9cf1ed85)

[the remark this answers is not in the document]
bbroder-algo: [b17fbd7](https://github.com/algorand/indexer/pull/1519/commits/b17fbd75009c974d09f73404c6f3038e9cf1ed85)

Eric-Warehime (approved):
LGTM. Yes we need to regenerate the SDKs using the updated oapi json in order to get the new field in the responses.

PR #1530, opened 2023-05-23
encoding: Fix partupdrmv encoding.

winder:
## Summary

The go-algorand based version of Indexer missed the `partupdrmv` block header field. When switching back to the go-SDK this has caused a scenario where the old and new versions of Indexer database are no longer compatible. This fix makes the decoding more lenient and fixes the encoding to match the previous version.

In addition, this PR removes redundant overrides.

See commit 1 for removing redundant overrides.
See commit 2 for the `partupdrmv` override.

## Test Plan

Update unit tests.

### DIFF
--- a/idb/postgres/internal/encoding/encoding.go
+++ b/idb/postgres/internal/encoding/encoding.go
@@ -70,6 +70,9 @@ func (addr *AlgodEncodedAddress) UnmarshalText(text []byte) error {
 }
 
 func convertExpiredAccounts(accounts []sdk.Address) []AlgodEncodedAddress {
+	if len(accounts) == 0 {
+		return nil
+	}
 	res := make([]AlgodEncodedAddress, len(accounts))
 	for i, addr := range accounts {
 		res[i] = AlgodEncodedAddress(addr)
@@ -78,6 +81,9 @@ func convertExpiredAccounts(accounts []sdk.Address) []AlgodEncodedAddress {
 }
 
 func unconvertExpiredAccounts(accounts []AlgodEncodedAddress) []sdk.Address {
+	if len(accounts) == 0 {
+		return nil
+	}
 	res := make([]sdk.Address, len(accounts))
 	for i, addr := range accounts {
 		res[i] = sdk.Address(addr)
--- a/idb/postgres/internal/encoding/encoding.go
+++ b/idb/postgres/internal/encoding/encoding.go
@@ -52,18 +52,12 @@ func decodeBase64(data string) ([]byte, error) {
 
 func convertBlockHeader(header sdk.BlockHeader) blockHeader {
 	return blockHeader{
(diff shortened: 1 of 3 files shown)


idb/postgres/internal/encoding/encoding.go
```diff
@@ -52,18 +52,12 @@ func decodeBase64(data string) ([]byte, error) {
...
+		BlockHeader: header,
 	}
 }
 
 func unconvertBlockHeader(header blockHeader) sdk.BlockHeader {
 	res := header.BlockHeader
-	res.Branch = sdk.BlockHash(header.BranchOverride)
-	res.FeeSink = sdk.Address(header.FeeSinkOverride)
-	res.RewardsPool = sdk.Address(header.RewardsPoolOverride)
 	return res
```
shiqizng, line 104: nit: `return header.BlockHeader`

idb/postgres/internal/encoding/encoding.go
```diff
@@ -333,17 +277,17 @@ func unconvertEvalDelta(delta evalDelta) sdk.EvalDelta {
 
 func convertSignedTxnWithAD(stxn sdk.SignedTxnWithAD) signedTxnWithAD {
 	return signedTxnWithAD{
-		SignedTxnWithAD:   stxn,
-		TxnOverride:       convertTransaction(stxn.Txn),
-		AuthAddrOverride:  sdk.Digest(stxn.AuthAddr),
+		SignedTxnWithAD: stxn,
+		TxnOverride:     convertTransaction(stxn.Txn),
+		//AuthAddrOverride:  sdk.Digest(stxn.AuthAddr),
```
shiqizng, line 325: remove

shiqizng (commented):
Minor comments. Looks good.

idb/postgres/internal/encoding/types.go
```diff
@@ -1,41 +1,27 @@
...
 
+// AlgodEncodedAddress is an address encoded in the format used by algod.
+type AlgodEncodedAddress sdk.Address
+
 type blockHeader struct {
 	sdk.BlockHeader
-	BranchOverride      sdk.Digest `codec:"prev"`
-	FeeSinkOverride     sdk.Digest `codec:"fees"`
-	RewardsPoolOverride sdk.Digest `codec:"rwd"`
+	ExpiredParticipationAccountsOverride []AlgodEncodedAddress `codec:"partupdrmv"`
```
shiqizng, line 12: why only this Address field need to be AlgodEncodedAddress but the others can be sdk.Address?

algochoi (approved):
Nice, fewer type conversions and consolidating to b64 seem like a win to me. Test looks good 👍 

I'm less familiar with indexer encoding/decoding, so probably worth getting another approval

PR #1531, opened 2023-05-24
encoding: Better support for address encodings.

winder:
## Summary

There is a bug in the newer version of Indexer used by Conduit which caused addresses to be encoded differently. This PR makes the decoding more lenient to allow for the different encodings.

## Test Plan

New unit test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
       go_version:
         type: string
     environment:
-      CI_E2E_FILENAME: "faab6dcf/rel-nightly"
+      CI_E2E_FILENAME: "rel-nightly"
     steps:
       - go/install:
           version: << parameters.go_version >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
       go_version:
         type: string
     environment:
-      CI_E2E_FILENAME: "rel-nightly"
+      CI_E2E_FILENAME: "f99e7b0c/rel-nightly"
     steps:
       - go/install:
           version: << parameters.go_version >>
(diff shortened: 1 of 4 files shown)


algochoi (commented):
Changes look fine but do we need to cherry-pick fixes to this hotfix branch everytime? I'm not familiar with this dual branch system.

shiqizng (commented):
this looks good. I'm starting an indexer deployment to verify that it works as expected.

winder:
> Changes look fine but do we need to cherry-pick fixes to this hotfix branch everytime? I'm not familiar with this dual branch system.

@algochoi This is for the next release. It's a [gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) thing, but also a little strange because of the licensing work currently staged in `develop`.

winder:
Confirmed regression fix with mainnet data. 

Scenario: Initialize postgres DB using Conduit master branch and a mainnet follower node.

Indexer running the current tip of master: 1c79c90ed66c3455f22ea26a903f0ee589dc6e24 reproduces the encoding error 
```command
$ curl localhost:10000/v2/blocks/20215865?pretty
{
  "message": "error while looking up block for round '20215865': json decode error [pos 641]: failed to decode address 2GpQYYQK6VSxGFumcF/wZ4OLnJDJTASMIG5VL/AyU4k= to base 32"
}
```

This PR applied to the `hotfix/2.15.4` branch: 8d6161b04338bb6c922cb4bbfb472a0785010697 is able to fetch the data.
```bash
$ curl -s localhost:10000/v2/blocks/20215865?pretty|grep part
  "participation-updates": {
    "expired-participation-accounts": [

```

PR #1532, opened 2023-05-24
tests: Fix tests with app and asset ID less than 1000.

winder:
## Summary

Fix integration test errors associated with assetid's less than 1000.

## Test Plan

These are tests.

### DIFF
--- a/idb/postgres/internal/writer/writer_test.go
+++ b/idb/postgres/internal/writer/writer_test.go
@@ -1394,9 +1394,10 @@ func TestWriterAddBlockInnerTxnsAssetCreate(t *testing.T) {
 	require.NotContains(t, txns[3].txn, "itx", "The inner transactions should be pruned.")
 
 	// Verify correct App and Asset IDs
-	require.Equal(t, 1, txns[0].asset, "intra == 0 -> ApplicationID = 1")
+	require.Equal(t, 1001, txns[0].asset, "intra == 0 -> ApplicationID = 1")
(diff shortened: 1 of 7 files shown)


idb/postgres/internal/writer/writer_test.go
```diff
@@ -1394,9 +1394,10 @@ func TestWriterAddBlockInnerTxnsAssetCreate(t *testing.T) {
 	require.NotContains(t, txns[3].txn, "itx", "The inner transactions should be pruned.")
 
 	// Verify correct App and Asset IDs
-	require.Equal(t, 1, txns[0].asset, "intra == 0 -> ApplicationID = 1")
+	require.Equal(t, 1001, txns[0].asset, "intra == 0 -> ApplicationID = 1")
+	// TODO: How is this one allowed to be <1000 ?
 	require.Equal(t, 789, txns[4].asset, "intra == 4 -> ApplicationID = 789")
```
winder, line 1399: @bbroder-algo this is one of our regression tests that calls into the go-algorand eval function. Somehow it has an appID < 1000

For additional context, I have no idea what this test is doing, I'm just adding 1000 to all the failing assertions.

tzaffi (approved):
Looks good from the perspective of boxes. But I see that E2E test is still failing.

winder:
> Looks good from the perspective of boxes. But I see that E2E test is still failing.

I noticed that too. I'll address in a followup.

PR #1535, opened 2023-05-26
DevOps: update go-algorand submodule to v3.16.0-beta

onetechnical:
## Summary

This is a bookkeeping edit. There were late additions to the release, so the commit has advanced from the initial commit. While there are catchup fixes, these mostly apply for when catchpoint tracking is enabled, which shouldn't be the case for indexer local ledger.

## Test Plan

N/A


onetechnical:
Covered in #1531

PR #1543, opened 2023-06-13
api: Convert timezones before txn query

### LINKED ISSUE #458: Query parameters after-time and before-time for /v2/transactions always assume UTC even if a different timezone is specified
## Summary

Query parameters after-time and before-time for /v2/transactions always assume UTC even if a different timezone is specified.

```bash
$ curl -s "https://testnet.algoexplorerapi.io/idx2/v2/transactions?after-time=2021-04-24T21:30:00-04:00&before-time=2021-04-24T21:30:08-04:00" | jq
{
  "current-round": 13748828,
  "next-token": "9LrRAAAAAAAAAAAA",
  "transactions": [
    {
      "asset-transfer-transaction": {
        "amount": 100000000,
        "asset-id": 12726651,
        "close-amount": 0,
        "receiver": "YM7QN2ORZGARPPMBPMZZMLLJDHSVPZU75HUPMRQUEUU6VQOXADQJ4HUIQM"
      },
      "close-rewards": 0,
      "closing-amount": 0,
      "confirmed-round": 13744884,
      "fee": 1000,
      "first-valid": 13744882,
      "genesis-hash": "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=",
      "genesis-id": "testnet-v1.0",
      "id": "LGM5ZPNGECOXVKQOBUVVF7FX4DIAE24ZDAXWOUCP5ORWX5VTQSFA",
      "intra-round-offset": 0,
      "last-valid": 13745882,
      "note": "gqZwVHhuSWTZJDcwNmVlZGVmLWI0OTctNDcwMi1iYzcwLWNjNTdkYzRjZTIzNKR0eXBlpWlzc3Vl",
      "receiver-rewards": 0,
      "round-time": 1619299805,
      "sender": "UPCPV3UZ4GKXERPFSKDZ3ZRJRN3AFKZYXU7CIENQQ6E4KOWEXRUJ7KOHUQ",
      "sender-rewards": 439,
      "signature": {
        "sig": "YsqUETiATb7n/AauJ0zCdZxPn2WVGpGcAYbJgXD35DCCJ5ZNdg5ZbTsvimnxdMRN2HvKTWJYQJZlimBSi+aSBg=="
      },
      "tx-type": "axfer"
    }
  ]
}
```
where the round timestamp is 1619299805, that is 2021-04-24 17:30:05-04:00.

This is the same result as:
```bash
$ curl -s "https://testnet.algoexplorerapi.io/idx2/v2/transactions?after-time=2021-04-24T21:30:00Z&before-time=2021-04-24T21:30:08Z" | jq
```

Same issue with PureStake API indexer endpoint.

## Scope/Requirements

Support for timezone or require explicitly parameters to be in the UTC timezone and fail otherwise.

## Urgency
Since you can always convert the time you get, this isn't super urgent. However, this is a bug that we should fix.

Eric-Warehime:
## Summary

Fixes #458 

## Test Plan

Added unit tests.

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -564,13 +564,15 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 		partNumber++
 	}
 	if !tf.BeforeTime.IsZero() {
+		convertedTime := tf.BeforeTime.In(time.UTC)
 		whereParts = append(whereParts, fmt.Sprintf("h.realtime < $%d", partNumber))
-		whereArgs = append(whereArgs, tf.BeforeTime)
+		whereArgs = append(whereArgs, convertedTime)
 		partNumber++
 	}
 	if !tf.AfterTime.IsZero() {
+		convertedTime := tf.AfterTime.In(time.UTC)
 		whereParts = append(whereParts, fmt.Sprintf("h.realtime > $%d", partNumber))
-		whereArgs = append(whereArgs, tf.AfterTime)
+		whereArgs = append(whereArgs, convertedTime)
 		partNumber++
 	}
 	if tf.AssetID != 0 || tf.ApplicationID != 0 {
--- a/idb/postgres/postgres_integration_test.go
+++ b/idb/postgres/postgres_integration_test.go
@@ -314,6 +314,58 @@ func TestMultipleWriters(t *testing.T) {
 	assert.Equal(t, amt, balance)
 }
 
+// TestTransactionsTimestamps tests that the transactions endpoint responds to times properly.
+func TestTransactionsTimestamps(t *testing.T) {
+	db, shutdownFunc := setupIdb(t, test.MakeGenesis())
+	defer shutdownFunc()
+
+	round := uint64(1)
+	///////////
+	// Given // A block at with 8 transactions at ts 1671036853.
+	///////////
+	usEastTz, err := time.LoadLocation("America/New_York")
+	require.NoError(t, err)
+	usWestTz, err := time.LoadLocation("America/Los_Angeles")
+	require.NoError(t, err)
+	vb, err := test.ReadValidatedBlockFromFile("test_resources/validated_blocks/BlockWithTransactions.vb")
+	require.NoError(t, err)
+	err = db.AddBlock(&vb)
+	require.NoError(t, err)
+
+	//////////
+	// When // We call Transactions with timestamp filters
+	//////////
+	blkTime := vb.Block.BlockHeader.TimeStamp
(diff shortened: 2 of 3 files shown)


shiqizng:
also add a test in handler to validate this works. 
https://github.com/algorand/indexer/blob/develop/api/handlers_test.go#L80-L82

tzaffi:
> also add a test in handler to validate this works. https://github.com/algorand/indexer/blob/develop/api/handlers_test.go#L80-L82

I echo this sentiment. We can see that previously the coverage was lacking as all the Before/AfterTimes were **UTC**. So we  ought to be able to add a test case that failed previously and is now fixed thanks to this PR.

tzaffi (approved):
LGTM. Optionally, it would be great to change or add one of the cases in `handlers_test.go` to be non-UTC

Eric-Warehime:
@shiqizng @tzaffi
With this implementation we can't add a handlers test because I haven't changed the `idb.TransactionFilter`. It seems like at the handlers level we would assume that the idb implementation would properly handle any `time.Time` objects regardless of their timezone. So instead of converting all idb filter objects to have UTC timezone, I've changed it in the postgres implementation itself.

So the handlers output would be unchanged, but when you get the resulting sql query for postgres it would go from `idb.TransactionFilter{BeforeTime: Non-UTC-TIME}` to something like `SELECT...ON...WHERE h.realtime > UTC-Converted-Time`

tzaffi:
> @shiqizng @tzaffi With this implementation we can't add a handlers test because I haven't changed the `idb.TransactionFilter`. It seems like at the handlers level we would assume that the idb implementation would properly handle any `time.Time` objects regardless of their timezone. So instead of converting all idb filter objects to have UTC timezone, I've changed it in the postgres implementation itself.
> 
> So the handlers output would be unchanged, but when you get the resulting sql query for postgres it would go from `idb.TransactionFilter{BeforeTime: Non-UTC-TIME}` to something like `SELECT...ON...WHERE h.realtime > UTC-Converted-Time`

I wonder if the fixtures test is a good place to add this. I'm not proposing to add it to this PR because the `transactions/` endpoint isn't yet implemented for the fixtures (though it's pretty easy to add it). I can create a small issue to add such a test case if others agree.

shiqizng:
the existing tests are missing cases for `idb.TransactionFilter{BeforeTime: Non-UTC-TIME}` though.  The handler output should change, right? we want prove query provided in [458](https://github.com/algorand/indexer/issues/458) now returns the correct results?

shiqizng (approved):

PR #1544, opened 2023-06-13
api: Omit asset params from acfg delete txns

### LINKED ISSUE #804: Getting blocks from the indexer (and not algod), not possible to determine asset destruction
### Subject of the issue

When getting blocks from algod, if "apar" is missing that means destruction. Great!
However, when getting blocks from indexer, "params" is always present, and it's not possible to distinguish modification/destruction.

Please see conversation here:
https://forum.algorand.org/t/indexer-block-query-asset-destruction/5313/3

Fabrice from the Algorand Foundation agrees this is most probably an issue. 

### Your environment

irrelevant, you can see this using algoexplorer.io indexer (block) vs algod (block).

### Steps to reproduce

1. Find an asset that was destroyed, see that algod for that block has empty "apar", indexer has populated "params".
2. Modify an asset to remove all values (clawback, reserve, manager, etc...) and see that getting blocks from the indexer this is indistinguishable from destruction. 

### Expected behaviour

Any way to distinguish destruction. Either deleted=false, or the "params" section missing (like "apar" for algod) or changing the indexer so modifications report actual quantity (and not zero) and turn that to zero do signal destruction. So, any way would do. I suppose if many are de-referencing the "params" removing it might cause issues and maybe the latter suggestion might help there .

### Actual behaviour

Modification removing all modifiable parameters is undistinguishable from destruction, that is getting *block* from the index (not asset, the block, and looking at asset configuration tx).

# Acceptance Criteria

* Review the current behavior, and the API specification.
* Propose a solution and how it would effect the spec, Indexer and the SDKs.
* Create a followup issue to implement the proposal.

Eric-Warehime:
## Summary

Resolves #804 

## Test Plan

Added to existing acfg delete e2e test in the handlers.

### DIFF
--- a/api/converter_utils.go
+++ b/api/converter_utils.go
@@ -328,26 +328,29 @@ func signedTxnWithAdToTransaction(stxn *sdk.SignedTxnWithAD, extra rowData) (gen
 		}
 		keyreg = &k
 	case sdk.AssetConfigTx:
-		assetParams := generated.AssetParams{
-			Clawback:      addrPtr(stxn.Txn.AssetParams.Clawback),
-			Creator:       stxn.Txn.Sender.String(),
-			Decimals:      uint64(stxn.Txn.AssetParams.Decimals),
-			DefaultFrozen: boolPtr(stxn.Txn.AssetParams.DefaultFrozen),
-			Freeze:        addrPtr(stxn.Txn.AssetParams.Freeze),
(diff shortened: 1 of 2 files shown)


winder (approved):
Nice :+1:

tzaffi (approved):
LGTM

PR #1559, opened 2023-07-07
chore: update deprecated goreleaser option.

winder:
## Summary

Replaces the deprecated replacements option according to recommendation in docs: https://goreleaser.com/deprecations/#archivesreplacements

I don't recall why the old template had a replacement for Arm. @Eric-Warehime do you know if it's needed?

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -90,7 +90,15 @@ nfpms:
         packager: deb
 
 archives:
-  - name_template: '{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
+  - name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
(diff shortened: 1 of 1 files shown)


Eric-Warehime:
> @Eric-Warehime do you know if it's needed?

Not at all.

.goreleaser.yaml
```diff
@@ -90,7 +90,15 @@ nfpms:
...
 archives:
-  - name_template: '{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
+  - name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
+      {{- else if eq .Arch "windows" }}Windows
+      {{- else if eq .Arch "linux" }}Linux
+      {{- else if eq .Arch "darwin" }}Darwin
```
Eric-Warehime, line 100: It looks like based on the docs you linked you don't need overrides for these?

winder: You're right, these are not needed.

Eric-Warehime (approved):

PR #1560, opened 2023-07-24
e2e: set algod version 

shiqizng:
add channel env var in ci configs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,7 @@ jobs:
         type: string
     environment:
       CI_E2E_FILENAME: "f99e7b0c/rel-nightly"
+      CHANNEL: nightly
     steps:
       - go/install:
           version: << parameters.go_version >>
@@ -65,6 +66,7 @@ jobs:
         type: string
     environment:
       CI_E2E_FILENAME: "rel-nightly"
+      CHANNEL: nightly
     steps:
       - go/install:
           version: << parameters.go_version >>
--- a/e2e_tests/docker/indexer/Dockerfile
+++ b/e2e_tests/docker/indexer/Dockerfile
@@ -1,6 +1,5 @@
 ARG GO_IMAGE=golang:1.14.7
 FROM $GO_IMAGE
-ARG CHANNEL=stable
 
 RUN echo "Go image: $GO_IMAGE"
 
--- a/e2e_tests/docker/indexer/docker-compose.yml
+++ b/e2e_tests/docker/indexer/docker-compose.yml
@@ -5,11 +5,14 @@ services:
     build:
       context: ../../../.
       dockerfile: ./e2e_tests/docker/indexer/Dockerfile
+      args:
+        CHANNEL: "${CHANNEL}"
     environment:
       CONNECTION_STRING: "host=e2e-db port=5432 user=algorand password=algorand dbname=indexer_db sslmode=disable"
       CI_E2E_FILENAME: "${CI_E2E_FILENAME}"
       # Add --keep-alive to this variable to pause the container on an error
       EXTRA: "${EXTRA}"
+
 
   e2e-db:
     image: "postgres:13-alpine"


.circleci/config.yml
```diff
@@ -47,6 +47,7 @@ jobs:
         type: string
     environment:
       CI_E2E_FILENAME: "f99e7b0c/rel-nightly"
+      CHANNEL: nightly
```
shiqizng, line 50: should this be pinned to `stable`. do we have a stable test artifact?

winder: I think you're right, we should pin to `stable`.

.circleci/config.yml
```diff
@@ -47,14 +47,14 @@ jobs:
         type: string
     environment:
       CI_E2E_FILENAME: "f99e7b0c/rel-nightly"
```
winder, line 49:
```suggestion
      # A stable environment for PRs.
      # Set CHANNEL to nightly and update CI_E2E_FILENAME when adding a new feature.
      CI_E2E_FILENAME: "f99e7b0c/rel-nightly"
```

Makefile
```diff
@@ -49,6 +49,7 @@ fmt:
 
 # note: when running e2e tests manually be sure to set the e2e filename:
 # 	'export CI_E2E_FILENAME=rel-nightly'
+#	'export CHANNEL=nightly'
```
tzaffi, line 52: Maybe provide default a value with `CHANNEL ?= nightly` ?

shiqizng: setting the default in docker-compose file

.circleci/config.yml
```diff
@@ -145,18 +148,6 @@ commands:
       - run: make e2e
       - run: make e2e-filter-test
 
-  run_e2e_tests_nightly:
```
shiqizng, line 148: this isn't needed anymore after adding the channel var.

Makefile
```diff
@@ -47,8 +47,6 @@ lint:
 fmt:
 	go fmt ./...
 
-# note: when running e2e tests manually be sure to set the e2e filename:
-# 	'export CI_E2E_FILENAME=rel-nightly'
```
shiqizng, line 51: added a default value in docker-compose

tzaffi (approved):
yay!!!

.circleci/config.yml
winder: update.sh is defining `stable` / `nightly` in these two jobs, isn't that still needed?

shiqizng: oops. you're right. we can remove the -nightly suffix, right?

.circleci/config.yml
```diff
@@ -154,8 +157,8 @@ commands:
             wget https://raw.githubusercontent.com/algorand/go-algorand/rel/stable/cmd/updater/update.sh && chmod 744 update.sh
             ./update.sh -i -c nightly -n -d ./ -p /usr/local/go/bin
             export GOPATH=/usr/local/go/
-      - run: make e2e-nightly
-      - run: make e2e-filter-test-nightly
+      - run: make e2e
+      - run: make e2e-filter-test
```
winder, line 161: Change these back?
```suggestion
      - run: make e2e-nightly
      - run: make e2e-filter-test-nightly
```

Makefile
```diff
@@ -58,12 +56,6 @@ e2e: cmd/algorand-indexer/algorand-indexer
 e2e-filter-test: cmd/algorand-indexer/algorand-indexer
 	cd e2e_tests/docker/indexer-filtered/ && docker-compose build --build-arg GO_IMAGE=${GO_IMAGE} && docker-compose up --exit-code-from e2e-read
 
-e2e-filter-test-nightly: cmd/algorand-indexer/algorand-indexer
-	cd e2e_tests/docker/indexer-filtered/ && docker-compose build --build-arg GO_IMAGE=${GO_IMAGE} --build-arg CHANNEL=nightly && docker-compose up --exit-code-from e2e-read
-
-e2e-nightly: cmd/algorand-indexer/algorand-indexer
-	cd e2e_tests/docker/indexer/ && docker-compose build --build-arg GO_IMAGE=${GO_IMAGE} --build-arg CHANNEL=nightly && docker-compose up --exit-code-from e2e
```
winder, line 65: I guess these are removed because nightly is the new default? Does that mean the `make e2e` now runs the nightly channel instead of stable?

.circleci/config.yml
shiqizng: since we set the channel this script and picked up in docker-compose file. I didn't think we need e2e-nightly or e2e-filter-test-nightly anymore.

Makefile
shiqizng: yes, make e2e runs nightly by default. It gets overwritten if CHANNEL is set.

winder: Could you change the defaults so that it normally gives you the stable test and and the nightly job overrides the default?

.circleci/config.yml
winder: Got it.

Makefile
shiqizng: updated

.circleci/config.yml
winder: missed this one?
```suggestion
      CHANNEL: stable
```

winder (approved):

PR #1571, opened 2023-08-31
ci: fix release notes template formatting.

winder:
## Summary

There were two issues with the new release template:
* missing forward slash in the diff URL.
* The triple-dash separator turned the changelog into a header, which was not intentional.


algochoi (approved):
/ 👍

Eric-Warehime (approved):

PR #1590, opened 2023-12-31
Patch: Potential fix for panic on call to block w/ group resource sharing.

gmalouf:


jannotti (approved):
Looks good. I'd kind of like to know why the old code is copying stxn.Txn.Account[k-1] instead of just returning it.  But it seem it's being overzealous. If `sdk.Address` is declared as an array (not slice), this has to be safe.

jasonpaulos (approved):

api/converter_utils.go
```diff
@@ -629,6 +625,20 @@ func signedTxnWithAdToTransaction(stxn *sdk.SignedTxnWithAD, extra rowData) (gen
 	return txn, nil
 }
 
+func edIndexToAddress(index uint64, txn sdk.Transaction, shared []sdk.Address) (sdk.Address, error) {
+	// index into [Sender, txn.Accounts[0], txn.Accounts[1], ..., shared[0], shared[1], ...]
+	switch {
+	case index == 0:
+		return txn.Sender, nil
+	case int(index-1) < len(txn.Accounts):
```
algorandskiy, line 633: I wonder if this new logic should be conditional based on protocol version, @jannotti ?

jannotti: I'd say it's ok. This is a copy of the same code in go-algorand. My reasoning is that the old would error if the index was out of bounds. This code will too, unless the bug that caused a large index _also_ created a SharedAccts array (which would not exist in old versions). So the behavior should always be the same as it was unless we have two coincidental bugs that happen to compensate for each other. In such a case, we return the account that has somehow made its way into `SharedAccts`

PR #1591, opened 2024-01-05
Tests: Group resource shared accounts handling

gmalouf:
## Summary

Introduce test ensuring indexer handles indexing into shared addresses properly when group resource sharing occurs.

## Test Plan

Updated test, ran both before and after patch in place (fails when patch removed, succeeds in current state).


jasonpaulos (approved):
Confirmed test inputs & outputs look good

PR #1592, opened 2024-01-12
writer: deallocate partially prepared statements on error

algorandskiy:
## Summary

Writer does not call Close (and so that DEALLOCATE) if some of statements fail in the middle.
Although go's pg driver's `Prepare` is has map of prepared statements to prevent such errors (see the code excerpt below) it looks more correct to clean up in case of an error.

```golang
// Prepare creates a prepared statement with name and sql. sql can contain placeholders
// for bound parameters. These placeholders are referenced positional as $1, $2, etc.
//
// Prepare is idempotent; i.e. it is safe to call Prepare multiple times with the same
// name and sql arguments. This allows a code path to Prepare and Query/Exec without
// concern for if the statement has already been prepared.
func (c *Conn) Prepare(ctx context.Context, name, sql string) (sd *pgconn.StatementDescription, err error) {
	if name != "" {
		var ok bool
		if sd, ok = c.preparedStatements[name]; ok && sd.SQL == sql {
			return sd, nil
		}
	}
...
	sd, err = c.pgConn.Prepare(ctx, name, sql, nil)
...
	if name != "" {
		c.preparedStatements[name] = sd
	}

	return sd, nil
}
```

## Test Plan

No test, it is hard/impossible to test b/c driver's `Prepare` caches statements.

### DIFF
--- a/idb/postgres/internal/writer/writer.go
+++ b/idb/postgres/internal/writer/writer.go
@@ -124,11 +124,17 @@ func MakeWriter(tx pgx.Tx) (Writer, error) {
 		tx: tx,
 	}
 
+	names := make([]string, 0, len(statements))
 	for name, query := range statements {
 		_, err := tx.Prepare(context.Background(), name, query)
 		if err != nil {
+			for _, name := range names {
+				// this is an escape path, ignore error
+				_ = tx.Conn().Deallocate(context.Background(), name)
+			}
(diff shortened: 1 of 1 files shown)


jasonpaulos (approved):
Makes sense

gmalouf:
@algorandskiy I suggest we ice box/close this for now - the driver for why the work was done may be resolved through configuration

PR #1620, opened 2024-07-31
api: Add support for Private Network Access header preflight requests

nullun:
## Summary

During development of Algorand smart contracts and platforms users will often run local environments consisting of algod, kmd, and indexer via sandbox or more recently algokit. By default all of these services are running on local/private network addresses (e.g. 127.0.0.1), however popular tools such as [DappFlow](https://app.dappflow.org/explorer) and [Lora](https://lora.algokit.io/localnet) are hosted on public network addresses and require the user to specify their local endpoints. Additionally some dapps allow their users to provide their own endpoints for a _more_ decentralised experience.

Schedule for [Google Chrome 130](https://developer.chrome.com/blog/private-network-access-update-2024-03) (although many users are already experiencing it), PNA protections will be enabled by default, disallowing public websites from making requests to local/private resources without a specific header response during a preflight request. This PR introduces a new configuration option that will add middleware to the API handler to support responding to the Private Network Access request header.

## Test Plan

I've currently just compiled this branch locally and tested it by changing the indexer.yml config and passing the command line flag.

### DIFF
--- a/api/handlers_e2e_test.go
+++ b/api/handlers_e2e_test.go
@@ -2613,3 +2613,39 @@ func TestAccounts(t *testing.T) {
 	assert.Equal(t, uint64(1000000000000), (*acctA.Account.CreatedAssets)[0].Params.Total)
 	assert.Equal(t, "bogo", *(*acctA.Account.CreatedAssets)[0].Params.UnitName)
 }
+
+func TestPNAHeader(t *testing.T) {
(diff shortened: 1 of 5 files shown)


nullun:
I rebased on the latest main, as the removal of rewind (changing DeveloperMode) had caused some conflicts.

algorandskiy (approved):

PR #1631, opened 2024-09-15
Fix load of custom consensus - get protocol err

scholtz:
## Summary

This is follow up on 
- https://github.com/algorand/indexer/pull/1327
- https://github.com/algorand/indexer/pull/1041

This PR related to 
- https://github.com/algorand/conduit/pull/162
- https://github.com/algorand/go-algorand-sdk/pull/625

Currently there is the bug in indexer and conduit that prevents usage with custom consensus file.

image

## Test Plan

We have built and we use the custom build of indexer and conduit made from this pull requests.
-https://github.com/scholtz/AlgorandNodes/blob/main/docker/indexer/aramidmain/compose-indexer-fix.dockerfile

Docker images:
- https://hub.docker.com/r/scholtz2/aramid-indexer/tags
- https://hub.docker.com/r/scholtz2/aramid-conduit/tags

### DIFF
--- a/cmd/algorand-indexer/daemon.go
+++ b/cmd/algorand-indexer/daemon.go
@@ -15,6 +15,8 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
+	sdkConfig "github.com/algorand/go-algorand-sdk/v2/protocol/config"
+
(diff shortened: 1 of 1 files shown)


scholtz:
updated pull request in 

https://github.com/algorand/go-algorand-sdk/issues/754

PR #1638, opened 2024-11-18
Implement the endpoint `GET /v2/block-headers`

agodnic:
## Summary

This pull request implements the endpoint `GET /v2/block-headers` (ported from https://github.com/AlgoNode/indexer-api-cdb/pull/10).

In particular, this enables searching for blocks by different participation-related parameters (i.e.: proposer address, expired accounts, absent accounts).

## Test Plan

The new endpoint has been tested manually, using the FNet dataset:
```
# no parameters
curl -Ss 'http://127.0.0.1:8980/v2/block-headers?limit=2' | jq

# filter by timestamp
curl -Ss 'http://127.0.0.1:8980/v2/block-headers?limit=2&after-time=2024-10-01T00:00:00Z' | jq

# filter by block proposer
curl -Ss 'http://127.0.0.1:8980/v2/block-headers?proposers=FNETC46DGTSSDHA6C54FWEGZ3Z5ADE4YAPYQTX5VEE2YGU5NFPAWTRANCE&limit=1' | jq

# filter by absent participation accounts
curl -Ss 'http://127.0.0.1:8980/v2/block-headers?absent=WXN65H5M7BRSABP545OES2NCMFGE7R7OICDZSXCTP5H5H2CHXCDKFQTF6U&limit=1' | jq

# filter by expired participation accounts
curl -Ss 'http://127.0.0.1:8980/v2/block-headers?expired=2D2HFM3EIYFNMBNLWZPK5E4SWVN6SR7RDSVCBQ4Z6HRHE26DBT6WQYO6XA&limit=1' | jq
```

So far there are no automated tests for this feature. Maybe the tests could look like these (mutatis mutandis):
- https://github.com/algorand/indexer/blob/e6a823d9d7f89ac789c5621fcee753f318f6f843/api/handlers_test.go#L30
- https://github.com/algorand/indexer/blob/e6a823d9d7f89ac789c5621fcee753f318f6f843/api/handlers_test.go#L226
- https://github.com/algorand/indexer/blob/e6a823d9d7f89ac789c5621fcee753f318f6f843/api/handlers_test.go#L407


api/handlers.go
```diff
@@ -981,6 +1002,79 @@ func (si *ServerImplementation) LookupTransaction(ctx echo.Context, txid string)
...
+	var round uint64
+	var nextToken string
+	results := make([]generated.Block, 0)
+	err := callWithTimeout(ctx, si.log, si.timeout, func(ctx context.Context) error {
+
+		// Open a channel from which result rows will be received
+		var rows <-chan idb.BlockRow
+		rows, round = si.db.Blocks(ctx, bf)
+
+		// Iterate receieved rows, converting each to a generated.Block
```
jannotti, line 1050: "received"

api/indexer.oas2.json
```diff
@@ -853,6 +853,67 @@
         }
       }
     },
+    "/v2/blocks": {
+      "get": {
+        "description": "Search for blocks. Blocks are returned in ascending round order. Transactions are not included in the output.",
```
jannotti, line 858: Since there are no transactions, the endpoint might better be called `blockheaders`. I think that's what the similar algod endpoint under development is called.

api/indexer.oas2.json
```diff
@@ -2677,6 +2738,61 @@
     }
   },
   "parameters": {
+    "proposer": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "x-algorand-format": "Address"
+      },
+      "description": "Account(s) marked as proposer in the block header's participation updates. This parameter accepts a comma separated list of addresses.",
```
jannotti, line 2747: I think it's fine, maybe even more correct, to simply say "Accounts" here, where the number is unknown.

api/indexer.oas2.json
```diff
@@ -2677,6 +2738,61 @@
...
+      "in": "query",
+      "required": false
+    },
+    "participation": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "x-algorand-format": "Address"
+      },
+      "description": "Account(s) marked as expired, absent or as proposer in the block header's participation updates. This parameter accepts a comma separated list of addresses.",
```
jannotti, line 2791: This seems like a weird catch-all. I guess the intent is for block explorers that want to show any relevant block for particular accounts?

agodnic: I believe so. Maybe @urtho can give more context on how the `participation` parameter is intended to be used.

Performance-wise, it is the most problematic parameter. I can see the standard version of the indexer not including the `participation` parameter (I regard it as a low benefit / high cost feature)

Something similar happens with the `updates` parameter, maybe not as bad.

agodnic: > Is there any meaningful advantage to it, compared to calling the endpoint three times, specifying the address for each of the three things? I would guess you're using an index for each, so it's basically three index lookups no matter how you slice it.
> 
> These lists (expired and suspended) are almost always empty. It seems like an index lookup would be super fast. Just do two of them?

That's exactly right. The essential functionality is already being exposed by the "atomic" parameters (that is: `proposer`, `absent` and `expired`)

The data returned by the `updates` and `participation` parameters could be obtained by calling the endpoint multiple times using the "atomic" parameters.

api/converter_utils.go
```diff
@@ -298,6 +308,94 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
 	return txn, nil
 }
 
+func rowToBlock(blockHeader *sdk.BlockHeader) generated.Block {
```
jannotti, line 311: I haven't written much indexer code, but this name seems a bit underspecified. Above, we see `txnRowToTransaction`, so this should probably be `hdrRowToBlock`. I'm not totally confident about this recommendation because I would have expected this to take some sort of `idb` type, rather than an `sdk` type.

api/converter_utils.go
```diff
@@ -34,8 +34,18 @@ func decodeDigest(str *string, field string, errorArr []string) (string, []strin
 	return "", errorArr
 }
 
-// decodeAddress returns the byte representation of the input string, or appends an error to errorArr
-func decodeAddress(str *string, field string, errorArr []string) ([]byte, []string) {
+// decodeAddress returns the sdk.Address representation of the input string, or appends an error to errorArr
+func decodeAddress(str string, field string, errorArr []string) (sdk.Address, []string) {
```
jannotti, line 38: I see that `decodeAddress` already existed, so this PR is not to blame for the unusual way that multiple errors are accumulated, but I think there is a better way to handle multiple errors these days.

I would use `sdk.DecodeAddress` directly, and if it returns an `error`, append that into a `[]error` instead of `[]string`.  Use `fmt.Errorf` with a `%w` to produce an error that wraps the error from `sdk.DecodeAddress` but also contains the `str` that would have been passed into `decodeAddress`), rather than into a slice of `string`.  Then at the end of a function that may have accumulated errors, test its size just as you do, but then use `errors.Join(errArr...)` to combine them into a single `error`.

I'll leave it to others to decide if this is important enough to change. I believe it will make it easier to follow the error handling, but it isn't critical if this is a common pattern in `indexer`.

api/converter_utils.go
```diff
@@ -298,6 +308,94 @@ func txnRowToTransaction(row idb.TxnRow) (generated.Transaction, error) {
...
+
+	// order these so they're deterministic
+	orderedTrackingTypes := make([]sdk.StateProofType, len(blockHeader.StateProofTracking))
+	trackingArray := make([]generated.StateProofTracking, len(blockHeader.StateProofTracking))
+	elems := 0
+	for key := range blockHeader.StateProofTracking {
+		orderedTrackingTypes[elems] = key
+		elems++
+	}
+	sort.Slice(orderedTrackingTypes, func(i, j int) bool { return orderedTrackingTypes[i] < orderedTrackingTypes[j] })
```
jannotti, line 363: `slices.Sort` is a more convenient, and usually faster, way to sort.

api/converter_utils.go
```diff
@@ -640,7 +738,7 @@ func edIndexToAddress(index uint64, txn sdk.Transaction, shared []sdk.Address) (
 }
 
 func (si *ServerImplementation) assetParamsToAssetQuery(params generated.SearchForAssetsParams) (idb.AssetsQuery, error) {
-	creator, errorArr := decodeAddress(params.Creator, "creator", make([]string, 0))
+	creator, errorArr := decodeAddressToBytes(params.Creator, "creator", make([]string, 0))
 	if len(errorArr) != 0 {
 		return idb.AssetsQuery{}, errors.New(errUnableToParseAddress)
```
jannotti, line 743: As an example of what I described earlier, I think this reads much better, without needing to understand the unusual error handling model of `decodeAddress`.

```suggestion
	creator, err := sdk.DecodeAddress(params.Creator)
	if err {
		return idb.AssetsQuery{}, fmt.Errorf("Unable to parse creator address: %v", err)
```

It's worth noting that the existing code supplies the word "creator" to `decodeAddressToBytes` in order to get a nicer error message, and then immediately throws that nice error message away in favor of a generic message.

You can also just use `creator[:]` in the code below, rather than have separate functions to decode to an `sdk.Address`, or to `[]byte`.

api/converter_utils.go
jannotti: Oh, even nicer: 
```Join returns an error that wraps the given errors. Any nil error values are discarded. Join returns nil if every value in errs is nil.```
therefore, you don't need to test the `err` that is returned from `sdk.DecodeAddress`, just append it to your error list.  Then, at the bottom, don't check to see if the list has errors in it, just:
```
  err := errors.Join(errList...)
  if err != nil {
    return err
  }
```

File: api/indexer.oas2.json (hunk 4 above)
Comment (agodnic) [reply api/indexer.oas2.json]: Addressed in https://github.com/algorand/indexer/pull/1638/commits/8cb90c11876a9bc8fa70f42c0b7321a6625bdfba

File: api/indexer.oas2.json (hunk 3 above)
Comment (agodnic) [reply api/indexer.oas2.json]: Addressed in https://github.com/algorand/indexer/pull/1638/commits/a302e5b9948fa1dace8f77f2b31bfeb8227d0bac

File: api/indexer.oas2.json (hunk 2 above)
Comment (agodnic) [reply api/indexer.oas2.json]: The route was changed to `GET /v2/block-headers`, in https://github.com/algorand/indexer/pull/1638/commits/a302e5b9948fa1dace8f77f2b31bfeb8227d0bac

File: api/converter_utils.go (hunk 7 above)
Comment (agodnic) [reply api/converter_utils.go]: Addressed in https://github.com/algorand/indexer/pull/1638/commits/82e864bfb760ebec9056d80c83d63f7929f1eb65.

There are more instances of the `sort.Slice` function being used, but I don't want to pollute the diff. Can totally do this change (potentially in a separate PR) if you guys want.

File: api/converter_utils.go (hunk 5 above)
Comment (agodnic) [reply api/converter_utils.go]: That's fair, the name `hdrRowToBlock` will generate less confusion when reading the code.

Addressed in https://github.com/algorand/indexer/pull/1638/commits/eaefec7e83a256b20b77ab63ac75fd63e5874fca

I could create a specific `idb` type for the input parameter, but it would be a struct with only one field. Seemed a bit like overengineering or unnecessary abstraction. If the struct would have had more than one field, I believe it would have made sense.

It could go the other way though - "similar functions having similar signatures" (as in `hdrRowToblock` looking similar to `txnRowToTransaction`) is a desirable property to preserve.

File: api/converter_utils.go (hunk 8 above)
Comment (agodnic) [reply api/converter_utils.go]: This change makes a lot of sense:
* removes unnecessary code (there are two functions that pretty much do the same thing as `sdk.DecodeAddress`)
* the logic in those functions is somewhat confusing to the reader
* handling errors as `[]error` is preferrable over `[]string`

Commit https://github.com/algorand/indexer/pull/1638/commits/7f26e0fbc48bfd655fcb0cc1cf6b3fced7e7fc8c adds more context information to the errors being returned. If this is considered a breaking change, I can make it return the same exact old error messages.

I'm not a fan of returning multiple errors, I think it is not worth the extra complexity compared to a fail-fast approach (returning upon the first error encountered). In fact this could introduce bugs, since the control flow keeps advancing with variables that may be in an unknown state (I'd rather stop the control flow immediately upon finding something that doesn't check out).
But there may be a reason why multiple errors are being used that I'm not aware of.

There are several instances in the codebase that are handling multiple errors as `[]string` instead of `[]error`. I didn't change all of these to avoid scope creep, but could do so in a separate PR.

File: api/converter_utils.go (hunk 6 above)
Comment (agodnic) [reply api/converter_utils.go]: See https://github.com/algorand/indexer/pull/1638#discussion_r1853159861

File: api/converter_utils.go (hunk 5 above)
Comment (agodnic) [reply api/converter_utils.go]: That's fair, the name `hdrRowToBlock` will generate less confusion when reading the code.

Addressed in https://github.com/algorand/indexer/pull/1638/commits/eaefec7e83a256b20b77ab63ac75fd63e5874fca and https://github.com/algorand/indexer/pull/1638/commits/8924c97009a56066562086b499273a822bbd4261

File: api/indexer.oas2.json [hunk 9]
```diff
@@ -2677,6 +2732,39 @@
     }
   },
   "parameters": {
+    "proposer": {
```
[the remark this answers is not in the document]
agodnic: @gmalouf: Yes we can!

Implemented in https://github.com/algorand/indexer/pull/1638/commits/dff12aac53f9157752d87e54c0b2acd763bae609

gmalouf (approved):

api/indexer.oas2.json
```diff
@@ -3242,4 +3356,4 @@
       "name": "search"
     }
   ]
-}
+}
```
jannotti, line 3359: I'd rather we didn't gratuitously remove final newlines.

jannotti (approved):

PR #1645, opened 2025-01-13
Rewind: Re-enable Rewinding

gmalouf:
## Summary

Revert "Disable rewinding and reject the  query param on account lookups and searches. (#1630)"

This reverts commit 249016cf51cf9c3aecf37c6fc8c09774523f02bc. The synthetic transaction implementation of payouts in the indexer should allow balances retrieved via rewind to calculate as before.

## Test Plan

Pre-existing tests were added back.


jannotti:
Should we add a test to show it works with incentive payouts?  I don't know enough about how indexer is tested to know if that's a big ask.

accounting/rewind.go
```diff
@@ -0,0 +1,188 @@
...
+			if addr == stxn.Txn.AssetSender || addr == stxn.Txn.Sender {
+				assetUpdate(&acct, uint64(stxn.Txn.XferAsset), stxn.Txn.AssetAmount+txnrow.Extra.AssetCloseAmount, 0)
+			}
+			if addr == stxn.Txn.AssetReceiver {
+				assetUpdate(&acct, uint64(stxn.Txn.XferAsset), 0, stxn.Txn.AssetAmount)
+			}
+			if addr == stxn.Txn.AssetCloseTo {
+				assetUpdate(&acct, uint64(stxn.Txn.XferAsset), 0, txnrow.Extra.AssetCloseAmount)
+			}
+		case sdk.AssetFreezeTx:
```
jannotti, line 165: We should add, `case sdk.HeartbeatTx:` assuming that's what it is called.  It should be fine to rewind past one.

gmalouf:
> Should we add a test to show it works with incentive payouts? I don't know enough about how indexer is tested to know if that's a big ask.

I wasn't planning to; the payout is itself a pay transaction and has tests ensuring indexer handles it. The existing rewind tests are light for sure, IOW I personally wouldn't hold this PR up for it.

accounting/rewind.go
gmalouf: Good call: [aa75fa323d1a47fffbfd59917eda281fd240c56e](https://github.com/algorand/indexer/pull/1645/commits/aa75fa323d1a47fffbfd59917eda281fd240c56e)

cshintov:
Does this mean the rewind works and the conditions existed that necessitated the removal is no longer there?

jannotti:
> Does this mean the rewind works and the conditions existed that necessitated the removal is no longer there?

Rewind works again, but only as well as it has in the past. App and keyreg transactions still cannot be rewound.

PR #1646, opened 2025-02-01
Search for online accounts

agodnic:
## Summary
_(This pull request is part of the features described in https://github.com/AlgoNode/indexer-api-cdb/issues/4. The aforementioned features will be presented in separate, smaller pull requests to make it easier to accept/reject/discuss each one individually)_

This pull request adds the `online-only` parameter to `GET /v2/accounts`, which lets the user search for accounts whose participation status is "online".

## Test Plan
* Added test cases to `handlers_e2e_test.go`
* Tested manually in a local environment.

## Performance
The performance of `is-online=true` lookups can be improved with the following index:

```
-- create the index
ledgerdb=# CREATE INDEX account_idx_online_only
        ON account ((((account_data ->> 'onl')) = '1'), addr)
        WHERE (NOT deleted) AND ((account_data ->> 'onl') IS NOT NULL) AND ((account_data ->> 'voteLst') IS NOT NULL)
;
CREATE INDEX

-- Verify the index is being used by the query optimizer.
-- This query corresponds to /v2/accounts?limit=10&online-only=true&exclude=all
ledgerdb=# explain (analyze) WITH qaccounts AS (SELECT a.addr, a.microalgos, a.rewards_total, a.created_at, a.closed_at, a.deleted, a.rewardsbase, a.keytype, a.account_data FROM account a WHERE NOT a.deleted AND (a.account_data->>'onl' IS NOT NULL) and (account_data->>'voteLst' IS NOT NULL) and ((a.account_data->>'onl') = '1') ORDER BY a.addr ASC LIMIT 10) SELECT za.addr, za.microalgos, za.rewards_total, za.created_at, za.closed_at, za.deleted, za.rewardsbase, za.keytype, za.account_data FROM qaccounts za ORDER BY za.addr ASC;
                                                                      QUERY PLAN                                                                      
------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.28..37.70 rows=10 width=103) (actual time=0.023..0.038 rows=10 loops=1)
   ->  Index Scan using account_idx_online_only on account a  (cost=0.28..407632.13 rows=108930 width=103) (actual time=0.023..0.036 rows=10 loops=1)
         Index Cond: (((account_data ->> 'onl'::text) = '1'::text) = true)
 Planning Time: 0.268 ms
 Execution Time: 0.049 ms
(5 rows)

```


agodnic:
## Discussion
1. If we want to treat suspended accounts differently from other online accounts, my understanding is that we'd need to somehow persist that information in the database (probably within the `account` table).
It could be a new integer nullable field: `account.account_data->suspended_until`

2. A follow-up question would be whether we want to add a parameter in `GET /v2/accounts` to search for suspended accounts.

jannotti:
A suspended account has status = 0 (offline) but has non-zero key material (SelectionID, VoterID, etc).

gmalouf:
@agodnic thank you for the PR, a couple thoughts:

- There could be performance impacts for API services on the new parameter, perhaps we should support enabling/disabling it as we do with other params: [Interpreting the Configuration](https://github.com/algorand/indexer?tab=readme-ov-file#interpreting-the-configuration). There are pre-existing disabled and enabled parameters tests to help exercise this change as well.
- Covering the functionality with an additional test case in [handlers_e2e_test.go](https://github.com/algorand/indexer/blob/main/api/handlers_e2e_test.go) would be ideal.

agodnic:
@gmalouf thanks for taking the time to review this.

> There could be performance impacts for API services on the new parameter, perhaps we should support enabling/disabling it as we do with other params: Interpreting the Configuration. There are pre-existing disabled and enabled parameters tests to help exercise this change as well.

The new `online-only` parameter has been set to disabled by default, and the parameter list in README.md and DisablingParametersGuide.md has been updated.

> Covering the functionality with an additional test case in handlers_e2e_test.go would be ideal.

Three basic test cases have been added to exercise the new parameter.

I could be more thorough with the e2e tests, but in order to do so I'd have to create new blocks like the ones in https://github.com/algorand/indexer/tree/084577338ad4882f5797b3e1b30b84718ad40333/api/test_resources/validated_blocks. Not sure what's the best approach to generate those test blocks - I could use a msgpack editor and create the block fields from scratch, but maybe there's a less error-prone way.

api/handlers_e2e_test.go
```diff
@@ -1147,6 +1147,100 @@ func TestVersion(t *testing.T) {
 	require.Equal(t, response.Version, "(unknown version)")
 }
 
+// TestAccountsOnlineOnlyParam exercises the `online-only` parameter in `GET /v2/accounts`.
+func TestAccountsOnlineOnlyParam(t *testing.T) {
+
+	db, shutdownFunc := setupIdb(t, test.MakeGenesis())
+	defer shutdownFunc()
+
+	//txn := transactions.SignedTxnWithAD{
```
gmalouf, line 1156: We should remove the commented out code prior to merging.

api/handlers_e2e_test.go
```diff
@@ -1147,6 +1147,100 @@ func TestVersion(t *testing.T) {
...
+		require.Equal(t, http.StatusOK, rec.Code)
+		var response generated.AccountsResponse
+		data := rec.Body.Bytes()
+		err = json.Decode(data, &response)
+		require.NoError(t, err)
+		require.Equal(t, len(test.MakeGenesis().Allocation), len(response.Accounts))
+	}
+	{
+		//////////
+		// When // We query for accounts using both `online-only=true` and `include-all=true` parameters
```
gmalouf, line 1228: Probably good to exercise the error on online-only + deleted here as well?

gmalouf (commented):
@agodnic this is looking really good. I put a few clean up asks in, but I think this will be mergeable after that.

api/handlers_e2e_test.go
agodnic: Removed in https://github.com/algorand/indexer/pull/1646/commits/63b444f2c479cb1bb15562223905b27277392dfa.

I intially wrote those comments because some of the tests in the file have them - I thought it was some kind of unwritten convention. Turned out it wasn't.

api/handlers_e2e_test.go
agodnic: My undestanding is that the error is already being tested for:

```go
generated.SearchForAccountsParams{OnlineOnly: boolPtr(true), IncludeAll: boolPtr(true)}
```

`include-all=true` would be equivalent to `deleted=true`, if I understand correctly.

```go
		err = api.SearchForAccounts(c, generated.SearchForAccountsParams{OnlineOnly: boolPtr(true), IncludeAll: boolPtr(true)})
		//////////
		// Then // The response should be a 404 "bad request"
		//////////
		require.NoError(t, err)
		require.Equal(t, http.StatusBadRequest, rec.Code)
		require.Contains(t, rec.Body.String(), errOnlineOnlyDeleted)
```

The caveat is that the `SearchForAccounts` method doesn't return an error per-se, but a `nil` error with a 404 "bad request" response.

Or maybe I misunderstood your comment - my apologies in that case

agodnic:
@gmalouf thanks again for allocating time to this. I'll be available if anything else needs to be changed.

gmalouf (approved):

Issue #135, opened 2020-06-22
Search by group-id to get all Transactions within Group


ryanRfox:
As an `algorand-indexer` user, I'd like the ability to search by "group-id" so I may receive the complete set of confirmed transactions matching this group-id.

ryanRfox:
Possible solution: 
Within /v2/transactions define "group-id" as a searchable parameter. Return each "transaction" within the "transactions" array.

brianolson:
The currently expected workaround is to fetch all the transactions for a round and then filter them by group id in client code.

ian-algorand:
Some design is needed still on this.

brianolson:
question 0: how did you _get_ the group-id such that you want to search for it?

If you submitted the group you can have followed it through to being committed and then know the round it was committed in and do the above workaround of fetching the txns for a round and filtering client side on group-id. We could add server side filtering; it would be pretty efficient to query for in current infrastructure if a round number was required to be submitted when querying to filter by group-id.

The expensive full service option is to add another index on the txn table comparable to the index by txid. I will habitually push back against this solution because I worked on a product that became cost-prohibitive and too slow because of all the indexes; so I want to be very cautious about any new index added. In my opinion, group-id, like txid, is not a good primary key to fetch transactions by. The best way to look up transactions is by round or by address.

brianolson:
Alternate possible implementation: if a query returns a single txn or some txns all of the same group-id, have a flag `?expand-to-group=1` or similar that would result in the server internally doing the second query for everything by group-id. Already having a txn to refer to the system would have everything it needs for an efficient query of neighboring txns in the same group. `expand-to-group=1` would do nothing if the return set includes multiple group-ids or any txns with empty group-id.

jeapostrophe:
I have a few questions and comments on @brianolson 's proposed workaround.

First, I do not know if the Indexer returns the group txn index. The spec --- https://developer.algorand.org/docs/reference/rest-apis/indexer/#transaction --- includes the "intra-round-offset" field. Is that intended to be the same thing as the "GroupIndex"?

Second, it would be very useful for the indexer to have the invariant that the transactions would always be returned in the group order, so that when I do response.transactions[i], I get transaction number i.

Third, is true that all rounds include only one transaction group? I was under the impression that a round was a block and a block could contain many transaction groups.

[...]

PR #1647, opened 2025-03-10
Add `group-id` parameter to `GET /v2/transactions`

agodnic:
## Summary

Issue: https://github.com/algorand/indexer/issues/135

This pull request adds the `group-id` parameter to the endpoint `GET /v2/transactions` (this feature has already been implemented for Nodely in https://github.com/AlgoNode/indexer-api-cdb/pull/2).

This new parameter is disabled by default.

## Test plan

End-to-end tests have been added to `api/handlers_e2e_test.go`.


agodnic:
Turning this PR back into a draft to implement the Postgres optimization

agodnic:
@gmalouf thanks for taking the time to review this. I've merged with the latest commit in the `main` branch, got the conflicts resolved

idb/postgres/postgres.go
```diff
@@ -708,9 +714,32 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 		// this should explicitly match the primary key on txn (round,intra)
 		query += " ORDER BY t.round, t.intra"
 	}
-	if tf.Limit != 0 {
-		query += fmt.Sprintf(" LIMIT %d", tf.Limit)
+
+	// Determine the LIMIT clause
```
gmalouf, line 718: Are you actually seeing better performance with the group id filter overriding the limit filter (vs a straight limit)?

agodnic: Saw it on a mainnet database. If I remember correctly, the server was running pg14.6. @urtho can confirm.

@urtho came up with the idea of removing the `LIMIT`, and it did lead to a better execution plan.

I still don't understand why that happens though - in theory the `LIMIT` clause should not affect the execution plan.

PR #1662, opened 2025-07-01
TransactionApplication Improvements

gmalouf:
## Summary

Based off additional feedback from #1661.
Initialize capacity of boxRefs returned from transactions endpoints based off internal boxrefs slice size. 
Handle the self/zero-index case for boxrefs application id specifically.

## Test Plan

Existing tests should pass.


api/converter_utils.go
```diff
@@ -483,7 +483,7 @@ func signedTxnWithAdToTransaction(stxn *sdk.SignedTxnWithAD, extra rowData) (gen
 			assets = append(assets, uint64(v))
 		}
 
-		boxRefs := make([]generated.BoxReference, 0)
+		boxRefs := make([]generated.BoxReference, len(stxn.Txn.BoxReferences))
```
algorandskiy, line 486:
```suggestion
		boxRefs := make([]generated.BoxReference, 0, len(stxn.Txn.BoxReferences))
```

jannotti (approved):

PR #1663, opened 2025-07-03
API: Make BoxReference API type consistent with Algod to assist in SDK cod…

gmalouf:
## Summary

Make BoxReference API type consistent with Algod to assist in SDK code generation compatibility.

## Test Plan

Existing tests should pass.


algorandskiy (approved):
Because the `BoxReference` type was introduced two days ago in https://github.com/algorand/indexer/pull/1661 it is ok to rename

Issue #116, opened 2020-06-05
queries without limits returning next token


JasonWeathersby:
curl localhost:8980/v2/accounts/XIU7HGGAJ3QOTATPDSIIHPFVKMICXKHMOR2FJKHTVLII4FAOA3CYZQDLG4/transactions?after-time=2020-06-03T10:00:00-05:00 | json_pp

brianolson:
**Improvement**
It'd be nicer and more efficient if the server detected that it finished short of its limit (whether default or specified) and didn't send a next token when it can know that there's nothing after that next token.

In the current state, adding the next token to the query produces an empty result, which is a pretty clear sign of being done, but yeah it could be better.

gmalouf:
Feedback from the community is that there are several applications that rely on the `nextToken` is always present.

PR #1667, opened 2025-07-21
Addressing issue #116 on all applicable query endpoints

gmalouf:
## Summary
NextToken should only be set if there are more results to query on all applicable query endpoints.

Closes issue #116 

## Test Plan

Added new `handlers_test` cases exercising the `nextToken` setting behavior when the db returns `limit` vs `limit+1` results.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -281,11 +281,12 @@ func (si *ServerImplementation) LookupAccountAppLocalStates(ctx echo.Context, ac
 		return notFound(ctx, errValueExceedingInt64)
 	}
 
+	requestedLimit := uintOrDefaultValue(params.Limit, si.opts.DefaultApplicationsLimit)
 	search := generated.SearchForApplicationsParams{
 		Creator:       &accountID,
 		ApplicationId: params.ApplicationId,
 		IncludeAll:    params.IncludeAll,
-		Limit:         params.Limit,
+		Limit:         uint64Ptr(requestedLimit + 1), // Request one more than needed
 		Next:          params.Next,
 	}
 	options, err := si.appParamsToApplicationQuery(search)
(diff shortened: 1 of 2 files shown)


api/handlers.go
```diff
@@ -281,11 +281,12 @@ func (si *ServerImplementation) LookupAccountAppLocalStates(ctx echo.Context, ac
 		return notFound(ctx, errValueExceedingInt64)
 	}
 
+	requestedLimit := uintOrDefaultValue(params.Limit, si.opts.DefaultApplicationsLimit)
```
jannotti, line 284: I'd prefer the was `effectiveLimit` or just `limit`, since the default thing means it may not be the "requested" limit.

api/handlers.go
```diff
@@ -676,12 +695,15 @@ func (si *ServerImplementation) SearchForApplicationBoxes(ctx echo.Context, appl
 	}
 	happyResponse := generated.BoxesResponse{ApplicationId: applicationID, Boxes: []generated.BoxDescriptor{}}
 
+	requestedLimit := uint64(20) // Default limit for boxes
```
jannotti, line 698: Why is this done differently?  The others had some structure that contained the limit?

api/handlers.go
```diff
@@ -1127,9 +1190,23 @@ func (si *ServerImplementation) SearchForTransactions(ctx echo.Context, params g
...
 
+	// Check if we have more results than requested (indicating more pages exist)
+	var nextToken *string
+	if len(txns) > int(requestedLimit) {
+		// More results exist, so we need to provide a next token
+		// The fetchTransactions function handles next token generation internally
+		nextToken = strPtr(next)
+		// Truncate results to only return the requested amount
+		txns = txns[:requestedLimit]
+	} else if len(txns) > 0 && next != "" {
```
jannotti, line 1201: I don't under stand why there's `if len(txns) > 0 && next != ""`.  If the first thing didn't happen, `nextToken should be nilled unconditionally, I think.

gmalouf: Yup, dropping - its codegen that I missed fixing.

algorandskiy (approved):

kadircancetin:
> @kadircancetin Is this something easily addressable or you find it better to always have a next-token even when there are no further transactions to page through?

We fixed it, and it is okay for us now. It was easy to fix. But it broke the existing pulling loop we had before. 

Before this implementation, we basically fetch the next token, and if it is empty, we know it is the last page. But suddenly, because the next token has disappeared, our code was broken.

PR #1671, opened 2025-08-08
API: Transactions endpoints NextToken fix

gmalouf:
## Summary

The `next` token is not being returned consistently when querying transaction pagination endpoints. This PR addresses the bug.

The logic determining whether to return the nextToken needs to check against raw db results prior to any deduplication when querying transactions.

## Test Plan

New tests were created ensuring that nextToken is returned in cases where there are more results in the db even after deduplication in-memory.


gmalouf:
Closing, we've decided to revert the nextToken change entirely/ go back to longstanding behavior of always returning it.

PR #1677, opened 2025-10-05
before-time/after-time optimization for participation queries

urtho:
## Summary

Recent optimization for before/after-time queries does not work for searches that filter by participating account.
Postgres plan optimizer chooses to loop through all of the account history without applying the round filter. 

This PR forces the round filter to be applied to the participation table index scan, resulting in way faster plans. 
Zero cons, all pros. 

Example API call where this optimization works:

`/v2/transactions?address=I2IE7JMC5PXOSC2XZE6DURBW4PASD2JGIAFMEJ5SKL4K5NV5YROUQOOHCE&after-time=2025-09-29T22%3A44%3A52.000Z`

### DIFF
--- a/idb/postgres/postgres.go
+++ b/idb/postgres/postgres.go
@@ -576,15 +576,25 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 	}
 	if !tf.BeforeTime.IsZero() {
 		convertedTime := tf.BeforeTime.In(time.UTC)
-		whereParts = append(whereParts, fmt.Sprintf("t.round <= ("+
-			"SELECT round from block_header WHERE realtime < $%d ORDER BY realtime DESC LIMIT 1)", partNumber))
+		if joinParticipation {
+			whereParts = append(whereParts, fmt.Sprintf("p.round <= ("+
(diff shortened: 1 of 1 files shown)


gmalouf (approved):

idb/postgres/postgres.go
```diff
@@ -529,32 +529,32 @@ func buildTransactionQuery(tf idb.TransactionFilter) (query string, whereArgs []
 				whereArgs = append(whereArgs, addrBase64)
 				partNumber++
 			}
-			if tf.AddressRole&idb.AddressRoleReceiver != 0 {
+			if tf.AddressRole&idb.AddressRoleReceiver != 0 && tf.TypeEnum == idb.TypeEnumPay {
```
jannotti: I don't understand these optimizations. I _think_ a query can specify that it wants to match if an address matches the reciever and it can just leave TypeEnum blank, meaning "any".  That should still cause the where clause to be added, right?  But this seems to require that the query explicitly specifies the type is pay.

I would have expected almost the exact opposite optimization. If a query specifies that it cares about the receiver, then you could _add_ to the where clause to check the type of the txn is a pay, since otherwise the rcv is blank (and I'm thinking maybe we index the type, so you would speed up the query).

PR #1679, opened 2025-10-22
build: use Go 1.25

gmalouf:
## Summary

Bumps Go version from 1.23 to 1.25 and golangci-lint to 2.5 (now executed via `go run` rather than expected to be base installed).

Various new linter warnings were addressed, and the configuration was migrated to be 2.x compatible.

## Test Plan

Existing tests should pass


jannotti (dismissed):

api/converter_utils.go
```diff
@@ -405,7 +405,7 @@ func signedTxnWithAdToTransaction(stxn *sdk.SignedTxnWithAD, extra rowData) (gen
 	switch stxn.Txn.Type {
 	case sdk.PaymentTx:
 		p := generated.TransactionPayment{
-			CloseAmount:      uint64Ptr(uint64(stxn.ApplyData.ClosingAmount)),
+			CloseAmount:      uint64Ptr(uint64(stxn.ClosingAmount)),
```
cce: I don't like this

api/converter_utils.go
```diff
@@ -721,7 +721,7 @@ func signedTxnWithAdToTransaction(stxn *sdk.SignedTxnWithAD, extra rowData) (gen
 		keys := make([]tuple, 0)
 
 		for k := range stxn.ApplyData.EvalDelta.LocalDeltas {
-			addr, err := edIndexToAddress(k, stxn.Txn, stxn.ApplyData.EvalDelta.SharedAccts)
+			addr, err := edIndexToAddress(k, stxn.Txn, stxn.EvalDelta.SharedAccts)
```
cce: Similarly, I think this is bad, you should know it's ApplyData.EvalDelta

jannotti: On this one, I probably wouldn't comment on this regardless of which way the author originally wrote it, but I agree we should not force the author not to use `ApplyData`.

algorandskiy (approved):

PR #1696, opened 2026-06-24
API: add Value to BoxDescriptor and Round to BoxResponse

algorandskiy:
## Summary

Sync `BoxDescriptor` API type with algod.
Additionally had to bump go-sdk version here and in conduit to make the test pass without failing msgp decoding of blocks with new `ld` (Load) field.

## Test Plan

1. Added new tests
2. Updated box responses in e2e test with Round field
3. Checked e2e docker tests pass with corresponding conduit PR https://github.com/algorand/conduit/pull/196


api/handlers_e2e_test.go
```diff
@@ -1964,11 +1964,46 @@ func compareAppBoxesAgainstHandler(t *testing.T, db *postgres.IndexerDb,
...
+		for _, box := range valuesResp.Boxes {
+			require.Contains(t, expectedBoxes, string(box.Name), msg)
+			require.NotNil(t, box.Value, msg)
+			require.Equal(t, expectedBoxes[string(box.Name)], string(*box.Value), msg)
+			// // An empty box value may round-trip as a nil pointer; treat it as "".
+			// var gotValue string
+			// if box.Value != nil {
+			// 	gotValue = string(*box.Value)
+			// }
+			// require.Equal(t, expectedBoxes[string(box.Name)], gotValue, msg)
```
joe-p: I assume this is commented out because there is no empty values? If so, we should probably just remove this commented-out code

joe-p (dismissed):
One nit, but otherwise LGTM

api/handlers_e2e_test.go
algorandskiy: oh, right, I was running CI on this code for debugging at one point, removed

algorandskiy:
Merged conduit PR, removed debugging code.

joe-p (approved):

PR #1701, opened 2026-09-14
api: fix application logs next-token looping on inner transactions

### LINKED ISSUE #1700: `/v2/applications/{id}/logs` returns the same `next-token` forever when the last result is an inner transaction
### Subject of the issue

On `GET /v2/applications/{application-id}/logs`, if the last result of a page comes from an **inner** transaction, the `next-token` points at the **root** transaction's intra instead of the inner transaction's. The next request selects that inner transaction again and builds the same token again.

- If the inner transaction is the last match in the query, the endpoint never stops: every following page contains that one entry and the same `next-token`. A client that pages until `next-token` is absent loops forever.
- If it is not the last match, the next page starts with a duplicate of the previous page's last entry.

This impacted the automated [xGov committee generation](https://github.com/algorandfoundation/xgov-committees/pull/75) and resulted in an infinite next() loop and a stalled job after a service timeout was reached.

### Your environment

- Indexer `3.10.0-ndly` (public Nodely mainnet endpoint `https://mainnet-idx.4160.nodely.dev`)
- Reproduced from source against `algorand/indexer` `main` (`f84ad3e`) with the e2e fixture `api/test_resources/validated_blocks/LookupMultiInnerLogs.vb`

### Steps to reproduce

Mainnet, app `3147789458`. Its logs in this range are partly emitted by inner calls from app `3378436620`.

1. Request the first page:
   ```
   GET /v2/applications/3147789458/logs?min-round=62000000&max-round=65000000&limit=1000
   ```
   → 58 entries, last txid `4L3L6EGUBATKUWBHCNDFFFB3IBUZUWNSJ6MN3A7ZZJUNINJPU76Q`, `next-token: upvfAwAAAAATAAAA`
2. Request the next page:
   ```
   GET /v2/applications/3147789458/logs?min-round=62000000&max-round=65000000&limit=1000&next=upvfAwAAAAATAAAA
   ```
   → 1 entry, identical to the last entry of step 1, `next-token: upvfAwAAAAATAAAA` again
3. Repeat step 2: it returns the same response indefinitely.

The token decodes to round `64986042`, intra `19`. The last txid is the **root** at intra `19`. It has 4 inner transactions, and the first of them (intra `20`) is the call to `3147789458` that emitted the log.

For comparison, `min-round=61000000&max-round=64000000` ends on a log emitted by a **top-level** call to the app (`WUPO7E…`, intra `86`, 2 inners). Its token decodes to intra `88` (86 + 2), and the following page is empty with no `next-token`, as expected.

### Expected behaviour

The token points past the last returned row, so the next page never repeats it and paging ends with no `next-token`.

### Actual behaviour

The token points at the root transaction's intra, so the last inner-transaction result is returned again with the same `next-token`.

tasosbit:
## Summary

Fix `next-token` for `/v2/applications/{application-id}/logs` when a page ends on an inner transaction

The logs endpoint sets `SkipInnerTransactionConversion`, so inner transactions are returned as their own rows and `TxnRow.RootTxn` is `nil`. `TxnRow.Next` still built the token from `Extra.RootIntra`, and when ascending added `countInner(tr.Txn)`. Stored inner rows have their nested inner transactions stripped, so that count is `0` and the token pointed at the root transaction's intra. The next request (`intra > rootIntra`) selected the same inner transaction again and produced the same token. On the last page this repeated forever; on earlier pages it duplicated the last entry.

`Next` now returns the row's own intra for an inner transaction that was returned without its root. Rows with `RootTxn` populated (the default conversion mode) and top-level transactions are unchanged.

Closes #1700

## Test Plan

- **`api/handlers_e2e_test.go`:** added `TestLookupInnerLogsPagination`. It loads `LookupMultiInnerLogs.vb` and pages through `LookupApplicationLogsByID` with `limit=1` for app `789` (one inner transaction with logs) and app `222` (three inner transactions with logs), asserting that every log is returned exactly once in order, that `next-token` changes on every page, and that paging ends with an empty page and no `next-token` within 10 pages. Without the fix both subtests fail with `next token did not advance` (token `AQAAAAAAAAAAAAAA`, round 1 / root intra 0). With the fix both pass.
- **`idb/idb_test.go`:** added `TestTxnRowNext` cases for an inner transaction row without `RootTxn`, ascending and descending. Both expect the row's own intra (`51`), not `RootIntra` (`50`). The existing "root txn absent" error case is unchanged.
- **Regression runs:** `go test ./api/... ./idb/... -count=1` passes on the previous `main` (`3d20fd7`). On current `main` (`f84ad3e`), `TestTxnRowNext`, `TestLookupInnerLogs`, `TestLookupMultiInnerLogs` and `TestLookupInnerLogsPagination` pass. `gofmt` and `go vet ./api/ ./idb/` are clean.
- **Mainnet repro (before the fix):** paging `/v2/applications/3147789458/logs?min-round=62000000&max-round=65000000&limit=1000` returns 58 entries with `next-token=upvfAwAAAAATAAAA`. Every following page returns only the last entry (`4L3L6E…`, an inner call at round `64986042`, root intra `19`) with the same token.


idb/idb.go
```diff
@@ -74,6 +74,15 @@ func (tr TxnRow) Next(ascending bool) (string, error) {
...
 
+	// An inner transaction without its root was returned as its own result
+	// (SkipInnerTransactionConversion), so resume right after this row. Using the
+	// root intra would select this row again, and on the last page return the same
+	// next token forever.
+	if tr.Extra.RootIntra.Present && tr.RootTxn == nil && tr.Txn != nil {
+		binary.LittleEndian.PutUint32(b[8:], uint32(tr.Intra))
+		return base64.URLEncoding.EncodeToString(b[:]), nil
+	}
+
```
jannotti, line 85: While this isn't code I know well, it seems reasonable.

